PR #1, opened 2018-10-25
Ensure project can be published as an NPM module

dfunckt:
Change-type: patch

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM balena/open-balena-base:v4.4.2
 EXPOSE 80
 
 COPY package.json package-lock.json /usr/src/app/
-RUN npm ci --unsafe-perm --production && npm cache clean --force
+RUN npm ci --unsafe-perm --production --ignore-scripts && npm cache clean --force
 
 COPY . /usr/src/app
 
--- a/package.json
+++ b/package.json
@@ -1,11 +1,21 @@
 {
   "name": "@balena/open-balena-api",
-  "private": true,
   "description": "Internet of things, Made Simple",
   "version": "0.0.1",
   "repository": {
     "type": "git",
     "url": "https://github.com/balena-io/open-balena-api"
+  },
+  "main": "dist/index",
+  "files": [
+    "dist/"
+  ],
+  "scripts": {
+    "build": "npm run clean && tsc --project . && copyup \"src/**/*.sbvr\" \"src/**/*.sql\" dist/",
+    "clean": "rimraf dist/",
+    "lint": "resin-lint --typescript src/ && tsc --noEmit --project .",
+    "prettify": "prettier --config ./node_modules/resin-lint/config/.prettierrc --write \"src/**/*.ts\" \"typings/**/*.ts\"",
+    "prepare": "npm run build"
   },
   "dependencies": {
     "@resin.io/device-types": "^10.4.1",
@@ -66,7 +76,14 @@
     "typed-error": "^2.0.0",
     "typescript": "^3.1.3"
   },
-  "optionalDependencies": {},
+  "devDependencies": {
+    "copyfiles": "^2.1.0",
+    "husky": "^0.14.3",
+    "lint-staged": "^7.3.0",
+    "prettier": "^1.14.3",
+    "resin-lint": "^2.0.0",
+    "rimraf": "^2.6.1"
+  },
   "engines": {
     "node": ">=8.0"
   }
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -1,9 +1,5 @@
-import * as Promise from 'bluebird';
-import { readdir } from 'fs';
 import { sbvrUtils } from '../platform';
 import { retrieveAPIKey } from '../platform/api-keys';
-
-const readdirAsync = Promise.promisify(readdir);
 
 sbvrUtils.addPureHook('all', 'all', 'all', {
 	PREPARSE: ({ req }) => {
@@ -13,8 +9,14 @@ sbvrUtils.addPureHook('all', 'all', 'all', {
 	},
 });
 
-readdirAsync(__dirname + '/resources').each(initScript => {
-	if (/\.ts$/.test(initScript)) {
-		return require(__dirname + '/resources/' + initScript);
-	}
-});
+import './resources/api_key';
+import './resources/application';
+import './resources/device';
+import './resources/envvars';
+import './resources/image__is_part_of__release';
+import './resources/image';
+import './resources/release';
+import './resources/service_install';
+import './resources/service_instance';
+import './resources/service';
+import './resources/user';
--- a/src/lib/device-types/index.ts
+++ b/src/lib/device-types/index.ts
@@ -12,7 +12,7 @@ import {
 } from './build-info-facade';
 import { getImageKey, IMAGE_STORAGE_PREFIX, listFolders } from './storage';
 
-const { BadRequestError, NotFoundError } = sbvrUtils;
+export const { BadRequestError, NotFoundError } = sbvrUtils;
 
 export type DeviceType = deviceTypesLib.DeviceType;
 
--- a/src/lib/errors.ts
+++ b/src/lib/errors.ts
@@ -1,7 +1,7 @@
 import TypedError = require('typed-error');
 import { sbvrUtils } from '../platform';
 
-const { NotFoundError } = sbvrUtils;
+export const { NotFoundError } = sbvrUtils;
 
 export class NoDevicesFoundError extends NotFoundError {}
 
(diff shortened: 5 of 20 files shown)


Dockerfile
```diff
@@ -2,12 +2,8 @@ FROM balena/open-balena-base:v4.4.2
 
 EXPOSE 80
 
-COPY package.json package-lock.json /usr/src/app/
-RUN npm ci --unsafe-perm --production && npm cache clean --force
-
 COPY . /usr/src/app
-
-RUN npx tsc --noEmit
+RUN npm ci --unsafe-perm --production && npm cache clean --force
```
Page-: Why did you remove the separate
```
COPY package.json package-lock.json /usr/src/app/
RUN npm ci --unsafe-perm --production && npm cache clean --force
```
?
The new version reduces the caching a lot so will slow down builds

package.json
```diff
@@ -1,12 +1,22 @@
...
   "repository": {
     "type": "git",
     "url": "https://github.com/balena-io/open-balena-api"
   },
+  "main": "dist/index",
+  "files": [
+    "dist/"
+  ],
+  "scripts": {
+    "build": "npm run clean && tsc --project . && copyup \"src/**/*.sbvr\" \"src/**/*.sql\" dist/",
```
Page-: Is `copyup` from this `copyfiles` dep?

package.json
```diff
@@ -38,6 +48,7 @@
     "compression": "^1.7.3",
     "cookie-parser": "^1.4.1",
     "cookie-session": "^1.0.0",
+    "copyfiles": "^2.1.0",
```
Page-: This looks like it should be a dev dep

package.json
```diff
@@ -60,13 +71,19 @@
     "request": "^2.85.0",
     "resin-device-config": "^4.2.0",
     "resin-semver": "^1.4.0",
+    "rimraf": "^2.6.1",
```
Page-: This also seems like a dev dep

src/hooks/index.ts
```diff
@@ -13,8 +9,14 @@ sbvrUtils.addPureHook('all', 'all', 'all', {
 	},
 });
 
-readdirAsync(__dirname + '/resources').each(initScript => {
-	if (/\.ts$/.test(initScript)) {
-		return require(__dirname + '/resources/' + initScript);
-	}
-});
+import('./resources/api_key');
```
Page-: Any reason for these being async imports?  It looks like they could just be imported directly since we're at the top level

src/lib/device-types/index.ts
```diff
@@ -12,7 +12,7 @@ import {
 } from './build-info-facade';
 import { getImageKey, IMAGE_STORAGE_PREFIX, listFolders } from './storage';
 
-const { BadRequestError, NotFoundError } = sbvrUtils;
+export const { BadRequestError, NotFoundError } = sbvrUtils;
```
Page-: Why are these being exported now?

src/lib/errors.ts
```diff
@@ -1,7 +1,7 @@
 import TypedError = require('typed-error');
 import { sbvrUtils } from '../platform';
 
-const { NotFoundError } = sbvrUtils;
+export const { NotFoundError } = sbvrUtils;
```
Page-: Why is this being exported now?

src/routes/services.ts
```diff
@@ -31,16 +32,19 @@ export const vpn = {
 				captureException(err, 'Error authenticating device for VPN', { req });
 				res.status(500).send(err);
 			}),
-	clientConnect: (req: Request, res: Response) => {
+	clientConnect: (req: Request, res: Response): Promise<void> => {
 		const body = req.body || {};
 		if (!body.common_name) {
-			return res.sendStatus(400);
+			res.sendStatus(400);
+			return null as any;
```
Page-: Why is this not just `return`?  That would avoid this odd type casting

src/routes/services.ts
```diff
@@ -31,16 +32,19 @@ export const vpn = {
...
 		}
 		if (!body.virtual_address) {
-			return res.sendStatus(400);
+			res.sendStatus(400);
+			return null as any;
```
Page-: Why is this not just `return`?  That would avoid this odd type casting

src/routes/services.ts
```diff
@@ -31,16 +32,19 @@ export const vpn = {
...
 		}
 		if (!body.service_id) {
-			return res.sendStatus(400);
+			res.sendStatus(400);
+			return null as any;
```
Page-: Why is this not just `return`?  That would avoid this odd type casting

src/routes/services.ts
```diff
@@ -31,16 +32,19 @@ export const vpn = {
 				captureException(err, 'Error authenticating device for VPN', { req });
 				res.status(500).send(err);
 			}),
-	clientConnect: (req: Request, res: Response) => {
+	clientConnect: (req: Request, res: Response): Promise<void> => {
```
Page-: The real response type is `void | Promise<void>` - which is why I think you've done the weird type casting below

src/routes/services.ts
```diff
@@ -67,10 +71,11 @@ export const vpn = {
 			});
 	},
 
-	clientDisconnect: (req: Request, res: Response) => {
+	clientDisconnect: (req: Request, res: Response): Promise<void> => {
```
Page-: The real response type is `void | Promise<void>` - which is why I think you've done the weird type casting below

Dockerfile
dfunckt: `npm` will invoke the `prepare` script during `npm ci` which triggers a build of the project. Without first copying the whole project into `/usr/src/app` there are no files to compile (or a tsconfig.json) and it fails. I don't like this TBH, because we're compiling sources during build but aren't really using them -- open-balena-api still starts up like resin-api, compiling on-demand. Any ideas?

package.json
dfunckt: Yes, `copyup` is a "binary" equivalent to `copyfiles --up 1 ...`. I could change back to that if you think it's clearer.

package.json
dfunckt: It can't unfortunately because we're triggering a build during `npm ci` when building the image.

Dockerfile
Page-: There's a `--ignore-scripts` for `npm i` that might work for ci?

package.json
Page-: No it's fine, I just wasn't sure if it was some mac binary or not

Dockerfile
dfunckt: Ah, interesting, I'll take a look.

package.json
Page-: `npm ci` will still install dev deps unless you specify `npm ci --production`

src/lib/device-types/index.ts
dfunckt: TS complains we're extending a private type below.

src/lib/errors.ts
dfunckt: Same as above -- TS complains.

src/routes/services.ts
dfunckt: TS complains again :) It treats `return` as equivalent to `return undefined` and complains the signature is incompatible.

src/lib/device-types/index.ts
Page-: Ah right, because `declaration: true` was added, that kinda sucks because it's not actually private, just not exported by this module

src/routes/services.ts
Page-: Well the signature *is* incompatible - it specifies a promise is returned when it's not

Dockerfile
dfunckt: That worked nicely 👍

Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM balena/open-balena-base:v4.4.2
 EXPOSE 80
 
 COPY package.json package-lock.json /usr/src/app/
-RUN npm ci --unsafe-perm --production && npm cache clean --force
+RUN npm ci --unsafe-perm --ignore-scripts && npm cache clean --force
```
Page-: You probably still want the `--production`

dfunckt: Oh right, we're not building the project anymore.

src/hooks/index.ts
```diff
@@ -13,8 +9,14 @@ sbvrUtils.addPureHook('all', 'all', 'all', {
 	},
 });
 
-readdirAsync(__dirname + '/resources').each(initScript => {
-	if (/\.ts$/.test(initScript)) {
-		return require(__dirname + '/resources/' + initScript);
-	}
-});
+import * as _api_key from './resources/api_key';
```
Page-: Just to check, these imports are still in the generated js right?

dfunckt: They're not… good catch. Fixed.

PR #2, opened 2018-10-26
Add the /config/var endpoint to the open API

pimterry:
This extracts the env var config logic out of the cloud API, and reduces the existing vars down to only the supported options (no deltas).


src/routes/config.ts
```diff
@@ -0,0 +1,113 @@
...
+	RESERVED_NAMESPACES,
+	INVALID_CHARACTER_REGEX,
+	WHITELISTED_NAMES,
+	WHITELISTED_NAMESPACES,
+	BLACKLISTED_NAMES,
+} from '../lib/env-vars';
+
+// Note that this list will be mutated by the cloud API to add
+// extra variables as for cloud-only features.
+export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
```
dfunckt: I think it makes sense to move all these `const`s into `lib/env-vars` since they're really tied together.

pimterry: You're right, that makes a lot of sense, done.

src/routes/config.ts
```diff
@@ -0,0 +1,45 @@
...
+import { JSONSchema6 } from 'json-schema';
+
+import {
+	RESERVED_NAMES,
+	RESERVED_NAMESPACES,
+	INVALID_CHARACTER_REGEX,
+	WHITELISTED_NAMES,
+	WHITELISTED_NAMESPACES,
+	BLACKLISTED_NAMES,
+    SUPERVISOR_CONFIG_VAR_PROPERTIES,
```
dfunckt: Weird spacing here. `npm run prettify` should be available on current master (otherwise it'll be introduced by #1).

src/routes/config.ts
```diff
@@ -0,0 +1,45 @@
...
+            ...SUPERVISOR_CONFIG_VAR_PROPERTIES,
+
+            ...(_.includes(
+                RESIN_HOST_CONFIG_CAPABLE_DEVICE_TYPES,
+                req.query.deviceType,
+            ) ? HOST_CONFIG_VAR_PROPERTIES : {})
+        }
+    };
+
+	const varsConfig = {
```
dfunckt: Here too.

pimterry:
No, seems like prettier isn't available. That's fine - I'll wait until your PR is merged, and then rebase & fix formatting.

dfunckt:
@pimterry I just merged #1.

pimterry:
All prettified, merging :+1:

PR #4, opened 2018-10-26
Expose API as v5

dfunckt:
Change-type: patch


dfunckt:
@Page- @thgreasi @pimterry I won't be merging this before resin-api has v5 translations and the SDK switches to v5 as well.

PR #8, opened 2018-11-01
Drop the pubnub backend

Page-:
Change-type: major


dfunckt:
And I also think you'll still want `Change-type: minor` instead of major, despite being on pre-1.0.0 versions.

Page-:
Oh I didn't release this was 0.x versions, and yeah they do change the minor/major semantics

PR #9, opened 2018-11-01
Reduce api queries used to update service installs by using $in

Page-:
Change-type: patch


Page-:
@dfunckt https://github.com/balena-io/open-balena-api/pull/10

PR #13, opened 2018-11-02
Make config generation more generic

dfunckt:
This moves resolving the device type and OS version to callers of `generateConfig` so that it can be parameterised.

Also fixes deprecated usages of `req.param()`.

Change-type: minor


src/lib/device-config.ts
```diff
@@ -13,7 +13,7 @@ const readFileAsync = Promise.promisify(readFile as (
 import * as deviceConfig from 'resin-device-config';
 import * as resinSemver from 'resin-semver';
 
-import * as deviceTypes from './device-types';
+import * as _deviceTypesLib from './device-types';
```
zvin: nit: can't we use 
```typescript
import { DeviceType } from './device-types';
```
?

dfunckt: We could -- it's just out of habit when importing only types from a module. In these cases, TS will not emit a `require()` and the underscore prefix in the "module" name (`_deviceTypesLib`) makes it visually apparent (though it's not statically enforced, TS will still emit a require if you reference an actual value exported by the module).

zvin: I think it won't emit a require if you only import a type, but I'm not 100% sure.

dfunckt: Yes, that's correct. To be clear, it will not emit a require using either form of import. This form is just me applying the style I found being used in other places in the codebase.

Issue #18, opened 2018-11-06
Should allow users permissions to lookup service instances and request device tunnels


wrboyce:
service instance lookups and tunnels to port 22222 are required for balena-cli ssh.

PR #20, opened 2018-11-06
perms: Permit users permissions required to ssh into a device

wrboyce:
Connects-to: #18
Change-type: patch

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -94,6 +94,9 @@ export const ROLES: {
 		'resin.service_install.all',
 		'resin.service_label.all',
 		'resin.user.read',
+
+		`resin.service_instance.get?service_type eq 'vpn'`,
+		`resin.device.tunnel-22222?`,
 	],
 };
 


src/lib/auth.ts
```diff
@@ -94,6 +94,9 @@ export const ROLES: {
 		'resin.service_install.all',
 		'resin.service_label.all',
 		'resin.user.read',
+
+		`resin.service_instance.get?service_type eq 'vpn'`,
+		`resin.device.tunnel-22222?`,
```
dfunckt: Stray question mark right there

dfunckt: You also don’t need the backticks — single or double (above) quotes will do as you’re doing no interpolation.

dfunckt: One more thing :) Can you please either sort them a/b or add a nice comment about them just above so they’re grouped together?

wrboyce: Fixed.

Issue #17, opened 2018-11-06
Should allow service instances to specify their ip_address when registering


wrboyce:
See balena-io/open-balena-vpn#17

dfunckt:
We aren't going forward with this @wrboyce right?

wrboyce:
Yep, closing.

PR #21, opened 2018-11-06
service_instance: Allow instances to specify their ip_address

wrboyce:
Connects-to: #17
Change-type: minor


wrboyce:
@Page- from balena-io/open-balena-vpn#17

> This will be useful in cases such as the default `open-balena` setup where haproxy is relied on for forwarding traffic and the vpn's ip_address would be the container's address on the (inaccessible) docker network.

PR #22, opened 2018-11-06
auth: Create admin account on first run

richbayliss:
If the container is started with the following ENV vars set;
- BALENA_ROOT_EMAIL
- BALENA_ROOT_PASSWORD

then on startup it will try and create the admin user with these values.

Change-type: patch


index.js
```diff
@@ -68,6 +68,26 @@ setup(app, {
...
+			registerUser({ username: 'root', email, password }, tx),
+		)
+		.catch(() => {
+			console.log("Root account already exists!")
+		})
+		.return(result);
+	} else {
+		return result;
+	}	
+})
```
dfunckt: Can you please move this into its own function and just invoke from here?

richbayliss: sure thing

src/commands/dummy.ts
```diff
@@ -0,0 +1,17 @@
+// import * as express from "express";
```
Page-: Why comment out this file?  I think it makes more sense to leave it executable as the equivalent of a "hello world"

index.js
```diff
@@ -68,6 +93,9 @@ setup(app, {
 	onInitModel,
 	onInitHooks,
 })
+.tap(() => {
+	return onInitCreateAdmin();
```
Page-: This can be:
```typescript:
.tap(onInitCreateAdmin)

File: index.js [hunk 4]
```diff
@@ -58,6 +58,31 @@ function onInitHooks() {
...
+	return runInTransaction(tx =>
+		registerUser({ username: 'root', email, password }, tx),
+	)
+	.then(() => {
+		console.log("Admin account created successfully!");
+	})
+	.catch(() => {
+		console.log("Admin account already exists!");
+	})
+	.return();
```
Page-: This `.return()` is unnecessary as you're already returning void

index.js
```diff
@@ -58,6 +58,31 @@ function onInitHooks() {
...
+	const { registerUser } = require('./src/platform/auth');
+
+	return runInTransaction(tx =>
+		registerUser({ username: 'root', email, password }, tx),
+	)
+	.then(() => {
+		console.log("Admin account created successfully!");
+	})
+	.catch(() => {
+		console.log("Admin account already exists!");
```
Page-: I think that a specific error type will be thrown if it already exists (`ConflictError`?), so it'd probably be better to catch that one specifically

index.js
```diff
@@ -58,6 +58,31 @@ function onInitHooks() {
...
 
+function onInitCreateAdmin() {
+	const email = process.env.BALENA_ROOT_EMAIL || '';
+	const password = process.env.BALENA_ROOT_PASSWORD || '';
+
+	if (email == '' || password == '') {
+		return;
+	}
+
+	console.log("Creating admin account...")
```
Page-: This should be single quotes, did prettier not run?

index.js
```diff
@@ -58,6 +58,31 @@ function onInitHooks() {
 	)
 }
 
+function onInitCreateAdmin() {
+	const email = process.env.BALENA_ROOT_EMAIL || '';
```
Page-: These should be coming from `config.ts` which handles grabbing env vars/defaulting if necessary

src/commands/dummy.ts
```diff
@@ -0,0 +1,17 @@
...
+// export function execute(
+//   _app: express.Application,
+//   args: string[]
+// ): Promise<void> {
+//   // grab your args as an array, like so...
+//   const [arg1, arg2, argN] = args;
+
+//   // return a promise which resolves to void...
+//   return Promise.resolve();
+// }
```
dfunckt: Can you please delete this file? If you still need to show an example command, please move it into the README.

index.js
dfunckt: You can invoke it as `.tap(onInitCreateAdmin)`

index.js
```diff
@@ -58,6 +58,31 @@ function onInitHooks() {
 	)
 }
 
+function onInitCreateAdmin() {
```
dfunckt: The function can be named simply `createAdmin`. Also, we've been referring to it as `superuser` all along (admins are a different thing in the context of Resin and can be confusing) and it'd be good to change the function name and messages below accordingly.

index.js
```diff
@@ -58,6 +58,31 @@ function onInitHooks() {
...
+	const email = process.env.BALENA_ROOT_EMAIL || '';
+	const password = process.env.BALENA_ROOT_PASSWORD || '';
+
+	if (email == '' || password == '') {
+		return;
+	}
+
+	console.log("Creating admin account...")
+
+	const { runInTransaction } = require('./src/platform');
```
Page-: These are untyped, you can either use
```typescript
import('./src/platform')
.then(({ runInTransaction }) => {
	...
}
```
 or
```typescript
import * as _platform from './src/platform'
...
const { runInTransaction }: typeof _platform = require('./src/platform')
```
to make sure they get typed.  You can also use:
```typescript
const { runInTransaction } = await import('./src/platform')
```
as a variation of the first one

index.js
```diff
@@ -58,6 +58,31 @@ function onInitHooks() {
 	)
 }
 
+function onInitCreateAdmin() {
+	const email = process.env.BALENA_ROOT_EMAIL || '';
+	const password = process.env.BALENA_ROOT_PASSWORD || '';
```
dfunckt: I think you don't need `|| ''` -- your conditional below would then become:

```
if (email == null || password == null) {
  return;
}
```

index.js
richbayliss: It did... weird, I shall take a look...

index.js
richbayliss: OK cool - that's a good suggestion

index.js
richbayliss: It did run, so I don't know what happened here...

index.js
richbayliss: done

src/commands/dummy.ts
richbayliss: I have kept it - but made it actually executable, as a working example

index.js
richbayliss: agreed - and have switched to referencing as "superuser"

index.js
```diff
@@ -58,6 +58,34 @@ function onInitHooks() {
...
 
+function createSuperuser() {
+
+	const { SUPERUSER_EMAIL, SUPERUSER_PASSWORD } = require('./src/lib/config');
+
+	if (!SUPERUSER_EMAIL || !SUPERUSER_PASSWORD) {
+		return;
+	}
+
+	console.log("Creating superuser account...")
```
dfunckt: Prettier is not processing this file because it's plain Javascript and I think we configure it to only touch .ts files in src/.

richbayliss:
@resin-ci re-test

richbayliss:
@resin-ci re-check

richbayliss:
@resin-ci retest

Issue #16, opened 2018-11-05
Should not retry to fetch device types for ever


thgreasi:
If something goes wrong with the s3 device types we currently recursively try to refetch them.
Chaining all those promises increasingly consumes more and more memory.
We should fail with an error after a few retries since the http requests would have timeout anyway.

dfunckt:
Device type fetching has changed significantly and this is no longer really relevant.

thgreasi:
Thanks for spotting this!

PR #23, opened 2018-11-06
device-types: Limit the retries for fetching the device types

thgreasi:
* This prevents the infinite chain of retried request promises. The respective client requests will fail with a 500 after some time, instead of a timeout.
* Uses the InternalServerError pine class so that the rejection details are not propagated to the clients not only when the methods are used in routes but also from inside hooks.

Resolves: #16
Change-type: patch

### DIFF
--- a/src/lib/device-types/index.ts
+++ b/src/lib/device-types/index.ts
@@ -1,6 +1,7 @@
 import * as arraySort from 'array-sort';
 import * as Promise from 'bluebird';
 import * as _ from 'lodash';
+import { InternalRequestError } from '@resin/pinejs/out/sbvr-api/errors';
 import * as deviceTypesLib from '@resin.io/device-types';
 import * as semver from 'resin-semver';
 import { sbvrUtils } from '../../platform';
@@ -282,12 +283,16 @@ export const getImageSize = (slug: string, buildId: string) => {
 					throw new UnknownVersionError(slug, buildId);
 				}
 
-				return getCompressedSize(normalizedSlug, buildId).tapCatch(err => {
-					captureException(
-						err,
-						`Failed to get device type ${slug} compressed size for version ${buildId}`,
-					);
-				});
+				return getCompressedSize(normalizedSlug, buildId);
+			},
+		).catch(
+			err => !(err instanceof UnknownVersionError),
+			(err: Error) => {
+				captureException(
+					err,
+					`Failed to get device type ${slug} compressed size for version ${buildId}`,
+				);
+				throw new InternalRequestError(err);
 			},
 		);
 	});
@@ -322,6 +327,10 @@ export const getImageVersions = (slug: string): Promise<ImageVersions> => {
 					versions: buildIds,
 					latest: buildIds[0],
 				};
+			})
+			.catch(err => {
+				captureException(err, `Failed to get device type ${slug} versions`);
+				throw new InternalRequestError(err);
 			});
 	});
 };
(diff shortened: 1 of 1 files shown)


src/lib/device-types/index.ts
```diff
@@ -135,8 +154,7 @@ function updateDeviceTypesCache(
 	freshDeviceTypes: Promise<Dictionary<DeviceTypeInfo>>,
 ) {
 	if (!deviceTypesCache) {
-		deviceTypesCache = freshDeviceTypes;
-		return freshDeviceTypes;
+		return freshDeviceTypes.return();
```
thgreasi: Removes the redundant `deviceTypesCache = freshDeviceTypes;` so that this only gets assigned from a within `fetchDeviceTypesAndReschedule`.
No need to return anything from in here.

src/lib/device-types/index.ts
```diff
@@ -186,6 +204,12 @@ function fetchDeviceTypesAndReschedule(): Promise<Dictionary<DeviceTypeInfo>> {
 	// in case another api request comes before the first completes
 	if (!deviceTypesCache) {
 		deviceTypesCache = promise;
+		// make sure we remove rejected promises from the cache
+		promise.catch(() => {
+			if (deviceTypesCache === promise) {
+				deviceTypesCache = undefined;
```
thgreasi: The initial request is only one that is assigned to the cache before it resolves.
Before this PR we would retry fetching the device types forever, so the cached promise could never settle as rejected.
We now have to handle make sure to clear the cache if the initial request fails.

src/lib/device-types/index.ts
```diff
@@ -103,29 +107,44 @@ function fetchDeviceTypes(): Promise<Dictionary<DeviceTypeInfo>> {
...
+			captureException(
+				{ ...err, retryAttemptsLeft },
+				'Failed to get device types',
+			);
+			if (retryAttemptsLeft >= 1) {
+				return Promise.delay(RETRY_DELAY).then(() =>
+					fetchDeviceTypes(retryAttemptsLeft - 1),
+				);
+			}
+			throw new InternalRequestError(err);
```
thgreasi: This ensures that we will not return any details about the error in the response body both when used in a route, as well as when used in a hook.

src/lib/device-types/index.ts
Page-: Any reason for not using memoizee promise mode for caching the device types?  It will automatically handle clearing the cache on failure

thgreasi: @Page-  The only reason I did this manually since the first iteration is that memoizee doesn't provide a "cache-then-update" caching schema.
The reasoning is that after the API boots and gets a warm cache, we prefer to have near zero cost on accessing the deviceTypes, since it's used in various endpoints as well as in hooks.
If we simply went the `memoizee` road, even when using `prefetch` we could end up in situations that incoming requests are waiting as long as 5 secs for the cache to be repopulated.

For other not so critical part, like fetching all the available versions of a device type, we do use memoizee on the s3 requests.

The img maker is doing something similar as well, so I guess the reasoning might be similar.

dfunckt:
@thgreasi any reason not to get this merged?

thgreasi:
@dfunckt not really, I was just overwhelmed by other deadlines.
I will try to pick this up this Friday.

thgreasi:
Also, thanks for the heads-up :+1:

thgreasi:
Let me close this, since it got old enough...

Issue #25, opened 2018-11-08
Should change the env var descriptions to not contain `resin`


thgreasi:
See: https://github.com/balena-io/open-balena-api/blob/master/src/lib/env-vars.ts#L97
See: https://www.flowdock.com/app/rulemotion/namechange/threads/UzSK9Dggf0vLWVmg1KnZ_sT23L2

PR #26, opened 2018-11-08
env-vars: Change descriptions for the balena rename

thgreasi:
Resolves: #25
Change-type: minor

### DIFF
--- a/src/lib/env-vars.ts
+++ b/src/lib/env-vars.ts
@@ -76,12 +76,12 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
 	},
 	RESIN_SUPERVISOR_LOG_CONTROL: {
 		enum: ['false', 'true'],
-		description: 'Enable / Disable logs from being sent to Resin',
+		description: 'Enable / Disable logs from being sent to balena',
 		default: 'true',
 	},
 	RESIN_SUPERVISOR_POLL_INTERVAL: {
 		type: 'integer',
-		description: 'Define the Resin API Poll interval in milliseconds',
+		description: 'Define the balena API poll interval in milliseconds',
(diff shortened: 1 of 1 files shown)


src/lib/env-vars.ts
```diff
@@ -76,12 +76,12 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
...
 	RESIN_SUPERVISOR_LOG_CONTROL: {
 		enum: ['false', 'true'],
-		description: 'Enable / Disable logs from being sent to Resin',
+		description: 'Enable / Disable logs from being sent to balena',
 		default: 'true',
 	},
 	RESIN_SUPERVISOR_POLL_INTERVAL: {
 		type: 'integer',
-		description: 'Define the Resin API Poll interval in milliseconds',
+		description: 'Define the balena API Poll interval in milliseconds',
```
dfunckt: Might as well change `Poll` to `poll` while you're at it.

richbayliss: ^^ This. good spot man

src/lib/env-vars.ts
```diff
@@ -94,7 +94,7 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
 	RESIN_SUPERVISOR_PERSISTENT_LOGGING: {
 		enum: ['false', 'true'],
 		description:
-			'Enable persistent logging. Only supported by resin-supervisor >= v7.15.0.',
+			'Enable persistent logging. Only supported by balena-supervisor >= v7.15.0.',
```
alisondavis17: I guess technically this should be "only supported by balena and resin supervisors >= v7.15.0" since there is no "balena" supervisor v 7.15.0

PR #28, opened 2018-11-09
Update package-lock.json

Page-:
Change-type: patch


package.json
```diff
@@ -76,9 +76,9 @@
   },
   "devDependencies": {
     "copyfiles": "^2.1.0",
-    "husky": "^0.14.3",
+    "husky": "^1.1.3",
     "lint-staged": "^8.0.4",
-    "prettier": "^1.14.3",
+    "prettier": "^1.15.1",
```
dfunckt: Can you run `npm run prettify` and include any changes? I just updated Prettier on resin-api and it caused a small diff.

Page-: done

PR #29, opened 2018-11-09
env-vars: Update persistent logging description

thgreasi:
As requested by @alisondavis17 in #26

Change-type: patch
See: https://github.com/balena-io/open-balena-api/pull/26#discussion_r231946843

### DIFF
--- a/src/lib/env-vars.ts
+++ b/src/lib/env-vars.ts
@@ -94,7 +94,7 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
 	RESIN_SUPERVISOR_PERSISTENT_LOGGING: {
 		enum: ['false', 'true'],
 		description:
-			'Enable persistent logging. Only supported by balena-supervisor >= v7.15.0.',
+			'Enable persistent logging. Only supported by supervisor versions >= v7.15.0.',
 		default: 'false',
 	},
 };


src/lib/env-vars.ts
```diff
@@ -94,7 +94,7 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
 	RESIN_SUPERVISOR_PERSISTENT_LOGGING: {
 		enum: ['false', 'true'],
 		description:
-			'Enable persistent logging. Only supported by balena-supervisor >= v7.15.0.',
+			'Enable persistent logging. Only supported by balena and resin supervisors >= v7.15.0.',
```
dfunckt: Could we not just say "Only supported by Supervisor >= v7.15.0", otherwise I think it's confusing and begs for a link to a longer description.

thgreasi: SGTM, do you insist on this being Title case?
How about:
```suggestion
			'Enable persistent logging. Only supported by supervisor versions >= v7.15.0.',
```
or
```suggestion
			'Enable persistent logging. Only supported by supervisors >= v7.15.0.',
```

alisondavis17: @thgreasi I like "supervisor versions" I think that makes sense and is nice and simple

alisondavis17: (and not title case)

dfunckt: I don't feel strongly about the case. Lowercase can be mistaken to something generic that has the role of a supervisor. Title-case refers to the specific thing we call "Supervisor" by name.

What I feel strongly about is the plural we're currently using -- I think it's incorrect -- so "supervisor versions" or "Supervisor versions" are both good for me.

PR #45, opened 2019-01-08
Add support for application, device and release tags

dfunckt:
Fixes https://github.com/balena-io/open-balena/issues/28

Change-type: minor


src/hooks/resources/tags.ts
```diff
@@ -0,0 +1,20 @@
...
+	const tagHook: sbvrUtils.Hooks = {
+		POSTPARSE: ({ request }) => {
+			if (request.values[nameProp] != null) {
+				checkTagKeyValidity(request.values[nameProp]);
+			}
+		},
+	};
+
+	sbvrUtils.addPureHook('POST', 'resin', resource, tagHook);
+	sbvrUtils.addPureHook('PUT', 'resin', resource, tagHook);
```
thgreasi: Ideally only the tag value should be updated, so we might want to also run the `tagHook` for PATCH requests as well in open-balena-api.

PR #47, opened 2019-01-10
config: Embed device OpenVPN config

hedss:
We've currently got multiple sources of truth for
the OpenVPN config. This change ensures there's a single
point which is dynamically filled in for the appropriate
VPN endpoints.

Connects-to: #46
Change-type: patch
Signed-off-by: Heds Simons <heds@balena.io>

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1,9 +1,8 @@
 import * as _ from 'lodash';
-import * as memoizee from 'memoizee';
 
 const openVpnConfig = `
 client
-remote VPN_DETAILS
+remote {{VPN_DETAILS}}
 resolv-retry infinite
 
 remote-cert-tls server
@@ -55,15 +54,6 @@ export function intVar<R>(varName: string, defaultValue?: R): number | R {
 	return i;
 }
 
-export const openVpnVar = memoizee(
-	(): string => {
-		return openVpnConfig.replace(
-			'remote VPN_DETAILS',
-			`remote ${requiredVar('VPN_HOST')} ${requiredVar('VPN_PORT')}`,
-		);
-	},
-)();
-
 export const API_HOST = requiredVar('API_HOST');
 export const API_VPN_SERVICE_API_KEY = requiredVar('API_VPN_SERVICE_API_KEY');
 export const AUTH_RESINOS_REGISTRY_CODE =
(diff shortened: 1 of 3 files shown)


hedss:
Have tested and works.

src/lib/config.ts
```diff
@@ -1,10 +1,38 @@
...
+user openvpn
+group openvpn
+`;
 
 export const requiredVar = (varName: string): string => {
 	const s = process.env[varName];
 	if (s == null) {
 		console.error(`Missing environment variable: ${varName}`);
-		process.exit(1);
+		process.exitCode = 1;
```
wrboyce: Nice, didn't know you could do this.

src/lib/config.ts
```diff
@@ -1,10 +1,38 @@
 import * as _ from 'lodash';
+import * as memoizee from 'memoizee';
+
+const openVpnConfig = `
+client
+remote VPN_DETAILS
```
dfunckt: Can you please make it more apparent that this is a placeholder that will be replaced later on? Maybe wrap it in curly braces or percent signs (`{{ VPN_DETAILS }}` or `%%VPN_DETAILS%%`)?

src/lib/config.ts
richbayliss: If we are changing the `exit()` call, then I think it would be better to throw the missing environment variable error instead of the `Unreachable` error. This looks like we should expect an error, followed by a random "Unreachable" exception message.

src/lib/config.ts
```diff
@@ -28,6 +56,15 @@ export function intVar<R>(varName: string, defaultValue?: R): number | R {
 	return i;
 }
 
+export const openVpnVar = memoizee(
```
dfunckt: You don't really need all that -- isn't plain assignment working?

src/lib/config.ts
hedss: Yep, it's the best way of exiting Node so that any async ops still outstanding (file closures, etc.) happen.

src/lib/config.ts
hedss: I've done it purely to ensure the replace doens't get called every time.

dfunckt: Agreed, there's no point invoking the replace more than once, but doesn't the following work as well?

```
export const openVpnVar = openVpnConfig.replace(
	'remote VPN_DETAILS',
	`remote ${requiredVar('VPN_HOST')} ${requiredVar('VPN_PORT')}`,
);
```

hedss: It does, I've actually assigned it properly now. :)

PR #49, opened 2019-01-16
✨ Add redis prefix for rate limiting

afitzek:
This scopes the rate limit entries for redis
into a prefix called `api:ratelimiting:`.

Change-type: minor


dfunckt:
VB apparently doesn't like emojis.

PR #52, opened 2019-01-25
Add connectivity-check endpoint that just sends a 204

Page-:
Change-type: patch

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -57,6 +57,14 @@ export function setup(app: _express.Application, options: SetupOptions) {
 
 	app.disable('x-powered-by');
 
+	app.use('/connectivity-check', (_req, res) => {
+		res.status(204);
+		// Remove some automatically added headers, we want the response to be as small as possible
+		res.removeHeader('ETag');
+		res.removeHeader('Date');
+		res.end();
+	});
+
 	if (SENTRY_DSN != null) {
 		Raven.config(SENTRY_DSN, {
 			captureUnhandledRejections: true,


Page-:
@majorz it will add a `ETAG`, `Date`, and `Connection: keep-alive` headers I believe, which are automatically added by express, I can potentially get rid of those but it'll be more involved as there's no easy way to turn them off for a single endpoint

Page-:
Or at least it will add those if a browser sends the request, I'm not certain for network-manager as it may depend on the request headers that are sent

majorz:
@Page- Will `res.removeHeader` be helpful in purging those? The docs read "Removes a header that's queued for implicit sending."

I looked at the ping output and I see some additional headers applied also:
```
$ curl --verbose http://api.resin.io/ping
*   Trying 18.235.93.134...
* TCP_NODELAY set
* Connected to api.resin.io (18.235.93.134) port 80 (#0)
> GET /ping HTTP/1.1
> Host: api.resin.io
> User-Agent: curl/7.63.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Tue, 29 Jan 2019 10:19:33 GMT
< Content-Type: text/html; charset=utf-8
< Content-Length: 2
< Connection: keep-alive
< X-Frame-Options: DENY
< X-Content-Type-Options: nosniff
< ETag: W/"2-nOO9QiTIwXgNtWtBJezz8kv3SLc"
< 
* Connection #0 to host api.resin.io left intact
```

If this does not work and there is not an easy way, I think we can leave it like it is.

majorz:
I should have mentioned that NM uses libcurl for making those requests and does not preserve state across them. So need to have caching support or keeping the socket alive.

PR #55, opened 2019-01-28
access: lower host OS access to OS version 2.0.0

imrehg:
Was discussed that there's no compelling reason to tie the host OS access (approximately) to the first hostapp enabled OS versions, but it can be lowered to earliest 2.x. Those OS versions have read-only root as well by default, so harder to mess things up (a bit). It's also required for an ongoing FleetOps task, giving some users access to their devices to do maintenance.

As a note, likely the variable's name should change as well (from `..._RESINOS_...` to `..._HOST_OS_..` or just `..._OS_`). If a form is agreed to, will add that change as an extra commit to this PR, as logically separate.

Change-type: minor

### DIFF
--- a/src/routes/access.ts
+++ b/src/routes/access.ts
@@ -8,7 +8,7 @@ import { captureException } from '../platform/errors';
 
 const { UnauthorizedError } = sbvrUtils;
 
-const HOSTOS_ACCESS_MIN_RESINOS_VER = '2.7.5';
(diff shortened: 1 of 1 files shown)


richbayliss:
@imrehg I don't think we will solve the tests story in time for this to miss your requirement time. I am happy to merge it as-is, but you'll need to update it first (master has moved on) -- once done I will double check and approve it for you dude.

PR #57, opened 2019-02-14
added `skipHttpsPaths` to setup options

dt-rush:
Change-type: patch
Signed-off-by: Nick Payne <nickp@balena.io>

This will be used by `resin-api` to pass `/metrics` as a path which should not be redirected, as it is going to be hit by Prometheus instances sitting inside the VPC network, not going through the https-terminating ELB. (We could have simply added `X-Forwarded-Proto: https`, but Prometheus doesn't allow the setting of outgoing headers).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -39,6 +39,7 @@ export interface SetupFunction {
 export interface SetupOptions {
 	configPath: string; // must be absolute or relative to `process.cwd()`
 	version?: string; // this will be reported along with exceptions to Sentry
+	skipHttpsPaths?: string[]; // a list of paths which should be exempt from https redirection
 
 	onInit?: SetupFunction;
 	onInitMiddleware?: SetupFunction;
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -66,7 +67,7 @@ export function setup(app: _express.Application, options: SetupOptions) {
 
 	// redirect to https if needed, except for requests to
 	// the /ping endpoint which must always return 200
-	app.use(fixProtocolMiddleware(['/ping']));
+	app.use(fixProtocolMiddleware(['/ping'].concat(options.skipHttpsPaths || [])));
```
mikesimos: Looks like this line needs to be reformated with prettier.

src/index.ts
```diff
@@ -66,7 +67,9 @@ export function setup(app: _express.Application, options: SetupOptions) {
 
 	// redirect to https if needed, except for requests to
 	// the /ping endpoint which must always return 200
-	app.use(fixProtocolMiddleware(['/ping']));
+	app.use(
+		fixProtocolMiddleware(['/ping'].concat(options.skipHttpsPaths || [])),
```
dfunckt: You don't need `|| []` really -- TS won't let anyone set `skipHttpsPaths` to `null`.

dt-rush: Good point, I'll set it to be a ? param, since it shouldn't force any existing code to add the param.

Issue #58, opened 2019-02-18
PATCHing device.should_be_running__release returns 404 where there is no filtered target


thgreasi:
Doing a request like the following returns an "OK"
```
await sdk.pine.patch({
	resource: `device`,
	options: {
		$filter: {
			belongs_to__application: 987654321,
		}
	},
	body: {
		note: 'asdf'
	}
})
```
(where 987654321 a non-existing app id)
while the following (& while using a valid release id)
```
releaseId = 609327;
await sdk.pine.patch({
	resource: `device`,
	options: {
		$filter: {
			belongs_to__application: 987654321,
		}
	},
	body: {
		should_be_running__release: releaseId
	}
})
```
rejects with
```
body: "Filter arrays must have at least 1 elements, got: []",
statusCode: 404,
```

See: https://github.com/balena-io/resin-api/issues/1747

PR #59, opened 2019-02-18
Fix setting release to zero filtered devices 404 error

thgreasi:
Change-type: patch
Resolves: #58
See: https://github.com/balena-io/resin-api/pull/1749


thgreasi:
I think that there are a few extra places that such a test should be added.
I can open a follow up PR for those as soon as I get some test cases ready.

PR #65, opened 2019-03-18
Add "build log" to "release" resource

sradevski:
Change-type: minor

### DIFF
--- a/src/migrations/00003-add-release-build-log.sql
+++ b/src/migrations/00003-add-release-build-log.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "release" ADD COLUMN IF NOT EXISTS "build log" TEXT NULL;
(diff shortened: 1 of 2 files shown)


src/migrations/00003-add-release-build-log.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "release" ADD COLUMN "build log" TEXT NULL;
```
dfunckt: It's good to be defensive and only attempt to add the column if it does not already exit, i.e. `ALTER TABLE "release" ADD COLUMN IF NOT EXISTS "build log" TEXT NULL;`

sradevski: @dfunckt Makes sense, updated.

PR #67, opened 2019-03-26
Add typings to the device state endpoint implementation

thgreasi:
I was investigating this code as part of some work on the SDK and ended up trying to type it.
I imported the models from the SDK and modified a bit to only contain the open-balena-api props.
I also got some helper methods that we use on the UI/SDK for navigating between resources that are connected w/ FKs.
With the above, I managed to remove all `AnyObject` uses from the state endpoint implementation.
Have added a TODO to one place that the typings don't agree with what we are already doing, but I thought that this PR could be just adding typings and we can fix that in a separate one (if it indeed is wrong).

Resolves: #66
Change-type: patch

### DIFF
--- a/src/lib/device-state.ts
+++ b/src/lib/device-state.ts
@@ -4,6 +4,15 @@ import * as semver from 'resin-semver';
 
 import { DEFAULT_SUPERVISOR_POLL_INTERVAL } from './env-vars';
 
+import {
+	Application,
+	Device,
+	Image,
+	getExpanded,
+	Release,
+	PineDeferred,
+	ServiceInstall,
+} from '../models';
 import { PinejsClient } from '../platform';
 
 // Set RESIN_SUPERVISOR_POLL_INTERVAL to a minimum of 10 minutes
@@ -19,22 +28,23 @@ export const setMinPollInterval = (config: AnyObject) => {
 
 export const getReleaseForDevice = (
 	api: PinejsClient,
-	device: AnyObject,
+	device: Device,
 	singleContainer = false,
-): Promise<AnyObject> => {
-	if (device.should_be_running__release[0] != null) {
-		return Promise.resolve(device.should_be_running__release[0]);
+): Promise<Release> => {
+	const release = getExpanded(device.should_be_running__release);
+	if (release != null) {
+		return Promise.resolve(release);
 	} else {
-		const app = device.belongs_to__application[0];
+		const app = getExpanded(device.belongs_to__application)!;
 		return releaseFromApp(api, app, singleContainer);
 	}
 };
 
 export const releaseFromApp = (
 	api: PinejsClient,
-	app: AnyObject,
+	app: Application,
 	singleContainer = false,
-) => {
+): Promise<Release> => {
 	let containsImgObj = {};
 	if (singleContainer) {
 		containsImgObj = { $top: 1 };
@@ -72,25 +82,32 @@ export const releaseFromApp = (
 				},
 			},
 		})
-		.then(([release]: AnyObject[]) => release);
+		.then(([release]: Release[]) => release);
 };
 
 export const serviceInstallFromImage = (
-	device: AnyObject,
-	image?: AnyObject,
-): undefined | AnyObject => {
+	device: Device,
+	image?:
+		| Image
+		| (Pick<Image, Exclude<keyof Image, 'is_a_build_of__service'>> & {
+				is_a_build_of__service: number;
+		  }),
+): ServiceInstall | undefined => {
 	if (image == null) {
 		return;
 	}
 
 	let id: number;
 	if (_.isObject(image.is_a_build_of__service)) {
-		id = image.is_a_build_of__service.__id;
+		id = (image.is_a_build_of__service as PineDeferred).__id;
 	} else {
-		id = image.is_a_build_of__service;
+		id = image.is_a_build_of__service as number;
 	}
 
-	return _.find(device.service_install, si => si.service[0].id === id);
+	return _.find(
+		device.service_install,
+		si => getExpanded(si.service)!.id === id,
+	);
 };
 
 export const formatImageLocation = (imageLocation: string) =>
--- a/src/models/resources.ts
+++ b/src/models/resources.ts
@@ -1,0 +1,206 @@
+import {
+	PineResourceBase,
+	NavigationResource,
+	ReverseNavigationResource,
+} from './utils';
+
+export interface User extends PineResourceBase {
+	username: string;
+	password: string;
+	jwt_secret: string;
+}
+
+export interface Application extends PineResourceBase {
+	app_name: string;
+	device_type: string;
+	slug: string;
+	commit: string;
+	should_track_latest_release: boolean;
+
+	application_type: NavigationResource<ApplicationType>;
+	user: NavigationResource<User>;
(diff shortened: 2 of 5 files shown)


src/lib/device-state.ts
```diff
@@ -72,25 +82,32 @@ export const releaseFromApp = (
...
+		  }),
+): ServiceInstall | undefined => {
 	if (image == null) {
 		return;
 	}
 
 	let id: number;
 	if (_.isObject(image.is_a_build_of__service)) {
-		id = image.is_a_build_of__service.__id;
+		id = (image.is_a_build_of__service as PineDeferred).__id;
```
thgreasi: For such cases, in the UI we use a `getId` method that works both on PineDeferred and expanded resources. I can bring this in here as well if you also like it:
```
export const getId = (obj: BalenaSdk.NavigationResource) =>
	getExpandedId(obj) || getDeferredId(obj);

export const getExpanded = <T>(obj: BalenaSdk.NavigationResource<T>) =>
	(_.isArray(obj) && obj[0]) || undefined;

export const getExpandedProp = <T, K extends keyof T>(
	obj: BalenaSdk.NavigationResource<T>,
	key: K,
) => (_.isArray(obj) && obj[0] && obj[0][key]) || undefined;

export const getExpandedId = (obj: BalenaSdk.NavigationResource) =>
	getExpandedProp(obj, 'id');

// custom type guard function
export const isPineDeferred = (obj: any): obj is BalenaSdk.PineDeferred =>
	obj && '__id' in obj;

// function Typing overloads
export function getDeferredId(obj: BalenaSdk.PineDeferred): number;
export function getDeferredId(
	obj: BalenaSdk.NavigationResource,
): number | undefined;
export function getDeferredId(obj: BalenaSdk.NavigationResource) {
	return (isPineDeferred(obj) && obj.__id) || undefined;
}
```

src/lib/device-state.ts
```diff
@@ -19,22 +28,23 @@ export const setMinPollInterval = (config: AnyObject) => {
 
 export const getReleaseForDevice = (
 	api: PinejsClient,
-	device: AnyObject,
+	device: Device,
```
thgreasi: Changing the types of the method args in this file, will cause the resin-api typings to fail since they will still be passing `AnyObject` is some cases. I'm not sure how you prefer to handle this
* Require the actual type and the resin-api will have to update its typings. If we go this way, we should probably change this to be a minor change type, (which since we are on < v1.0.0, it is typically allowed to contain some breaking changes).
* Use `AnyObject | Device` so that pre-existing consumer projects still work.
* <your reviewer suggestion here>

src/models/resources.ts
```diff
@@ -0,0 +1,206 @@
+import {
+	PineResourceBase,
+	NavigationResource,
+	ReverseNavigationResource,
+} from './utils';
+
+export interface User extends PineResourceBase {
```
thgreasi: A great question here is whether we would prefer splitting all these interfaces in separate files.
Having a single huge one might arguably make it slower to find what you are searching.
I have the same dilemma in the SDK, but I'm grabbing the opportunity to discuss this in here.
PS: I'm in favor of slitting into separate files.

src/routes/devices.ts
```diff
@@ -569,67 +634,77 @@ export const state: RequestHandler = (req, res) => {
...
 									const svcInstall =
 										image == null
 											? null
 											: serviceInstallFromImage(depDev, image);
 
 									if (image != null) {
 										varListInsert(
-											image.image_environment_variable,
+											// TODO: investigate whether this is a bug
+											((image as any) as NonNullable<
```
thgreasi: @CameronDiver can you clarify whether this is correct?
I think that image_environment_variable doesn't have an FK to an image
```
Fact type: image is part of release
	Synonymous Form: release contains image
	Term Form: release image

	Fact type: release image has name (Auth)
		Term Form: image environment variable
		Database Table Name: image environment variable
```
image

CameronDiver: Hmm I'm not sure the image environment variables are actually used anywhere, but the FK is from `image__is_part_of__release` and `image_environment_variable`

Page-: Fixed in https://github.com/balena-io/open-balena-api/pull/109

thgreasi: @Page- how about the rest of the PR ? :)
Should I rebase or do you prefer to close it?

Page-: I think it's worthwhile to rebase (although it might be easier to redo the types on the state endpoint specifically because it's gone through a lot of changes), I would ditch cases where we know the expand is safe though, eg for
```ts
depDev.belongs_to__application[0].id
```
vs
```ts
getExpanded(depDev.belongs_to__application)!.id
```
I'd keep the former rather than introducing the `!`

src/models/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+export type NavigationResource<T = WithId> = T[] | PineDeferred;
+
+/**
+ * When expanded holds an array, otherwise the property is not present.
+ * Selecting is not suggested,
+ * in that case it holds a deferred to the original resource.
+ */
+export type ReverseNavigationResource<T = WithId> = T[] | undefined;
+
+export const getExpanded = <T>(obj: NavigationResource<T>) =>
```
Page-: I think if you use
```suggestion
export function getExpanded <T>(obj: T[]): T | undefined
export function getExpanded (obj: NavigationResource<T>): undefined
export function getExpanded <T>(obj: NavigationResource<T>): T | undefined {
```
then you get a more accurate typing and it can catch cases where you're passing in a PineDeferred and it's impossible for the result to ever be anything other than `undefined`, eg
```ts
const x = getExpanded(pineDeferred);
if (x != null) {
	// typeof x === never
}
```
Unfortunately we cannot rely on `<T>(obj: T[]): T` because we might pass in `[]` which would still return undefined

thgreasi: I think that what you suggest is done by:
```ts
export function getExpanded <T>(obj: PineDeferred): undefined;
export function getExpanded <T>(obj: NavigationResource<T>): T | undefined;
export function getExpanded <T>(obj: NavigationResource<T>): T | undefined {
	return (_.isArray(obj) && obj[0]) || undefined;
}
```

src/routes/devices.ts
thgreasi: @Page- maybe we could do something like
```
type PropsOfType<T, P> = {
	[K in keyof T]: T[K] extends P ? K : never;
}[keyof T];

type AssociatedResource<T = WithId> =
	| NavigationResource<T>
	| ReverseNavigationResource<T>;

type ExpandableProps<T> = PropsOfType<T, AssociatedResource>;

export type ExpandedProps<T, U extends ExpandableProps<T>> = T & Required<Pick<T, U>>;
```
?

thgreasi: I mean, since the `Device` props are typed with `NavigationResource` and `ReverseNavigationResource` we somehow have to tell TS that `depDev` has them expanded...
On the other hand, if we get those types in PineClient, then we could have pine itself return the proper result by inspecting the provided props.
The SDK atm does check such props but just for validating the PineOptions. The next step of course is to return a result that is exactly typed as the PineOptions.

thgreasi:
@Page- I've rebased this, so let's discuss at some point (eg on a Friday) how we can make accessing props that are known to be expanded more convenient.

PR #68, opened 2019-03-27
Use a readTransaction for device state

Page-:
Change-type: patch


src/routes/devices.ts
```diff
@@ -276,370 +277,372 @@ export const state: RequestHandler = (req, res) => {
 		return res.send(400);
 	}
 
-	const resinApiTx = resinApi.clone({ passthrough: { req } });
+	db.readTransaction!(tx => {
```
dfunckt: This is hacky IMO and unnecessary, we alredy export wrapInTransaction/runInTransaction helpers from platform/index and I’d rather we did the same for read only transactions, where we also handle the case “readTransaction” being nil due to being unsupported by the current db driver.

Page-: The `runInTransaction` is the real hack, I'm not sure why it was added but all it does is add unnecessary function creation - I've added a commit to remove it and standardise on the direct version.  As for the `nil` case it's impossible for postgres/mysql, it's only if we were to somehow get the api to target websql that it'd cause issues (also btw this would be the most minor of issues for that, the bigger one is that $expand is only supported on postgres).
It would be great if the specialised version of the db type could automatically cascade through everything and remove the need for the ! since I'm not a fan of them either but it'd be very in-depth typings if even possible since the type of the db comes from prefix of the `DATABASE_URL` env var

dfunckt: Agreed, I don’t mind removing runTransaction completely instead, my point is mostly that both forms of transactions should be via a similar api.

I’m personally strongly against force-unwrapping the optional here and if we can’t assert that statically and point the developer towards the reason it is nil, a better option would to just make the attribute a non-optional and either do a normal transaction in the backends that don’t support readxonly transactions or just call process.exit in there so it’s caught early.

thgreasi: I don't have a strong preference on having a separate `runInTransaction` fn or not, but I would prefer to not repeat the `!` in every use of `db.readTransaction`.
Let's check whether `readTransaction` is available inside `platform/index.ts` and export non nullable property for it, using something like:
```
const { dbBase } = sbvrUtils;

if (!dbBase.readTransaction) {
	throw new Error('Open-balena-api requires a DB that supports read transactions!');
}

const db = dbBase as typeof dbBase &  Required<Pick<typeof dbBase, 'readTransaction'>>;

export { db };
```

Page-: I really like that compromise, it avoid the need to check or ignore the nil each time and fits in fairly nicely.  It does suffer from the potential that you can do `sbvrUtils.db.readTransaction = null` and not be stopped by the compiler.. but I doubt that would happen accidentally :P

PR #69, opened 2019-04-01
Add an index for the vpn's service instance lookup

Page-:
Change-type: patch


dfunckt:
@Page- we need a migration as well.

Page-:
done @dfunckt

PR #71, opened 2019-04-02
device-config: allow devices going back to v1.2.1 to use registry v2

imrehg:
That OS version and upwards have docker with capability to pull from registry v2.

This shouldn't really affect the open API, as the device on the open one should not go back to this old version. The balenaCloud related changes and relevant tests will be added in the other relevant API repo.

Change-type: minor

### DIFF
--- a/src/lib/device-config.ts
+++ b/src/lib/device-config.ts
@@ -41,10 +41,10 @@ export const generateConfig = (
 ) => {
 	const userPromise = getUser(req);
 
-	// Devices running ResinOS >= 2.0 can use Registry v2
-	const registryHost = resinSemver.satisfies(osVersion, '<2.0.0')
-		? REGISTRY_HOST
-		: REGISTRY2_HOST;
+	// Devices running ResinOS >=1.2.1 are capable of using Registry v2, while earlier ones must use v1
+	const registryHost = resinSemver.gte(osVersion, '1.2.1')
+		? REGISTRY2_HOST
+		: REGISTRY_HOST;
 
 	const apiKeyPromise = Promise.try(() => {
 		// Devices running ResinOS >= 2.7.8 can use provisioning keys


src/lib/device-config.ts
```diff
@@ -41,10 +41,10 @@ export const generateConfig = (
 ) => {
 	const userPromise = getUser(req);
 
-	// Devices running ResinOS >= 2.0 can use Registry v2
-	const registryHost = resinSemver.satisfies(osVersion, '<2.0.0')
-		? REGISTRY_HOST
-		: REGISTRY2_HOST;
+	// Devices running ResinOS >= 1.2.1 can use Registry v2
+	const registryHost = resinSemver.satisfies(osVersion, '>=1.2.1')
```
Page-: This returns false for `2.0.0-betaX` which would mean those versions would suddenly use `REGISTRY_HOST`, this should be
```suggestion
	const registryHost = resinSemver.satisfies(osVersion, '<1.2.1')
		? REGISTRY_HOST
		: REGISTRY2_HOST;
```
instead

imrehg: @Page- what the heck, I **did not** expect that, but checks out.

```
> let resinSemver = require('resin-semver');
undefined
> let osVersion = '2.0.0-beta1'
undefined
> resinSemver.satisfies(osVersion, '>=1.2.1')
false
```

 Will do the changes as you suggest. I thought it was clearer this original way, but it shouldn't matter much.

imrehg: Went with `.gte` for clarity.

src/lib/device-config.ts
```diff
@@ -41,10 +41,10 @@ export const generateConfig = (
 ) => {
 	const userPromise = getUser(req);
 
-	// Devices running ResinOS >= 2.0 can use Registry v2
-	const registryHost = resinSemver.satisfies(osVersion, '<2.0.0')
-		? REGISTRY_HOST
-		: REGISTRY2_HOST;
+	// Devices running ResinOS >1.2.1 are capable of using Registry v2, while earlier ones must use v1
```
Page-:
```suggestion
	// Devices running ResinOS >=1.2.1 are capable of using Registry v2, while earlier ones must use v1
```

imrehg: Such silly oversight, sorry, @Page- , fixed an pushed.

PR #73, opened 2019-04-05
Add device online state manager

richbayliss:
A mechanism to track a device state more closely based on API state polls. The general process is as follows:
- when a device hits the state `v2` endpoint, mark it as `online`.
- add a delayed message to the Redis-based message queue; delayed to be consumed when the expected next state poll should have happened.
- if the device hits the `v2` state endpoint in the meantime, delete the queued message and re-queue with a fresh delay
- if the device fails to hit the state endpoint before the message is consumed, then mark the device as `timeout` and queue a new message with a 15 sec delay. Again, if the device hit's the state `v2` endpoint this will be removed.
- if this message is consumed, then the device has missed it's window to poll the state `v2` endpoint and is marked `offline`

The state is held in the `device` model in the field `api heartbeat state` as a text enum:
- unknown (default after the migration)
- timeout
- offline
- online


src/platform/middleware.ts
```diff
@@ -73,3 +75,48 @@ export const gracefullyDenyDeletedDevices: RequestHandler = (
...
+			if (!uuid) return;
+
+			// patch the api_heartbeat_state value to the new state...
+			resinApi.patch({
+				resource: 'device',
+				options: {
+					$filter: {
+						uuid,
+						api_heartbeat_state: {
+							$ne: newState,
```
richbayliss: This should mean that an UPDATE query is run, but no rows will match if the device has the same state. It would be good to get @Page- views on this, and if we should make it a SELECT/UPDATE instead.

src/routes/index.ts
```diff
@@ -63,12 +64,14 @@ export const setup = (app: Application) => {
 		'/device/v2/:uuid/state',
 		gracefullyDenyDeletedDevices,
 		apiKeyMiddleware,
+		registerDeviceContact('params.uuid'),
```
richbayliss: This allows the param to be chosen from the request, but this shouldn't really change. I would be happy to make this statically coded to the `:uuid` param if others are in agreement.

src/platform/middleware.ts
```diff
@@ -73,3 +75,48 @@ export const gracefullyDenyDeletedDevices: RequestHandler = (
 			next();
 		});
 };
+
+// register the device state change handler which updates the device model...
+_.once(() => {
```
Page-: You can ditch this `_.once` altogether and just run the code at module level

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,280 @@
...
+		return this.redis
+			.getAsync(`${DeviceOnlineStateManager.REDIS_NAMESPACE}:${uuid}`)
+			.then(value => {
+				const { id, currentState } = JSON.parse(value);
+
+				if (!id || DeviceOnlineStates.Online !== currentState) {
+					this.deviceOnline(uuid);
+				}
+			})
+			.catch(() => noop)
```
Page-: Do you really mean to return a noop function here?

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,280 @@
...
+				if (id) {
+					return this.rsmq
+						.deleteMessage({
+							qname: DeviceOnlineStateManager.EXPIRED_QUEUE,
+							id,
+						})
+						.return();
+				}
+			})
+			.catch(() => noop) // ignore errors when deleting the old queued state and queue the new one...
```
Page-: Do you really mean to return a noop function here?

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,280 @@
...
+							break;
+					}
+				}).then(() =>
+					this.rsmq.deleteMessage({
+						qname: DeviceOnlineStateManager.EXPIRED_QUEUE,
+						id: id,
+					}),
+				);
+			})
+			.catch(NoMessagesToConsumeSignal, noop)
```
Page-: It would be better to wrap this into the `.catch` below because there's a cost to this sugar which could add up since I believe this function is called a lot

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,280 @@
...
+					const { uuid, nextState } = JSON.parse(message);
+
+					// raise and event for the state change...
+					switch (nextState) {
+						case DeviceOnlineStates.Timeout:
+							this.scheduleChangeOfStateForDevice(
+								uuid,
+								DeviceOnlineStates.Timeout,
+								DeviceOnlineStates.Offline,
+								15, // put the device into a timeout state for 15 sec if it misses it's scheduled heartbeat window... then mark as offline
```
Page-: Can we have this as a named constant please

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,280 @@
...
+		// pull a message from the queue...
+		this.rsmq
+			.receiveMessage({
+				qname: DeviceOnlineStateManager.EXPIRED_QUEUE,
+				vt: 30, // prevent other consumers seeing the same message (if any) for 30 secs (prevents multiple API agents from processing it)...
+			})
+			.tap((msg: any) => {
+				// no message! this is thrown to restart the consume process...
+				if (!msg || !msg.id) {
+					delayBeforeNextConsume = 1000;
```
Page-: This could be
```ts
return Promise.delay(1000)
.throw(new NoMessagesToConsumeSignal())
```
and avoid having to have this somewhat hacky var + setTimeout (alternatively the `Promise.delay` can be at the point of `catch` instead of `throw`)

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,280 @@
...
+	rsmq: RedisSMQ;
+	redis: PromisifedRedisClient;
+
+	constructor() {
+		super();
+
+		// create a new Redis client...
+		this.redis = createPromisifedRedisClient({
+			host: process.env.REDIS_HOST,
+			port: parseInt(process.env.REDIS_PORT || '6379'),
```
Page-: This should be fetched/defaulted in the config module (all process.env accesses should be there)

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,280 @@
...
+					if (pollInterval) {
+						return pollInterval.value;
+					}
+
+					return DEFAULT_SUPERVISOR_POLL_INTERVAL;
+				});
+		});
+};
+
+export const getDefaultManager = once(() => {
```
Page-: There is no other manager you can get right?  I think it'd be nicer to do
```suggestion
export const manager = new DeviceOnlineStateManager();
```
and simplify call sites a little (name can be different)

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,280 @@
...
+									},
+								},
+							},
+							name: 'RESIN_SUPERVISOR_POLL_INTERVAL',
+						},
+					},
+				})
+				.then(([pollInterval]: Array<{ value: number }>) => {
+					if (pollInterval) {
+						return pollInterval.value;
```
Page-: This ignores the minimum poll interval that we set (check device-state.ts, `setMinPollInterval`)

typings/rsmq-promise.d.ts
```diff
@@ -0,0 +1,34 @@
+declare module 'rsmq-promise' {
```
Page-: It would be better to update to rsmq 0.10 and remove the need for custom types + manual promisification altogether

richbayliss: I have done this now, but it requires a bit of a nasty hack to make `Promise` a `bluebird` one and not native. Is this desirable?

src/lib/device-online-state/index.ts
richbayliss: I have passed the final value through a `Math.max(...)` and pulled our coded minimum value in.

src/lib/device-online-state/index.ts
richbayliss: Done

src/lib/device-online-state/index.ts
richbayliss: They were already, I just forgot to use them :grin: Done

src/lib/device-online-state/index.ts
richbayliss: I had concerns over this approach as I wasn't sure if the `delay` call allowed the event loop to continue, or if it blocked etc. I have put it in place as you suggested and I will test.

src/lib/device-online-state/index.ts
richbayliss: Done

src/lib/device-online-state/index.ts
richbayliss: Yes. If there is an error clearing the existing value, then it just needs to carry on and set the new one; it isn't an error that we need to worry about.

src/platform/middleware.ts
richbayliss: So if I just call the function directly, then when the `middleware.ts` is imported it will be run... what happens if it's imported twice in different locations? will it be evaluated twice and therefore add 2 event handlers?

src/lib/device-online-state/index.ts
richbayliss: Again, this is getting an ID for a message which is queued for later. If it cannot find one, then it could be an error, but it is more than likely that this device has just come online from being offline for a while. We couldn't determine if this is a real error or not, so just drop it and move on.

src/lib/device-online-state/index.ts
richbayliss: Done

src/platform/middleware.ts
richbayliss: Speaking to Petros, this approach seems the most logical, so unless you have any concerns @Page- then I suggest we stick with it :+1:

src/lib/config.ts
```diff
@@ -55,6 +55,9 @@ export function intVar<R>(varName: string, defaultValue?: R): number | R {
 }
 
 export const API_HOST = requiredVar('API_HOST');
+export const API_HEARTBEAT_STATE_TIMEOUT = intVar(
+	process.env.API_HEARTBEAT_STATE_TIMEOUT || '15',
```
dfunckt: It's generally good to convey the expected unit either on the env var (eg. `API_HEARTBEAT_STATE_TIMEOUT_SECONDS`) or at least as a comment next to the number, because things will get messy if it's set to eg. 15000 by someone thinking this is in milliseconds, as is typical with timeouts.

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,282 @@
+import * as Bluebird from 'bluebird';
```
dfunckt: I don't see any other files in this folder except this `index.ts` one, it's better to move it into the parent directory.

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,282 @@
...
+declare global {
```
dfunckt: Does this literally apply globally or it is limited to file-scope? It should be moved elsewhere if so.

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,282 @@
...
+					},
+				},
+				$filter: {
+					name: 'RESIN_SUPERVISOR_POLL_INTERVAL',
+				},
+			},
+		})
+		.then(([pollInterval]: Array<{ value: number }>) => {
+			if (!pollInterval) {
+				throw new PollIntervalNotFoundError();
```
dfunckt: Why introduce this sort of indirection here? I mean you throw it in here and just catch it and handle it in the catch handler a few lines below.

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,282 @@
...
+		// initialise the RedisSMQ object using our Redis client...
+		this.rsmq = new RedisSMQ({
+			client: this.redis,
+			ns: DeviceOnlineStateManager.REDIS_NAMESPACE,
+		});
+
+		// create the RedisMQ queue and start consuming messages...
+		this.rsmq
+			.createQueueAsync({ qname: DeviceOnlineStateManager.EXPIRED_QUEUE })
+			.catch((err: Error) => {
```
dfunckt: You should be able to remove the type annotation, there may be an issue with typings if not.

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,282 @@
...
+			})
+			.tap((msg: any) => {
+				// no message! this is thrown to restart the consume process...
+				if (!msg || !msg.id) {
+					return Bluebird.delay(1000).thenThrow(
+						new NoMessagesToConsumeSignal(),
+					);
+				}
+			})
+			.then((msg: any) => {
```
dfunckt: This whole block can be simplified down to something like this:

```js
.then(({ id, message }) => {
	const { uuid, nextState } = JSON.parse(message);
	switch (nextState) {
		// ...
	}
	return this.rsmq.deleteMessageAsync({
		// ...
	});
})
```

Any errors thrown by `JSON.parse` or from event handlers will be caught by the next catch handler.

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,282 @@
...
+				const { id, message } = msg;
+
+				return Bluebird.try(() => {
+					const { uuid, nextState } = JSON.parse(message);
+
+					// raise and event for the state change...
+					switch (nextState) {
+						case DeviceOnlineStates.Timeout:
+							this.scheduleChangeOfStateForDevice(
```
dfunckt: You don't seem to wait for this promise to resolve -- is that intended?

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,282 @@
...
+				if (err instanceof NoMessagesToConsumeSignal) {
+					return;
+				}
+
+				captureException(
+					err,
+					'DeviceStateManager ==> Error when consuming queue',
+				);
+			})
+			.then(() => this.consume());
```
dfunckt: `.then` is fine, but I think `.finally` conveys intend slightly better in this case.

src/lib/device-online-state/index.ts
dfunckt: I think @Page- means this should be `.catch(noop)` instead, because currently as it is it will return the `noop` function.

src/lib/device-online-state/index.ts
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	private scheduleChangeOfStateForDevice(
+		uuid: string,
+		currentState: DeviceOnlineStates,
+		nextState: DeviceOnlineStates,
+		inSeconds: number,
```
dfunckt: `inSeconds` is a confusing identifier :) `delaySeconds: number` or even `delay: number // in seconds` seem better.

src/platform/middleware.ts
dfunckt: Nope, import never evaluates a file twice.

src/platform/middleware.ts
```diff
@@ -73,3 +75,45 @@ export const gracefullyDenyDeletedDevices: RequestHandler = (
...
+		const uuid = _.get(req, path, '');
+		if (uuid !== '') {
+			DeviceOnlineState.getPollIntervalForDevice(uuid).then(pollInterval =>
+				DeviceOnlineState.manager.captureEventFor(uuid, pollInterval / 1000),
+			);
+		}
+
+		next();
+	}) as RequestHandler;
+};
```
dfunckt: All this code definitely doesn't belong here -- `platform` is much more generic than that and should know nothing about devices or their online state. In fact, we generally plan to move `platform` out of the API altogether into something more reusable.

Why having this code in `lib/device-online-state.ts` isn't good enough?

src/lib/config.ts
richbayliss: Good call.

src/lib/device-online-state/index.ts
richbayliss: I am not sure @dfunckt I will try to find out, but this is the only way I found to make the RSMQ library work with Bluebird

src/platform/middleware.ts
```diff
@@ -73,3 +75,16 @@ export const gracefullyDenyDeletedDevices: RequestHandler = (
...
 		});
 };
+
+export const registerDeviceContact = (path: string) => {
+	return ((req, _res, next) => {
+		const uuid = _.get(req, path, '');
+		if (uuid !== '') {
+			DeviceOnlineState.getPollIntervalForDevice(uuid).then(pollInterval =>
+				DeviceOnlineState.manager.captureEventFor(uuid, pollInterval / 1000),
+			);
```
Page-: Should probably have a `.catch`/report here rather than potentially an `UnhandledPromiseRejection`

src/platform/middleware.ts
```diff
@@ -73,3 +75,16 @@ export const gracefullyDenyDeletedDevices: RequestHandler = (
 			next();
 		});
 };
+
+export const registerDeviceContact = (path: string) => {
```
Page-: This would be better as
```suggestion
export const registerDeviceContact = (path: _.PropertyPath) => {
```

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,306 @@
...
+	PromisifedRedisClient,
+} from './redis-promise';
+import * as RedisSMQ from 'rsmq';
+import {
+	REDIS_HOST,
+	REDIS_PORT,
+	API_HEARTBEAT_STATE_TIMEOUT_SECONDS,
+} from './config';
+
+export class PollIntervalNotFoundError extends Error {}
```
dfunckt: This is not used anymore and can be removed.

src/platform/middleware.ts
Page-: Can you add a
```ts
path = _.toPath(path)
```
this will resolve the path once and avoid paying that cost every time the returned function is called (only an issue because the function will be called a lot) - the alternative would be to use `registerDeviceContact(['params', 'uuid']),` but it's nicer to avoid there even being the possibility of a performance pitfall

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,306 @@
...
+				},
+			},
+		},
+		body: {
+			api_heartbeat_state: newState,
+		},
+		passthrough: {
+			req: root,
+		},
+	});
```
Page-: This should have a catch handler, to at the least report the error

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,306 @@
...
+				const { id, currentState } = JSON.parse(value);
+
+				if (!id || DeviceOnlineStates.Online !== currentState) {
+					this.deviceOnline(uuid);
+				}
+			})
+			.catch(() => noop)
+			.then(() => {
+				// record the activity...
+				return this.scheduleChangeOfStateForDevice(
```
Page-: You could drop the `{` and use an implicit return if you want

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,306 @@
...
+		return this.redis
+			.getAsync(`${DeviceOnlineStateManager.REDIS_NAMESPACE}:${uuid}`)
+			.then(value => {
+				const { id, currentState } = JSON.parse(value);
+
+				if (!id || DeviceOnlineStates.Online !== currentState) {
+					this.deviceOnline(uuid);
+				}
+			})
+			.catch(() => noop)
```
Page-: Pretty sure this should be
```suggestion
			.catch(noop)
```

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,306 @@
...
+							break;
+					}
+				}).then(() =>
+					this.rsmq.deleteMessageAsync({
+						qname: DeviceOnlineStateManager.EXPIRED_QUEUE,
+						id: id,
+					}),
+				);
+			})
+			.catch((err: Error | NoMessagesToConsumeSignal) => {
```
Page-: I think things should still work without overriding the typing, ie
```suggestion
			.catch((err) => {
```

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,306 @@
...
+							this.deviceOffline(uuid);
+							break;
+						default:
+							// nothing should be raised here...
+							break;
+					}
+				}).then(() =>
+					this.rsmq.deleteMessageAsync({
+						qname: DeviceOnlineStateManager.EXPIRED_QUEUE,
+						id: id,
```
Page-:
```suggestion
						id,
```

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,306 @@
...
+								DeviceOnlineStates.Offline,
+								API_HEARTBEAT_STATE_TIMEOUT_SECONDS, // put the device into a timeout state if it misses it's scheduled heartbeat window... then mark as offline
+							);
+							this.deviceTimeout(uuid);
+							break;
+						case DeviceOnlineStates.Offline:
+							this.deviceOffline(uuid);
+							break;
+						default:
+							// nothing should be raised here...
```
Page-: Maybe worth reporting this if it shouldn't happen?

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,306 @@
...
+			})
+			.tap((msg: any) => {
+				// no message! this is thrown to restart the consume process...
+				if (!msg || !msg.id) {
+					return Bluebird.delay(1000).thenThrow(
+						new NoMessagesToConsumeSignal(),
+					);
+				}
+			})
+			.then((msg: any) => {
```
Page-: Why is the `any` needed?

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,306 @@
...
+	private deviceTimeout(uuid: string) {
+		this.emit('change', uuid, DeviceOnlineStates.Timeout);
+	}
+
+	private consume() {
+		// pull a message from the queue...
+		this.rsmq
+			.receiveMessageAsync({
+				qname: DeviceOnlineStateManager.EXPIRED_QUEUE,
+				vt: 30, // prevent other consumers seeing the same message (if any) for 30 secs (prevents multiple API agents from processing it)...
```
Page-: Can you move this to a constant please

src/lib/config.ts
```diff
@@ -55,6 +55,9 @@ export function intVar<R>(varName: string, defaultValue?: R): number | R {
 }
 
 export const API_HOST = requiredVar('API_HOST');
+export const API_HEARTBEAT_STATE_TIMEOUT_SECONDS = intVar(
+	process.env.API_HEARTBEAT_STATE_TIMEOUT_SECONDS || '15',
```
Page-: This should be
```suggestion
	'API_HEARTBEAT_STATE_TIMEOUT_SECONDS', 15,
```
because it takes the var name as the first param, and an optional default value as the second - right now it was trying to assign `process.env[process.env.API_HEARTBEAT_STATE_TIMEOUT_SECONDS || '15']` to `API_HEARTBEAT_STATE_TIMEOUT_SECONDS`

src/lib/device-online-state.ts
richbayliss: This is more of a "nothing to do here" which could happen if some bad value made it into the system. I will call a `catchException` to get it logged

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,317 @@
...
+															uuid: uuid,
+														},
+													},
+												},
+											},
+										},
+									},
+								},
+							},
+							name: 'RESIN_SUPERVISOR_POLL_INTERVAL',
```
dfunckt: This can also be [named as `RESIN_SUPERVISOR_POLL_INTERVAL`](https://github.com/balena-io/open-balena-api/blob/master/src/lib/env-vars.ts#L31) -- unfortunately I don't think we normalise to a single prefix yet, the [PR that will do this](https://github.com/balena-io/open-balena-api/pull/34) is blocked on translations for a while now. You will need to check for both values, like so:

```
name: {
  $in: ['RESIN_...', 'BALENA_...'],
}
```

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,317 @@
...
+					),
+				)
+		);
+	}
+}
+
+export const manager = new DeviceOnlineStateManager();
+
+// register the device state change handler which updates the device model...
+manager.on('change', (uuid: string, newState: DeviceOnlineStates) => {
```
dfunckt: This is no big deal and it will work fine but it feels a bit odd for an object to register with itself in order to listen to its own events and it isn't all self-contained. What is your concern?

src/platform/middleware.ts
dfunckt: @richbayliss was this marked as resolved by accident? I still think this does not belong here. Can you please address this or discuss if you think otherwise?

src/lib/device-online-state.ts
richbayliss: Good spot mate, totally slipped my mind to cover that but you're absolutely right.

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,317 @@
...
+							`${DeviceOnlineStateManager.REDIS_NAMESPACE}:${uuid}`,
+							delay + 5, // add a small 5 second buffer to allow some wiggle room on top of the timeout window...
+						),
+					),
+			);
+	}
+
+	public captureEventFor(uuid: string, timeoutSeconds: number) {
+		// see if we already have a queued state for this device...
+		return (
```
Page-: What is the paren for?

src/lib/device-online-state.ts
richbayliss: No concerns; the original idea was to have this as an event you could register any code onto, so it would separate the function of the state manager from the code updating the API models. If this isn't something we desire, then moving the code into the main state manager class is fine.

dfunckt: It's not important, ignore my comment.

src/lib/device-online-state.ts
richbayliss: Prettier. :man_shrugging:

src/platform/middleware.ts
```diff
@@ -1,8 +1,8 @@
 import { reqHasPermission, getUser } from './auth';
 import { retrieveAPIKey } from './api-keys';
 import { RequestHandler } from 'express';
-
 import { resinApi, root } from './index';
+import * as _ from 'lodash';
```
dfunckt: Forgotten unused import.

src/routes/index.ts
```diff
@@ -26,6 +30,28 @@ export const loginRateLimiter = createRateLimitMiddleware(
 	{},
 );
 
+export const registerDeviceStateEvent = (pathToUuid: _.PropertyPath) => {
```
dfunckt: Is this used by anyone else? Why export it?

src/platform/middleware.ts
richbayliss: Will remove it

src/routes/index.ts
richbayliss: Not in `open-balena-api` but I was thinking of hooking the V1 state endpoint in `resin-api`

dfunckt: Cool, if it's useful elsewhere, sounds good 👍

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,328 @@
...
+								$any: {
+									$alias: 'a',
+									$expr: {
+										a: {
+											owns__device: {
+												$any: {
+													$alias: 'd',
+													$expr: {
+														d: {
+															uuid: uuid,
```
Page-:
```suggestion
															uuid,
```

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,328 @@
...
+				this.redis
+					.setAsync(
+						`${DeviceOnlineStateManager.REDIS_NAMESPACE}:${uuid}`,
+						JSON.stringify({
+							id,
+							currentState,
+						}),
+					)
+					.then(() =>
+						this.redis.expireAsync(
```
Page-: I think this can be combined into the `setAsync` via
```ts
				this.redis
					.setAsync(
						`${DeviceOnlineStateManager.REDIS_NAMESPACE}:${uuid}`,
						JSON.stringify({
							id,
							currentState,
						}),
						'EX',
						delay + 5
					)

File: src/lib/device-online-state.ts [hunk 42]
```diff
@@ -0,0 +1,328 @@
...
+				const { id } = JSON.parse(value) as { id: string };
+
+				if (id) {
+					return this.rsmq.deleteMessageAsync({
+						qname: DeviceOnlineStateManager.EXPIRED_QUEUE,
+						id,
+					});
+				}
+
+				return 1;
```
Page-: This looks to be unused?
```suggestion
```

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,328 @@
...
+								this.scheduleChangeOfStateForDevice(
+									uuid,
+									DeviceOnlineStates.Timeout,
+									DeviceOnlineStates.Offline,
+									API_HEARTBEAT_STATE_TIMEOUT_SECONDS, // put the device into a timeout state if it misses it's scheduled heartbeat window... then mark as offline
+								);
+								this.deviceTimeout(uuid);
+								break;
+							case DeviceOnlineStates.Offline:
+								this.deviceOffline(uuid);
```
Page-: These generate a promise right?  Should this be
```suggestion
								return this.deviceOffline(uuid);
```
?
If not then you'll need to `return null` to avoid a bluebird warning

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,328 @@
...
+						}
+					}).finally(() =>
+						this.rsmq.deleteMessageAsync({
+							qname: DeviceOnlineStateManager.EXPIRED_QUEUE,
+							id,
+						}),
+					);
+				} else {
+					// no message! this is thrown to restart the consume process...
+					return Bluebird.delay(1000).thenThrow(
```
Page-: There's no need to throw anything is there?  Since you just swallow this specific error anyway so it's equivalent to just doing
```suggestion
					return Bluebird.delay(1000);
```

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,328 @@
...
+			});
+	}
+
+	private updateDeviceModel(uuid: string, newState: DeviceOnlineStates) {
+		console.log(
+			`DeviceStateManager: ${uuid} is ${DeviceOnlineStates[newState]}`,
+		);
+
+		// patch the api_heartbeat_state value to the new state...
+		resinApi
```
Page-: Should this be returned?

src/lib/device-online-state.ts
richbayliss: Without this, a type error pops out:

```
Argument of type '(value: string) => Promise<0 | 1> | undefined' is not assignable to parameter of type '(value: string) => Resolvable<0 | 1>'
```

I will try to refactor it.

richbayliss:
Requires this PR in order to have working tests. https://github.com/balena-io/open-balena-api/pull/143

src/lib/device-online-state.ts
richbayliss: Added type definition to support this, and now using it.

src/lib/device-online-state.ts
richbayliss: I refactored this

src/lib/device-online-state.ts
richbayliss: Refactored

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,318 @@
...
+
+	private updateDeviceModel(uuid: string, newState: DeviceOnlineStates) {
+		// patch the api_heartbeat_state value to the new state...
+		return resinApi
+			.patch({
+				resource: 'device',
+				options: {
+					$filter: {
+						uuid,
+						api_heartbeat_state: {
```
Page-: Nice way of doing this is
```ts
		const body = {
			api_heartbeat_state: newState,
		}
		return resinApi
			.patch({
				resource: 'device',
				options: {
					$filter: {
						uuid,
						$not: body,
					}
				},
				body,
				passthrough: {
					req: root,
				},
			})
...
to only update if any of the update values in the body are different (it's equivalent but a bit more DRY and can be reused as a pattern elsewhere to be recognisable, eg the state endpoint already does this style)

Comment (richbayliss) [reply src/lib/device-online-state.ts]: cool

File: src/lib/device-online-state.ts [hunk 47]
```diff
@@ -0,0 +1,318 @@
...
+		// create the RedisMQ queue and start consuming messages...
+		this.rsmq
+			.createQueueAsync({ qname: DeviceOnlineStateManager.EXPIRED_QUEUE })
+			.catch(err => {
+				if (err.name !== 'queueExists') {
+					throw err;
+				}
+			})
+			.then(() => {
+				this.consume();
```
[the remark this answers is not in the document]
richbayliss: sure

test/02-device-state.ts
```diff
@@ -0,0 +1,340 @@
...
+			}));
+
+	it('Should stay as ONLINE as long as we GET /device/v2/:uuid/state within the poll interval', () => {
+		const pollStateInterval = (POLL_SEC / 3) * 2 * 1000;
+		const pollStateEndpoint = () =>
+			supertest(app, device1Key)
+				.get(`/device/v2/${uuid}/state`)
+				.expect(200);
+
+		return Bluebird.delay(pollStateInterval)
```
Page-:
```suggestion
	for (let i=0; i < 9; i++) {
		await Bluebird.delay(pollStateInterval)
		await pollStateEndpoint()
	}
	expect(stateUpdateSpy.callCount).is.equal(
		0,
		'Do not update the device model if as long as it stays online and polls regularly',
	);
```

richbayliss: Sure - in this case I totally agree

init.ts
```diff
@@ -18,6 +18,11 @@ async function onInitModel() {
 	await db.transaction(tx =>
 		updateOrInsertModel('application_type', filter, insert, tx),
 	);
+	await db.transaction(tx =>
+		tx.executeSql(
+			'ALTER TABLE "device" ALTER COLUMN "api heartbeat state" SET DEFAULT -2;',
```
Page-: This seems like it should be in resin-init.sql

test/test-lib/aws-mock.ts
```diff
@@ -123,5 +123,9 @@ export const AWSSdkMock = {
 	S3: S3Mock,
 };
 
+<<<<<<< HEAD
 mockery.enable({ warnOnUnregistered: false });
+=======
+mockery.enable({ warnOnUnregistered: false, warnOnReplace: true });
+>>>>>>> 36eed5a... tests: Add fixtures framework
```
Page-: Rebase conflict

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,320 @@
...
+							name: {
+								$in: [
+									'BALENA_SUPERVISOR_POLL_INTERVAL',
+									'RESIN_SUPERVISOR_POLL_INTERVAL',
+								],
+							},
+						},
+					},
+				})
+				.then(([pollInterval]: Array<{ value: number }>) => {
```
Page-: There can be potentially 2 results, which of the two should take priority?

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,320 @@
...
+					name: {
+						$in: [
+							'BALENA_SUPERVISOR_POLL_INTERVAL',
+							'RESIN_SUPERVISOR_POLL_INTERVAL',
+						],
+					},
+				},
+			},
+		})
+		.then(([pollInterval]: Array<{ value: number }>) => {
```
Page-: There can be potentially 2 results, which of the two should take priority?

package.json
```diff
@@ -1,106 +1,108 @@
 {
```
richbayliss: This should be tab-indented, so I corrected it here

package.json
```diff
@@ -1,106 +1,108 @@
...
+		"node-uuid": "~1.4.0",
+		"passport": "^0.4.0",
+		"passport-jwt": "^4.0.0",
+		"randomstring": "^1.1.5",
+		"raven": "^2.6.4",
+		"redis": "^2.8.0",
+		"request": "^2.85.0",
+		"resin-device-config": "^5.1.1",
+		"resin-semver": "^1.6.0",
+		"rsmq": "^0.11.0",
```
richbayliss: Added dependency

package.json
```diff
@@ -1,106 +1,108 @@
...
+		"@types/lodash": "^4.14.138",
+		"@types/memoizee": "^0.4.3",
+		"@types/node": "^10.14.18",
+		"@types/node-uuid": "0.0.28",
+		"@types/passport-jwt": "^3.0.2",
+		"@types/randomstring": "^1.1.6",
+		"@types/raven": "^2.5.3",
+		"@types/redis": "^2.8.13",
+		"@types/request": "^2.48.3",
+		"@types/rsmq": "^0.3.27",
```
richbayliss: Added dependency

src/lib/config.ts
```diff
@@ -59,6 +59,10 @@ export function intVar<R>(varName: string, defaultValue?: R): number | R {
 }
 
 export const API_HOST = requiredVar('API_HOST');
+export const API_HEARTBEAT_STATE_TIMEOUT_SECONDS = intVar(
```
richbayliss: This is the grace period; the time allowed from when it has missed it's state poll checkin and when we mark it as offline in the DB. 15 secs default value.

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,320 @@
...
+			api_heartbeat_state: newState,
+		};
+
+		return resinApi
+			.patch({
+				resource: 'device',
+				options: {
+					$filter: {
+						uuid,
+						$not: body,
```
richbayliss: This limits the DB write scope to only update values which are currently different. It is pointless to lock rows for update when nothing with be changed.

.editorconfig
```diff
@@ -10,3 +10,6 @@ insert_final_newline = true
 
 [*.md]
 trim_trailing_whitespace = false
+
+[package*.json]
+indent_style = unset
```
richbayliss: Prevent EditorConfig from messing with the `package.json` and `package-lock.json`

src/lib/device-online-state.ts
richbayliss: Should be resolved now, by ordering the responses and taking the top 1 result 👍

src/lib/device-online-state.ts
richbayliss: Should be resolved now, by ordering the responses and taking the top 1 result 👍

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,337 @@
...
+						currentState,
+					}),
+					'EX',
+					delay + 5,
+				),
+			);
+	}
+
+	public captureEventFor(uuid: string, timeoutSeconds: number) {
+		if (API_HEARTBEAT_STATE_ACTIVE !== 1) {
```
richbayliss: Early return if the feature isn't enabled

src/routes/index.ts
```diff
@@ -23,6 +28,35 @@ export const loginRateLimiter = createRateLimitMiddleware({
 	lifetime: 2 * SECONDS_PER_HOUR, // reset counter after 2 hours (in seconds)
 });
 
+export const registerDeviceStateEvent = (pathToUuid: _.PropertyPath) => {
+	// only register the state event if the feature is active...
+	if (API_HEARTBEAT_STATE_ACTIVE !== 1) {
```
richbayliss: Return an empty middleware if the feature isn't enabled

src/lib/config.ts
```diff
@@ -59,6 +59,14 @@ export function intVar<R>(varName: string, defaultValue?: R): number | R {
 }
 
 export const API_HOST = requiredVar('API_HOST');
+export const API_HEARTBEAT_STATE_ACTIVE = intVar(
+	'API_HEARTBEAT_STATE_ACTIVE',
```
Page-: Can you rename this
```suggestion
	'API_HEARTBEAT_STATE_ENABLED',
```
or
```suggestion
	'ENABLE_API_HEARTBEAT_STATE',
```

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,337 @@
...
+import * as RedisSMQ from 'rsmq';
+import {
+	REDIS_HOST,
+	REDIS_PORT,
+	API_HEARTBEAT_STATE_ACTIVE,
+	API_HEARTBEAT_STATE_TIMEOUT_SECONDS,
+} from './config';
+
+export const getPollIntervalForDevice = (uuid: string): Bluebird<number> => {
+	return resinApi
```
Page-: Can you extract these queries to prepared queries and use parameter aliases please (see routes/devices.ts for examples).  It'll allow them to be handled more efficiently internally (and they should be called regularly enough for that to matter)

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,337 @@
...
+				$filter: {
+					name: {
+						$in: [
+							'BALENA_SUPERVISOR_POLL_INTERVAL',
+							'RESIN_SUPERVISOR_POLL_INTERVAL',
+						],
+					},
+				},
+				$orderby: {
+					name: 'desc',
```
Page-: Can you add a comment as to why this ordering is important please

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,337 @@
...
+								);
+						}
+					})
+						.then(() =>
+							this.rsmq.deleteMessageAsync({
+								qname: DeviceOnlineStateManager.EXPIRED_QUEUE,
+								id,
+							}),
+						)
+						.catch((err: Error) => captureException(err));
```
Page-: Can you add a message to this `captureException` please, so it's easier to trace in sentry

src/lib/device-online-state.ts
Page-: Could you give a try manually removing the brackets here again so it's `return this.redis...` and then reformat with an up to date prettier, I'm hoping prettier has improved this case in the 5 months since

src/routes/index.ts
```diff
@@ -23,6 +28,35 @@ export const loginRateLimiter = createRateLimitMiddleware({
 	lifetime: 2 * SECONDS_PER_HOUR, // reset counter after 2 hours (in seconds)
 });
 
+export const registerDeviceStateEvent = (pathToUuid: _.PropertyPath) => {
```
Page-:
```suggestion
export const registerDeviceStateEvent = (pathToUuid: _.PropertyPath): RequestHandler => {
```
and remove the `as RequestHandler`s within the function

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,340 @@
...
+		$orderby: {
+			// we want the last value that would have been passed
+			// to the supervisor, as that is the one it would have used.
+			name: 'desc',
+		},
+	},
+});
+
+export const getPollInterval = (uuid: string): Bluebird<number> => {
+	return getPollIntervalForDevice({ uuid }, undefined, { req: root })
```
Page-: You can add the `{ req: root }` to the `prepare` call rather than needing to pass it explicitly each time

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,340 @@
...
+		.then(([pollInterval]: Array<{ value: number }>) => {
```
Page-: This would be nicer as
```suggestion
		.then((pollIntervals: Array<{ value: number }>) => {
			if (pollIntervals.length > 0) {
				return pollIntervals;
			}
```
to avoid the need to destructure and then restructure the array

src/lib/device-online-state.ts
```diff
@@ -0,0 +1,340 @@
...
+			}
+
+			return getPollIntervalForParentApplication({ uuid }, undefined, {
+				req: root,
+			});
+		})
+		.then(([pollInterval]: Array<{ value: number }>) => {
+			return pollInterval ? pollInterval.value : 0;
+		})
+		.then(pollInterval => {
```
Page-: This should be combined with the previous `.then`

test/test-lib/fake-device.ts
```diff
@@ -0,0 +1,56 @@
...
+			expect(res.body.d[0]).to.have.property('is_for__device_type');
+			expect(res.body.d[0].is_for__device_type)
+				.to.be.an('array')
+				.with.lengthOf(1);
+			expect(res.body.d[0].is_for__device_type[0]).to.have.property('slug');
+
+			return res.body.d[0].is_for__device_type[0].slug as string;
+		});
+
+	const device = await supertest(app, admin)
```
Page-:
```suggestion
	const { body: device } = await supertest(app, admin)
```
and drop the `.then(res => res.body)`

test/test-lib/fake-device.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { app } from '../../init';
+
+import supertest = require('./supertest');
+
+export async function provisionDevice(admin: string, appId: number) {
+	const deviceType = await supertest(app, admin)
+		.get(`/resin/application(${appId})?$expand=is_for__device_type`)
+		.expect(200)
+		.then(res => {
```
Page-: Can you switch this to await/async please (and elsewhere)

test/00-init.ts
```diff
@@ -44,4 +47,9 @@ Promise.resolve(fs.promises.readdir(__dirname))
 			require(`./${fileName}`);
 		});
 	})
+	.then(() =>
+		supertest(app)
```
Page-: What's this for?

src/migrations/00017-add-api-heartbeat-state.sql
```diff
@@ -0,0 +1,5 @@
+-- Simple integer to hold the state of the API heartbeat
+-- values: 1 = online, 0 = offline, -1 = timeout, -2 = unknown
+-- default: -2
+
+ALTER TABLE "device" ADD COLUMN "api heartbeat state" VARCHAR(255) NOT NULL DEFAULT "unknown";
```
Page-: This needs fixing:
```suggestion
ALTER TABLE "device" ADD COLUMN "api heartbeat state" VARCHAR(255) NOT NULL DEFAULT 'unknown';
```

test/00-init.ts
richbayliss: This is needed to load the device types table in the database. Without it, on a single test file run, the application creation will fail with a 404 since the application types cannot be found.

src/lib/device-online-state.ts
richbayliss: I tried, and checked we're using the latest version of `prettier` - it really does want it to look like this 👍

richbayliss:
re-test

nazrhom:
@balena-ci retest

PR #75, opened 2019-04-15
Rename `configPath` to `config` and make sure config.json is valid

Page-:
Also splits index.js to make sure the bulk of it is typed

Update pinejs from 10.8.0 to 10.11.0

Change-type: minor


Page-:
Done @dfunckt, but I'd want to avoid it in any hot paths for now still, until the async/await optimisations are available in the nodejs version we run

dfunckt:
Gotcha, cheers @Page-

PR #76, opened 2019-04-15
Handle `my_applications` within pinejs

Page-:
Depends on #75

### DIFF
--- a/src/abstract-sql-utils.ts
+++ b/src/abstract-sql-utils.ts
@@ -1,0 +1,32 @@
+import { Definition } from '@resin/odata-to-abstract-sql';
+
+import * as _ from 'lodash';
+import { readFileSync } from 'fs';
+import { sbvrUtils } from '@resin/pinejs';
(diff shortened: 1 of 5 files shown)


src/abstract-sql-utils.ts
```diff
@@ -0,0 +1,30 @@
+import { Definition } from '@resin/odata-to-abstract-sql';
+
+import * as _ from 'lodash';
+import { readFileSync } from 'fs';
+import { sbvrUtils } from '@resin/pinejs';
+import { AbstractSqlModel } from '@resin/abstract-sql-compiler';
+
+export const generateAbstractSql = (seModelPath: string): AbstractSqlModel => {
```
richbayliss:
```suggestion
export const generateAbstractSqlModel = (seModelPath: string): AbstractSqlModel => {
```

I have been confused before between an `AbstractSqlModel` and `AbstractSqlQuery` so it would be good to be explicit in the method names, I think.

PR #77, opened 2019-04-17
Update method-override to 3.x

Page-:
Change-type: patch


Page-:
https://github.com/expressjs/method-override/blob/3.0.0/HISTORY.md

Issue #78, opened 2019-04-17
Make the device permissions explicit


richbayliss:
The permission `resin.device.all` is allowing all ports to be available to tunnel unintentionally. This should be broken down to `create`, `read`, `update` and `delete`.

Additionally we would like to keep the ability to tunnel any port, so we will add a new permission `resin.device.tunnel-any` which can be checked for.

Please see https://github.com/balena-io/open-balena/issues/47 for the original issue.

PR #79, opened 2019-04-17
bug: Device permissions oversight

richbayliss:
The permission `resin.device.all` allowed any port to be tunnelled
by mistake, rather than intention. This permission is now broken
down to the component actions create, read, update and delete.

The existing `tunnel-22222` is kept, but also added is a rule to
allow the `tunnel-any` action. It is desirable that any port could
be tunnelled and this permission will be checked by the connect proxy
alongside the explicit port permission.

Change-type: patch
Connects-to: #78


src/lib/auth.ts
```diff
@@ -79,7 +79,11 @@ export const ROLES: {
 		'resin.application_environment_variable.all',
 		'resin.application_tag.all',
 		'resin.application_type.all',
-		'resin.device.all',
+		'resin.device.create',
+		'resin.device.read',
+		'resin.device.update',
+		'resin.device.delete',
+		`resin.device.tunnel-any`,
```
Page-: Why not keep it as `resin.device.all` rather than manually list all current possible permissions?

dfunckt: @Page- I was, as you, under the impression that `all` only granted CRUD actions, I thought it was literally automatically "expanded" by Pine to permissions for this set of 4 actions. But @richbayliss found this is not the case and `all` actually grants all permissions, hence this PR. IMO the first behaviour is more intuitive and I wonder if this is a bug with Pine or just wrong expectations (that we all 3 had :).

Page-: No, I was always fully aware that all means all, otherwise it'd be something like `crud`/`all-http`/etc, otherwise how would you name one that is actually all permissions? "no-really-I-actually-mean-all"? :P

richbayliss: @Page- @dfunckt in this instance, `resin.device.all` would be accurate, but we should therfore drop the `resin.device.tunnel-22222` as this makes it look like only 22222 would be allowed. Agreed?

richbayliss: Done

PR #88, opened 2019-04-29
tests: Add test framework

richbayliss:


package.json
```diff
@@ -1,93 +1,103 @@
...
-  },
-  "engines": {
-    "node": ">=8.0"
-  },
-  "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged"
-    }
-  }
+	"name": "@balena/open-balena-api",
```
Page-: Could you have npm auto-format this please

test/00-init.ts
```diff
@@ -0,0 +1,49 @@
...
+import * as _ from 'lodash';
+import * as fs from 'fs';
+import * as path from 'path';
+
+const testFiles = _(process.env.TEST_FILES)
+	.trim()
+	.split(' ')
+	.map(
+		(fileName): ((f: string) => boolean) => {
+			let [op, ...rest] = fileName;
```
Page-:
```suggestion
			const [op, ...rest] = fileName;
```

automation/fasttest.sh
```diff
@@ -0,0 +1,105 @@
...
+teardown=0
+extra_env=''
+extra_args=''
+
+while [[ $# -gt 0 ]]; do
+	key=$1
+	shift
+	case $key in
+		--without-external)
+			extra_env="${extra_env} --env DB_ANALYTICS_PASSWORD= "
```
Page-: This doesn't exist

package.json
richbayliss: yeah - not sure what happened here, but I was planning to sort it out

automation/common.sh
```diff
@@ -0,0 +1,149 @@
...
+		-e MIXPANEL_ACCOUNT_ID=347777 \
+		-e MIXPANEL_API_KEY=mixpanel_api_key \
+		-e MIXPANEL_API_SECRET=mixpanel_api_secret \
+		-e SIGNUP_EMAIL_DELAY=100 \
+		-e MAILGUN_API_KEY='abc' \
+		-e MAILGUN_DOMAIN='balena.io' \
+		-e MIXPANEL_SIGNUP_EVENT_DELAY=1 \
+		-e INTERCOM_APP_ID=intercom_app_id \
+		-e INTERCOM_SECRET_KEY=intercom_secret \
+		-e ZENDESK_DOMAIN=zendesk_domain \
```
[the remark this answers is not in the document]
richbayliss: done

automation/common.sh
```diff
@@ -0,0 +1,98 @@
...
+		-e DEVICE_CONFIG_SSH_AUTHORIZED_KEYS='ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQCc5cLVsu7b52tzxwa8A6cs4LC2I+IMtyKeVZFVAWdhfEWGEJL78rn0tC4YmHBSWW0hBeZmn8f7ZKcZ/r1W2RpuH7wpUc56/3hffJ0GoktMOX69pjNsE3rWr8/cZbl/LaJjeHBpyFZ6VimYl5KiYyHJH7RZwlEYYiJ2YDCbte51jvmotFTqiOJNPCM3SXWFjsUGDOWwNi7GDxG+ulb0cBnYzyYlHTj5dS/uya0v8S267TYO2jZ9lUvZHYYQBilYoN3qT79B/lRJq/AKsaSPI2VWkwoAWLObJRjQjoIsClxcfaSfRg6fgdezv3saeorVMIdEYAj3CQVmLszCNGC8dgqB9OoKdjd/OUG/bFUc86S+y0dZX3vqKBfZjE1dHNzNkQeulsbRj0FGFTxFYL0Ng7jlr98iXusnNg236LGsBTeod2MLfMRsrOc0Utfc5Pq2Zkyak4aJ6URBFqZnKlJwvI3AFznxvHD1PmIrwzCCmLA409v3WbXEK4KZrUIs81wlQWAK3919ViozdnR23f4hZtq3P5M6AbnMMdIOB1jDjPK//eaTMf3erHVa0nTMMwScBDnrT86yJBBEwpUVpIp6YUBFaVMQscX3gBTd+RyOU9jOvyinGjosJfF38qsMJTNHh48zgic5sSLhtPQgkUjWhpcnfjqrLRewKbVd+VZQc1WsdQ==' \
+		-e IMAGE_MAKER_URL=https://img.resindev.io \
```
Page-: Can you remove any `resin` references please

automation/fasttest.sh
```diff
@@ -0,0 +1,102 @@
...
+			extra_env="${extra_env} --env BLUEBIRD_LONG_STACK_TRACES=1"
+		;;
+		--debug)
+			extra_env="${extra_env} --env DEBUG=1"
+		;;
+		--profile)
+			extra_args="${extra_args} --inspect-brk=0.0.0.0"
+		;;
+		-v | --version)
+			test_versions="$test_versions $1"
```
Page-: This is unused

init.ts
```diff
@@ -81,7 +81,7 @@ async function createSuperuser() {
 		});
 }
 
-const app = express();
+export const app = express();
 app.enable('trust proxy');
 
 setup(app, {
```
Page-: There should be a check in the tests that this has completed before starting to run since otherwise there'll be a race condition

tsconfig.json
```diff
@@ -3,9 +3,5 @@
 	"compilerOptions": {
 		"rootDir": "./"
 	},
-	"include": [
-		"src/**/*.ts",
-		"typings/**/*.d.ts",
-		"*.ts"
-	]
+	"include": ["src/**/*.ts", "typings/**/*.d.ts", "*.ts", "test/**/*.ts"]
```
Page-: Could you put these back on separate lines like before please

init.ts
richbayliss: due to the way the tests are run, they only start after this promise resolves.

test/test-lib/platform.ts
```diff
@@ -0,0 +1,12 @@
+import * as _ from 'lodash';
+export = () => {
+	const platform = require('../../src/platform');
+	const permissions = require('../../src/platform/permissions');
+	return {
+		platform,
+		permissions,
+		catchForeignKeyError: (message: string) => (e: Error) =>
```
Page-: This is unused

test/test-lib/platform.ts
```diff
@@ -0,0 +1,12 @@
+import * as _ from 'lodash';
+export = () => {
+	const platform = require('../../src/platform');
```
Page-: Can you make these typed, either via `import()` or
```ts
import * as _platform from '../../src/platform'
...
const platform: typeof _platform = require('../../src/platform');
```
I think `import()` is nicer but it does turn every call to this into a promise, but that should be fine with await/async

richbayliss:
```
import * as _platform from '../../src/platform';
import * as _permissions from '../../src/platform/permissions';

export const platform: typeof _platform = require('../../src/platform');
export const permissions: typeof _permissions = require('../../src/platform/permissions');
```

test/00-init.ts
```diff
@@ -0,0 +1,49 @@
...
+					return f => f > compareFilename;
+				default:
+					return f => _.startsWith(f, fileName);
+			}
+		},
+	);
+
+const prefixes: Dictionary<true> = {};
+
+Promise.resolve(fs.readdirSync(__dirname))
```
Page-: This can be
```ts
Promise.resolve(fs.promises.readdir(__dirname))
```
to avoid the sync call

richbayliss: @Page- my types don't have `.promises.readdir`

richbayliss: I updated `"@types/node": "^10.14.4"` to provide the extra typing

init.ts
```diff
@@ -92,12 +92,15 @@ setup(app, {
...
 	.tap(createSuperuser)
-	.then(({ startServer, runFromCommandLine }) => {
-		if (process.argv.length > 2) {
-			return runFromCommandLine();
-		}
+	.then(({ startServer }) => {
 		return startServer(process.env.PORT || 1337).return();
 	})
+	.then(() => {
+		if (process.env.DEPLOYMENT === 'TEST') {
```
richbayliss:
```
function runFromMocha(context: AnyObject = global) {
	return ['afterEach', 'after', 'beforeEach', 'before', 'describe', 'it'].every(
		functionName => {
			return context[functionName] instanceof Function;
		},
	);
}
```
then...

```suggestion
		if (runFromMocha()) {
```

richbayliss: @Page- @dfunckt what do you think? We can only run tests IF the process is started via Mocha rather than solely relying on an ENV var...

richbayliss: We can stick the an ENV var for now

circle.yml
```diff
@@ -0,0 +1,31 @@
...
+      - image: library/docker:17
+    working_directory: /tmp/build
+    environment:
+      DOCKER_USERNAME: travisciresin
+      DOCKER_IMAGE: balena/open-balena-api
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run: |
+          docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"
```
dfunckt: We rely here on `set -x` not being in effect and CircleCI not printing the expanded env var which feels somewhat risky as we could end up with the password printed in clear text in the build logs if anything changes.

richbayliss: Good point - this is taken direct from `resin-api` and I don't think we're actually push/pulling any private images - so maybe it can go @dfunckt ?

dfunckt: Ah indeed, that's a good point -- yes, login can be completely removed.

dfunckt: Wait, I think it's CircleCI that pushes the images to DockerHub, not ResinCI, so it can't be removed.

circle.yml
```diff
@@ -0,0 +1,29 @@
+---
+version: 2
+jobs:
+  build:
+    docker:
+      - image: library/docker:17
+    working_directory: /tmp/build
+    environment:
+      DOCKER_USERNAME: travisciresin
```
dfunckt: This isn't needed anymore.

richbayliss: yup yup - let me have a tidy up

PR #89, opened 2019-05-01
Update dependencies

Page-:
Change-type: patch


src/lib/device-state.ts
```diff
@@ -84,7 +84,7 @@ export const serviceInstallFromImage = (
 	}
 
 	let id: number;
-	if (_.isObject(image.is_a_build_of__service)) {
+	if (typeof image.is_a_build_of__service === 'object') {
```
Page-: This is due to updated lodash types

PR #96, opened 2019-05-15
perms: grant `device-api-key` ability to lookup user public keys

wrboyce:
Change-type: minor


wrboyce:
I don't _think_ we need to be any more restrictive with these permission here?

Page-:
We don't *need* to be, but there's also no reason why we can't be

wrboyce:
Superseded by #147

PR #97, opened 2019-05-16
Update to pinejs 11

Page-:


src/migrations/00014-modified-at.sql
```diff
@@ -0,0 +1,84 @@
+-- The commented sections of this file must be manually run as they affect models
```
Page-: This is the biggest question around the change, since there are some built in tables that have to be manually migrated due to being required for the migrations system to actually get up and running

src/migrations/00014-modified-at.sql
```diff
@@ -0,0 +1,84 @@
...
+-- ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
+
+-- ALTER TABLE "user"
+-- ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
+-- ALTER TABLE "user-has-role"
+-- ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
+-- ALTER TABLE "user-has-permission"
+-- ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
+
+-- resin.sbvr
```
Page-: This needs to be checked over before merge to make sure any tables that have been added in the meantime also get migrated

roman-mazur:
@Page- so what is new in pinejs 11? :)
Is it changing the migrations mechanism?

src/migrations/00014-modified-at.sql
```diff
@@ -0,0 +1,84 @@
...
+-- ALTER TABLE "api key-has-role"
+-- ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
+
+-- ALTER TABLE "permission"
+-- ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
+
+-- ALTER TABLE "role"
+-- ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
+
+-- ALTER TABLE "user"
```
roman-mazur: Why users and permissions tables cannot be handled by normal migrations? Is migrations mechanism relying on data there?

Page-: Yeah, basically the migrations system is implemented as any other pinejs model for the most part and comes after users/permissions are set up since you require the right permissions to run queries against any standard model.  It can in theory be moved beforehand since it uses "root" permissions anyway but the manual migration will be required regardless so there's no incentive to change it until doing so can remove the need for a manual migration (changing it will make the code more complex)

Page-:
@roman-mazur the full list is at https://github.com/balena-io/pinejs/pull/264/commits but the primary change is the addition of "modified at" fields for every resource, with the rest being primarily cleanup that requires breaking changes (dropping callbacks in favour of purely promises, dropping node 6/8 support, dropping defaults when passing `null` instead of `undefined`, returning HttpError objects instead of status codes for internal api call errors so you get both status code + error message available)

src/migrations/00014-modified-at.sql
Page-: Up to date as of v0.50.1

Page-:
@balena-ci retest

Page-:
@balena-ci retest

PR #100, opened 2019-05-17
Fix proxy rejections

dfunckt:
This will significantly reduce reports for https://sentry.io/organizations/balena/issues/748471727

### DIFF
--- a/src/lib/device-proxy.ts
+++ b/src/lib/device-proxy.ts
@@ -170,7 +170,8 @@ export function requestDevices(
 		wait: false;
 	},
 ): Promise<void>;
-// This override is identical to the main form in order for `postDevices` to be able to call it with the generic form
+// This override is identical to the main form in order for `postDevices` to be able
+// to call it with the generic form
 export function requestDevices(
 	opts: RequestDevicesOpts,
 ): Promise<void | RequestResponse[]>;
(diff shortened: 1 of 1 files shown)


src/lib/device-proxy.ts
```diff
@@ -243,26 +244,33 @@ export function requestDevices({
...
+							proxy: `http://resin_api:${API_VPN_SERVICE_API_KEY}@${vpnIp}:3128`,
+							tunnel: true,
+							method: method,
+							timeout: DEVICE_REQUEST_TIMEOUT,
+						});
+						if (!wait) {
+							// this force-cast is super ugly but harmless because clearly noone
+							// cares about the return value of the promise (since wait == false)
+							// so we just need to satisfy the compiler
+							p = p.catchReturn((null as any) as RequestResponse);
```
dfunckt: For some reason I can't understand, the catch handler needs to be on the request promise itself -- if I do `Promise.all(promises).catchReturn(null)` below or `waitPromise.catchReturn()` further down, the exception still propagates. I believe it may have something to do with the promisification and the fact that the error is emit from an event handler, though I do feel that's surprising.

Page-: It'll be due to when the error handler is added, if you made the request fail after 10s instead then the handler would be added before the rejection actually happens and so should be fine

Page-: I think it'd be better to use `undefined`:
```suggestion
							p = p.catchReturn((undefined as any) as RequestResponse);
```
the reason we use null specifically below is because it's got a special meaning to bluebird in that case, but outside of that I'd default to undefined

PR #104, opened 2019-05-21
Fix error when calling `checkDevicesCanHaveDeviceURL` with no devices

Page-:
Change-type: patch


Page-:
I'd say it's more a case of a common assumption that a PATCH request will always affect at least one resource which just isn't the case.  From the pinejs-client side it makes sense to throw an error on this since `$in: []` just don't make any sense and there isn't a valid odata equivalent - it could maybe convert the whole chain to just a hardcoded `false` but that would just be hiding a programmer error and make the issue hard to spot/debug.

I do agree that it's a common issue it's just I don't have a good solution for it in mind and think changing pinejs-client would result in a worse situation overall

PR #106, opened 2019-05-22
Avoid paying the cost of jwt auth if there's no chance of success

Page-:
Change-type: patch


src/platform/jwt.ts
```diff
@@ -144,6 +146,12 @@ export const createJwt = (
 };
 
 export const middleware: RequestHandler = (req, res, next) => {
+	const jwtString = jwtFromRequest(req);
+	if (!jwtString || typeof jwtString !== 'string' || !jwtString.includes('.')) {
+		// If we don't have any possibility of a valid jwt string then we avoid attempting authentication with it altogether
```
dfunckt: Could you please break the comment into a couple of lines?

PR #107, opened 2019-05-23
Native functional

Page-:


src/routes/devices.ts
```diff
@@ -562,7 +566,7 @@ export const state: RequestHandler = (req, res) => {
 									}
 								})
 								.then(() => {
-									_.each(device.manages__device as AnyObject[], depDev => {
+									_.each(device.manages__device, depDev => {
```
dfunckt: It seems the reason you didn't do `device.manages__device` here is that it can be null? If so, you can instead do

```
if (device.manges__device == null) {
  return;
}
device.manages__device.forEach(...);
```

Page-: That was part of it, but it's more I'm just not certain what types it can be here so I prefer lodash for the safety it adds - maybe it could be null, an object, an array, etc and I'd rather play it safe since it's not like keeping `_.each` causes any particular problem

PR #108, opened 2019-05-23
Remove unused code

Page-:
Change-type: minor

### DIFF
--- a/src/lib/env-vars.ts
+++ b/src/lib/env-vars.ts
@@ -149,8 +149,6 @@ const startsWithAny = (ns: string[], name: string) => {
 	return _.some(ns, n => name.startsWith(n));
 };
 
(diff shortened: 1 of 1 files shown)


src/lib/env-vars.ts
```diff
@@ -194,30 +192,3 @@ export const checkEnvVarValueValidity = (value: string) => {
 		);
 	}
 };
-
-export interface EnvVar {
-	name: string;
-	value: string;
-}
-
-export const mergeEnvVars = (
```
dfunckt: This is used by the /environment endpoint in cloud API. Are you moving it out of here or completely removing it?

Page-: Removing it completely from open source because it's unused here and is purely used for very very old devices

Page-:
@balena-ci retest

PR #109, opened 2019-05-23
Fixes

Page-:


src/routes/devices.ts
```diff
@@ -584,15 +584,16 @@ export const state: RequestHandler = (req, res) => {
...
 										const svcInstall =
 											image == null
 												? null
 												: serviceInstallFromImage(depDev, image);
 
-										if (image != null) {
+										if (ipr != null) {
 											varListInsert(
-												image.image_environment_variable,
+												ipr.image_environment_variable,
```
dfunckt: This bug must have been lurking there for a while, I don’t see it having been touched by the recent changes.

Page-: I think it's been there since the original version but never noticed because I don't think anyone uses them (we don't expose them in the dashboard/sdk)

Page-: Oh but the reason it became visible if you didn't realise is due to recent changes to `varListInsert` which meant that the incorrect path would cause a failure (thanks to @thgreasi's work though that I was easily able to realise why/fix it)

PR #115, opened 2019-05-28
💥🗃: Device types

afitzek:
This adds a device type table to the datamodel
and links it to relevant tables (device, application).

Change-type: minor (This is actually a major (breaking) change, but we don't want to jump to version 1.0.0 yet.)


afitzek:
@balena-ci retest

afitzek:
@balena-ci retest

src/migrations/00012-add-device-types.sql
```diff
@@ -0,0 +1,317 @@
...
+END $$;
+
+-- populate device types table
+
+INSERT INTO "device type" ("slug", "name")
+	SELECT "dt" as "slug", "dt" as "name" FROM
+	(
+		select dt FROM
+		(
+			SELECT DISTINCT lower("device type") as "dt" FROM "device"
```
Page-: Pretty sure the `UNION` will handle this `DISTINCT` for you

src/platform/middleware.ts
```diff
@@ -28,6 +28,12 @@ export const authorized: RequestHandler = (req, res, next) =>
 			res.sendStatus(401);
 		});
 
+export const identify: RequestHandler = (req, _res, next) =>
```
Page-: What is this for?

src/routes/applications.ts
```diff
@@ -49,16 +54,20 @@ export const downloadImageConfig: RequestHandler = (req, res) => {
 		return;
 	}
 
+	const api = resinApi.clone({ passthrough: { req } });
+
 	return getApp(req)
+		.tap(app => console.error(`Found app! ${JSON.stringify(app)}`))
```
Page-: Debug code
```suggestion
```

src/lib/device-types/index.ts
```diff
@@ -246,21 +320,32 @@ export const validateSlug = (slug?: string) => {
...
-			)
-			.map(deviceTypesInfo => deviceTypesInfo.latest.deviceType)
-			.value();
-	});
+export const deviceTypes = (api: PinejsClient): Promise<DeviceType[]> => {
+	return getDeviceTypes()
+		.then(deviceTypesInfos => {
+			// exclude aliases
+			return _(deviceTypesInfos)
+				.filter(
```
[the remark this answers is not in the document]
afitzek: This would be something different, because this filter eliminates aliases, but `getAccessibleSlugs` should also work on alias slugs.

src/platform/middleware.ts
afitzek: The `/device-types/v1` api can be called without any authentication, this adds a middleware that optionally authenticates the user.

src/lib/device-types/index.ts
```diff
@@ -226,18 +260,58 @@ function getDeviceTypes(): Promise<Dictionary<DeviceTypeInfo>> {
 	return fetchDeviceTypesAndReschedule();
 }
 
+export const getAccessibleSlugs = (
+	api: PinejsClient,
+	slugs?: string[],
+): Promise<string[]> => {
+	const options: AnyObject = {
```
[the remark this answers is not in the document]
afitzek: Trying to replace this I get the error: `'PinejsClient' only refers to a type, but is being used as a namespace here.`

src/hooks/resources/application.ts
```diff
@@ -79,12 +79,41 @@ sbvrUtils.addPureHook('POST', 'resin', 'application', {
...
+					resource: 'device_type',
+					id: request.values.is_for__device_type,
+					options: {
+						$select: ['slug'],
+					},
+				});
+			})
+			.then((dt: { slug: string }) => {
+				if (!dt) {
+					throw new deviceTypes.UnknownDeviceTypeError('');
```
dfunckt: Why the empty string?

src/lib/device-types/index.ts
```diff
@@ -226,18 +260,51 @@ function getDeviceTypes(): Promise<Dictionary<DeviceTypeInfo>> {
...
+			options,
+		})
+		.then((accessibleDeviceTypes: { slug: string }[]) => {
+			return _.map(accessibleDeviceTypes, 'slug');
+		});
+};
+
 export const findDeviceTypeInfoBySlug = (
 	slug: string,
+	api: PinejsClient,
```
dfunckt: By convention, `api` is typically the first argument in a function.

src/lib/device-types/index.ts
```diff
@@ -246,21 +313,32 @@ export const validateSlug = (slug?: string) => {
...
+				});
+			});
+		});
 };
 
-export const findBySlug = (slug: string): Promise<DeviceType> =>
-	deviceTypes()
+export const findBySlug = (
+	slug: string,
+	api: PinejsClient,
```
dfunckt: Same here -- `api` should be the first argument.

And below in `normalizeDeviceType`, `getImageSize`, etc. Can you please ensure all changed functions accept `api` as their first argument?

src/routes/applications.ts
```diff
@@ -49,16 +54,19 @@ export const downloadImageConfig: RequestHandler = (req, res) => {
...
-				generateConfig(req, app, deviceType, osVersion),
+			findBySlug(deviceTypeSlug || app.is_for__device_type[0].slug, api).then(
+				deviceType => generateConfig(req, app, deviceType, osVersion),
 			),
 		)
 		.then(config => {
 			res.json(config);
 		})
 		.catch(UnauthorizedError, err => {
+			console.error(`Failed ${err.message}`);
```
dfunckt: Leftover debug comment.

src/lib/auth.ts
```diff
@@ -77,6 +78,7 @@ export const ROLES: {
 		'resin.actor.delete?id eq @__ACTOR_ID',
 		'resin.api_key.read?is_of__actor eq @__ACTOR_ID',
 		'resin.application.all',
+		'resin.device_type.all',
```
dfunckt: As things stand, device types are currently controlled by balenaCloud (given we're hardcoding our production S3 bucket into openBalena), therefore I think read-only permissions are better suited for this type of thing right now, if only to ensure an installation doesn't break by accident.

src/lib/device-types/index.ts
```diff
@@ -49,6 +50,14 @@ const SPECIAL_SLUGS = ['edge'];
 const RETRY_DELAY = 2000; // ms
 const DEVICE_TYPES_CACHE_EXPIRATION = 5 * 60 * 1000; // 5 mins
 
+const syncSettings: { map: Dictionary<string> } = {
+	map: {},
+};
+
+export function setSyncMap(map: typeof syncSettings['map']) {
```
dfunckt: This function is unused. If you plan to use it as an extension point from resin-api, then please add a comment to that effect.

src/hooks/resources/device.ts
```diff
@@ -141,25 +140,49 @@ sbvrUtils.addPureHook('POST', 'resin', 'device', {
...
+					resource: 'device_type',
+					id: request.values.is_of__device_type,
+					options: {
+						$select: ['slug'],
+					},
+				});
+			})
+			.then((dt: { slug: string }) => {
+				if (!dt) {
+					throw new deviceTypes.UnknownDeviceTypeError('');
```
dfunckt: Empty string here instead of error message.

src/hooks/resources/application.ts
```diff
@@ -79,12 +79,43 @@ sbvrUtils.addPureHook('POST', 'resin', 'application', {
...
+					resource: 'device_type',
+					id: request.values.is_for__device_type,
+					options: {
+						$select: ['slug'],
+					},
+				});
+			})
+			.then((dt: { slug: string }) => {
+				if (!dt) {
+					throw new deviceTypes.UnknownDeviceTypeError(
```
dfunckt: There's also `InvalidDeviceTypeError` which I think may be more suitable as an error type to throw here -- `UnknownDeviceTypeError` is used with a `slug` as the message somewhere below. There's also a couple of plain `Error`s thrown above -- can't they be more specific? Do you mind going through the errors thrown and sort of making sure they make sense in each case? Are the response codes suitable for each case?

src/hooks/resources/device.ts
```diff
@@ -141,25 +140,49 @@ sbvrUtils.addPureHook('POST', 'resin', 'device', {
...
+			) {
+				// translate device_type to is_for__device_type
+				return deviceTypes
+					.getDeviceTypeIdBySlug(api, request.values.device_type)
+					.then(dt => {
+						if (dt) {
+							request.values.is_of__device_type = dt.id;
+							return;
+						}
+						throw new Error('Invalid device type value');
```
dfunckt: `InvalidDeviceTypeError` seems more appropriate here.

src/lib/device-types/index.ts
afitzek: Good catch. I forgot to call it in the onInitHook function.

src/hooks/resources/device.ts
afitzek: This would probably make more sense as `deviceTypes.UnknownDeviceTypeError(request.values.device_type)`

init.ts
```diff
@@ -86,6 +88,12 @@ app.enable('trust proxy');
 
 const doRunTests = (process.env.RUN_TESTS || '').trim() === '1';
 
+// we have to load some mocks before the app starts...
+if (doRunTests) {
+	console.log('Loading mocks...');
+	import('./test/test-lib/aws-mock');
```
dfunckt: Can this whole block be moved into https://github.com/balena-io/open-balena-api/blob/master/test/00-init.ts?

I'm now seeing you already do the import there -- can you please remove this one?

src/hooks/resources/device.ts
```diff
@@ -141,25 +140,51 @@ sbvrUtils.addPureHook('POST', 'resin', 'device', {
...
+						$select: ['slug'],
+					},
+				});
+			})
+			.then((dt: { slug: string }) => {
+				if (!dt) {
+					throw new deviceTypes.InvalidDeviceTypeError();
+				}
+				return deviceTypes.findBySlug(api, dt.slug);
+			});
```
dfunckt: All this added code seems to be identical in POST and PATCH hooks -- can you please factor it out in a shared function that's called from both places?

test/00-init.ts
```diff
@@ -1,25 +1,26 @@
+import './test-lib/aws-mock';
 import * as Promise from 'bluebird';
 import * as _ from 'lodash';
 import * as fs from 'fs';
 import * as path from 'path';
 
+console.error(`Running tests 00`);
```
dfunckt: Leftover debug comment

test/02-device-types.ts
```diff
@@ -0,0 +1,320 @@
...
+import { app } from '../init';
+import * as _ from 'lodash';
+import { expect } from './test-lib/chai';
+
+import supertest = require('./test-lib/supertest');
+
+describe('device type endpoints', () => {
+	describe('/device-types/v1', () => {
+		it('should succeed to return a result', () => {
+			return supertest(app)
```
dfunckt: You can use async/await and vastly improve readability here. See `tests/XX_organizations.ts` in resin-api if you're unsure how.

init.ts
afitzek: @dfunckt I tried this, but it does not work if I only import it there. I stole this from the PR from rich here: https://github.com/balena-io/open-balena-api/pull/143/files

test/02-device-types.ts
afitzek: Good point. I migrated it from coffeescript, will refactor it to async/await.

init.ts
dfunckt: Can we instead create a new file in `tests/`, put all pre-run code there and require it from here? The end result is practically the same, but will not encourage our future selves to add more and more test related code, which just doesn't fit here.

src/hooks/resources/device.ts
afitzek: I think you mean between the application and device resource? I refactored these to a shared functions.

init.ts
afitzek: That sound good! Will refactor it like this.

src/hooks/resources/device.ts
dfunckt: Yes, sorry. I saw the changes and they look really good!

src/hooks/common.ts
```diff
@@ -0,0 +1,45 @@
+import { HookRequest } from '@resin/pinejs/out/sbvr-api/sbvr-utils';
+import { PinejsClient } from '../platform/index';
+import * as deviceTypes from '../lib/device-types';
+import * as Promise from 'bluebird';
+
+export const resolveDeviceTypeInHook = (
```
dfunckt: Given the context, this can be named just “resolveDeviceType”.

test/00-init.ts
```diff
@@ -1,3 +1,4 @@
+import './test-lib/aws-mock';
```
thgreasi: Do we still need this, even after adding `init_mocks.ts` in ` init.ts`?

src/migrations/00012-add-device-types.sql
```diff
@@ -0,0 +1,302 @@
...
+,	"slug" VARCHAR(255) NOT NULL UNIQUE
+,	"name" VARCHAR(255) NOT NULL
+);
+
+-- populate device types table
+
+INSERT INTO "device type" ("slug", "name")
+	SELECT "dt" as "slug", "dt" as "name" FROM
+	(
+		select dt FROM
```
thgreasi: NIT:
```suggestion
		SELECT dt FROM
```

src/migrations/00012-add-device-types.sql
```diff
@@ -0,0 +1,302 @@
...
+,	"modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL
+,	"id" SERIAL NOT NULL PRIMARY KEY
+,	"slug" VARCHAR(255) NOT NULL UNIQUE
+,	"name" VARCHAR(255) NOT NULL
+);
+
+-- populate device types table
+
+INSERT INTO "device type" ("slug", "name")
+	SELECT "dt" as "slug", "dt" as "name" FROM
```
thgreasi: NIT: Change the remaining `as` to `AS`  :)

src/migrations/00012-add-device-types.sql
```diff
@@ -0,0 +1,302 @@
...
+    'artik10',
+    'Samsung Artik 10 (DISCONTINUED)'
+  ),
+  (
+    'qemux86',
+    'QEMU X86 32bit (EXPERIMENTAL)'
+  ),
+  (
+    'parallella',
+    'Parallella (DISCONTINUED)'
```
thgreasi: I would expect us not adding discontinued device types in the DB. Even in case that a migrating user still uses them, we do insert the `FULL_DT` above based on all their old `device_type` records.
That's definitely not a blocker for me though, if we think adding them anyway will be safer.

src/lib/device-types/index.ts
```diff
@@ -214,7 +246,9 @@ function fetchDeviceTypesAndReschedule(): Promise<Dictionary<DeviceTypeInfo>> {
 		updateDeviceTypesCache(promise);
 	}
 
-	return promise;
+	return promise.tap(deviceTypeInfos => {
+		return syncDataModel(deviceTypeInfos, syncSettings.map);
```
thgreasi: * Since the updateOrInsert in here runs on each run of `fetchDeviceTypesAndReschedule` would it make sense to check whether the new device type cache did change at all, or (as mentioned above) add some the respective check in the `updateOrInsert`?
* Also, shouldn't we provide the option to consumer projects of the open-balena-api to only run this once (so that the DB is the source of truth of the known device types after the initial sync)? Or we don't yet care for this yet?

src/lib/device-types/index.ts
```diff
@@ -187,6 +196,29 @@ function updateDeviceTypesCache(
...
+	propertyMap: Dictionary<string>,
+) {
+	return db.transaction(tx => {
+		return Promise.each(Object.values(types), deviceTypeInfo => {
+			const deviceType = deviceTypeInfo.latest.deviceType;
+			const body = _.mapValues(
+				_.invert(propertyMap),
+				source => (deviceType as AnyObject)[source],
+			);
+			return updateOrInsertModel(
```
thgreasi: If we keep calling this on each run of `fetchDeviceTypesAndReschedule` would it make sense to change `updateOrInsert` to only run an UPDATE if the objects are different?

src/lib/device-types/index.ts
afitzek: I guess we can look at extending `updateOrInsert`, I would not only run this once, because in that case instances would not receive new device types without restarting.

src/migrations/00012-add-device-types.sql
afitzek: If we don't include it in the migration, they will be synced through s3 during runtime.

src/lib/device-types/index.ts
afitzek: I would not change `updateOrInsert` but maybe add a new function that behaves the way you describe, so that we don't break anything.

src/hooks/resources/application.ts
```diff
@@ -79,12 +79,7 @@ sbvrUtils.addPureHook('POST', 'resin', 'application', {
 			throw new Error('App name may only contain [a-zA-Z0-9_-].');
 		}
 
-		return Promise.all([
-			deviceTypes
-				.normalizeDeviceType(request.values.device_type)
-				.then(deviceType => (request.values.device_type = deviceType)),
-			checkDependentApplication(args),
```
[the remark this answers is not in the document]
afitzek: Good catch. The `checkDependentApplication` call fell through the cracks.

src/migrations/00012-add-device-types.sql
dfunckt: Not only that, but the migration will fail as well, because there won't be a device-type row matching the app's/device's device type slug.

thgreasi: @dfunckt there will be a match, since right above we do a union+toLowerCase with all current device type slugs in the DB.
@afitzek fair enough :+1:

src/migrations/00012-add-device-types.sql
thgreasi: There are still a few `as`, but anyway...

src/lib/device-types/index.ts
```diff
@@ -246,21 +329,34 @@ export const validateSlug = (slug?: string) => {
...
+			const deviceTypes: DeviceType[] = _(deviceTypesInfos)
+				.filter(
+					(deviceTypesInfo: DeviceTypeInfo, slug: string) =>
+						deviceTypesInfo.latest.deviceType.slug === slug,
+				)
+				.map(deviceTypesInfo => deviceTypesInfo.latest.deviceType)
+				.value();
+			const accessSet = new Set(accessibleDeviceTypes);
+			return _.filter(deviceTypes, (o: DeviceType) => {
+				return accessSet.has(o.slug);
```
thgreasi: How about moving this in the lodash chained filter above, instead of having a separate `_.filter()` call?
```ts
const accessSet = new Set(accessibleDeviceTypes);
const deviceTypes: DeviceType[] = _(deviceTypesInfos)
	.filter(
		(deviceTypesInfo: DeviceTypeInfo, slug: string) => {
			const dtSlug = deviceTypesInfo.latest.deviceType.slug;
			return dtSlug === slug && accessSet.has(dtSlug),
		}
	)
	.map(deviceTypesInfo => deviceTypesInfo.latest.deviceType)
	.value();
```
PS: I was probably not clear enough earlier.

src/migrations/00012-add-device-types.sql
```diff
@@ -0,0 +1,302 @@
...
+  ),
+  (
+    'via-vab820-quad',
+    'VIA VAB 820-quad (DISCONTINUED)'
+  ),
+  (
+    'zynq-xz702',
+    'Zynq ZC702 (DISCONTINUED)'
+  )
+ON CONFLICT ON CONSTRAINT "device type_slug_key"
```
thgreasi: NIT: How about updating the `name` `ON CONFLICT`?
If we don't, then the name will equal the slug, as inserted in the FULL_DT code above.
Not a blocker, since we will update them anyway with the next fetch from S3....

afitzek:
@thgreasi Cool. I think I solved the update problem as well.
I modified the patch operation to include a filter, that matches the values that we would set. So the generated UPDATE query looks like this: 
```
UPDATE "device type"
SET "name" = $1
WHERE ("device type"."id") IS NOT NULL AND ("device type"."id") = ($2)
AND "device type"."id" IN ((
	SELECT "device type"."id"
	FROM "device type"
	WHERE ("device type"."id") IS NOT NULL AND ("device type"."id") = ($3)
	AND (NOT(("device type"."name") IS NOT NULL AND ("device type"."name") = ($4)))
))
```
which will only perform an actual update operation if the values differ.

PR #118, opened 2019-05-29
Add an entry for the supervisor mixpanel controlling var

CameronDiver:
Change-type: patch


src/lib/env-vars.ts
```diff
@@ -24,6 +24,7 @@ export const WHITELISTED_NAMES = [
 	'RESIN_SUPERVISOR_VPN_CONTROL',
 	'RESIN_SUPERVISOR_PERSISTENT_LOGGING',
 	'RESIN_SUPERVISOR_INSTANT_UPDATE_TRIGGER',
+	'RESIN_SUPERVISOR_MIXPANEL_REPORT',
```
Page-: Is this necessary?  Has this var ever been supported under the old `RESIN` namespace?

richbayliss: I thought that same, but I figured it's just good practice to keep the `RESIN_` namespace for us, for now.

dfunckt: It isn't needed really -- users can't set anything prefixed with `RESIN_`

PR #121, opened 2019-06-03
auth: check username blacklist when registering new users

wrboyce:
Connects-to: #120
Change-type: major

### DIFF
--- a/src/platform/auth.ts
+++ b/src/platform/auth.ts
@@ -20,6 +20,8 @@ const {
 
 const SUDO_TOKEN_VALIDITY = 20 * 60 * 1000;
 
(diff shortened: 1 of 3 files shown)


src/platform/auth.ts
```diff
@@ -20,6 +20,8 @@ const {
 
 const SUDO_TOKEN_VALIDITY = 20 * 60 * 1000;
 
+const USERNAME_BLACKLIST = ['root'];
```
dfunckt: Err, the single openBalena user gets username `root`.

wrboyce: Ah… That is going to cause issues (with ssh). Is it easily changed?

wrboyce:
Superseded by #147

PR #123, opened 2019-06-04
Avoid unnecessary writes on login by adding an optional hook

Page-:
Change-type: patch

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -46,6 +46,8 @@ export interface SetupOptions {
 	onInitModel?: SetupFunction;
 	onInitHooks?: SetupFunction;
 	onInitRoutes?: SetupFunction;
+
+	onLogin?: (user: AnyObject) => PromiseLike<void> | void;
 }
 
 export function setup(app: _express.Application, options: SetupOptions) {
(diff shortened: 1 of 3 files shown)


src/routes/session.ts
```diff
@@ -46,6 +46,14 @@ export const login: RequestHandler = (req, res) => {
 						resource: 'user',
 						id: user.id,
 						passthrough: { req: root },
+						options: {
+							$filter: {
+								$or: {
+									password_reset_code: { $ne: null },
+									can_reset_password_until__expiry_date: { $ne: null },
```
dfunckt: These fields don't exist on the open source model, see the comment below. I don't think filtering on them will work.

src/index.ts
```diff
@@ -46,6 +46,8 @@ export interface SetupOptions {
 	onInitModel?: SetupFunction;
 	onInitHooks?: SetupFunction;
 	onInitRoutes?: SetupFunction;
+
+	onLoginHook?: (user: AnyObject) => PromiseLike<void> | void;
```
dfunckt: Naming this `onLogin` is more consistent with the other hooks.

PR #126, opened 2019-06-12
Avoid unnecessary writes on the state patch endpoint

Page-:
Change-type: patch


package.json
```diff
@@ -83,10 +83,10 @@
     "@types/supertest": "^2.0.7",
     "chai": "^4.2.0",
     "copyfiles": "^2.1.0",
-    "husky": "^2.3.0",
-    "lint-staged": "^8.1.7",
+    "husky": "^2.4.1",
+    "lint-staged": "^8.2.0",
     "mocha": "^6.1.4",
-    "prettier": "^1.17.1",
+    "prettier": "^1.18.2",
```
Page-: This is responsible for all changes outside of `src/routes/devices.ts`

src/hooks/resources/device.ts
```diff
@@ -365,15 +365,19 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
 			args.request.values.device_name != null
 		) {
 			waitPromises.push(
-				affectedIds.then(deviceIds =>
-					postDevices({
+				affectedIds.then(deviceIds => {
+					if (deviceIds.length === 0) {
```
Page-: Had to also add this check because with the unnecessary write avoidance suddenly no devices get affected :P

PR #132, opened 2019-06-14
auth: allow device API keys to fully control the device's env vars

imrehg:
Both device env vars and device service env vars are covered. Previously could only read the variables, but should be able to create/change/delete as well.

See: [Flowdock](https://www.flowdock.com/app/rulemotion/resin-tech/threads/hV567d4LHA6HNUaaxZVIiRfN29R)
Change-type: minor

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -28,7 +28,8 @@ export const ROLES: {
 		`resin.application_config_variable.get?${applicationControlsDevice}`,
 		`resin.release.read?should_be_running_on__device/any(d:d/${matchesActor})`,
 		`resin.release.read?${belongsToApplicationOrIsManagedByDevice}`,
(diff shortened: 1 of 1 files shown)


src/lib/auth.ts
```diff
@@ -42,7 +42,7 @@ export const ROLES: {
 
 		`resin.service_environment_variable.get?service/any(s:s/${applicationControlsDevice})`,
 
-		`resin.device_service_environment_variable.get?service_install/any(si:si/installs__service/any(s:s/${applicationControlsDevice}))`,
+		`resin.device_service_environment_variable.all?service_install/any(si:si/installs__service/any(s:s/${applicationControlsDevice}))`,
```
richbayliss: hmmm - I don't see a limit to the specific device here @imrehg

richbayliss: I don't think this is the permission to be changed.

imrehg: Indeed, actually, need to create a new permission line.

imrehg:
Tested downstream, merging based on comments here and those test results.

PR #133, opened 2019-06-17
Allow newlines in environment variables

CameronDiver:
Change-type: patch


CameronDiver:
@dfunckt yeah it's a good point, I have an arch call discussion waiting for this - so we'll see.

PR #134, opened 2019-06-17
Delete dependent hook

Page-:


src/hooks/resources/application.ts
```diff
@@ -177,73 +178,16 @@ sbvrUtils.addPureHook('DELETE', 'resin', 'application', {
...
 				}
-				return;
 			}
-			const { req, api } = args;
-			return Promise.mapSeries(
-				[
-					[
-						'device',
-						'belongs_to__application',
-						['release', 'belongs_to__application'],
```
Page-: I dropped this line as well because it was a duplicate

PR #139, opened 2019-06-21
Fix unreturned promise warning in log-stream

Page-:
Change-type: patch


dfunckt:
@Page- I can’t understand what actually changed. “schedule()” was returning void and still is after your changes so I’m not quite sure what is the issue this PR fixes and how. Can you please clarify?

Page-:
Yep, in this case the warning was a false positive so silencing it just means true positives are easier to spot/stand out more :)

PR #140, opened 2019-06-24
Add some indexes for delete cascading

Page-:
Change-type: patch


Page-:
They were getting ridiculously long and as long as they're unique it's fine.  That said I prefer them to have readable names, it's just hard to make it work nicely :|

PR #143, opened 2019-06-27
tests: Add fixtures framework

richbayliss:


test/test-lib/fixtures.ts
```diff
@@ -0,0 +1,697 @@
...
+	});
+	const [response, body] = await requestAsync(requestOpts);
+
+	if (response.statusCode !== 201) {
+		logErrorAndThrow('Failed to create device', jsonData, body);
+	}
+	return body;
+};
+
+loaders['image_environment_variables'] = async (jsonData, fixtures) => {
```
dfunckt: Some loaders are keyed by names that use dashes `-`, others with underscores `_`. I think it's worth picking one of the two styles now that no fixtures exist yet.

test/test-lib/resin-token.ts
```diff
@@ -0,0 +1,8 @@
+import * as resinToken from 'resin-token';
+import * as temp from 'temp';
+
+const trackedTemp = temp.track();
+
+export = resinToken({
```
dfunckt: Can you please switch to `export const ... = resinToken('...')` format? And possibly also rename the file to balena-token, or just token instead?

automation/common.sh
```diff
@@ -52,9 +52,9 @@ runapi () {
...
 		-e DEVICE_CONFIG_SSH_AUTHORIZED_KEYS='ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQCc5cLVsu7b52tzxwa8A6cs4LC2I+IMtyKeVZFVAWdhfEWGEJL78rn0tC4YmHBSWW0hBeZmn8f7ZKcZ/r1W2RpuH7wpUc56/3hffJ0GoktMOX69pjNsE3rWr8/cZbl/LaJjeHBpyFZ6VimYl5KiYyHJH7RZwlEYYiJ2YDCbte51jvmotFTqiOJNPCM3SXWFjsUGDOWwNi7GDxG+ulb0cBnYzyYlHTj5dS/uya0v8S267TYO2jZ9lUvZHYYQBilYoN3qT79B/lRJq/AKsaSPI2VWkwoAWLObJRjQjoIsClxcfaSfRg6fgdezv3saeorVMIdEYAj3CQVmLszCNGC8dgqB9OoKdjd/OUG/bFUc86S+y0dZX3vqKBfZjE1dHNzNkQeulsbRj0FGFTxFYL0Ng7jlr98iXusnNg236LGsBTeod2MLfMRsrOc0Utfc5Pq2Zkyak4aJ6URBFqZnKlJwvI3AFznxvHD1PmIrwzCCmLA409v3WbXEK4KZrUIs81wlQWAK3919ViozdnR23f4hZtq3P5M6AbnMMdIOB1jDjPK//eaTMf3erHVa0nTMMwScBDnrT86yJBBEwpUVpIp6YUBFaVMQscX3gBTd+RyOU9jOvyinGjosJfF38qsMJTNHh48zgic5sSLhtPQgkUjWhpcnfjqrLRewKbVd+VZQc1WsdQ==' \
 		-e IMAGE_MAKER_URL=https://img.balenadev.io \
-		-e IMAGE_STORAGE_BUCKET=balena-dev-img \
-		-e IMAGE_STORAGE_ENDPOINT=s3.balenadev.io \
-		-e IMAGE_STORAGE_PREFIX=images \
+		-e IMAGE_STORAGE_BUCKET=resin-production-img-cloudformation \
+		-e IMAGE_STORAGE_PREFIX=resinos \
+		-e IMAGE_STORAGE_ENDPOINT=s3.amazonaws.com \
```
dfunckt: Do tests depend on making actual calls to S3?

test/test-lib/fixtures.ts
richbayliss: agreed - I will go with whatever we use in Pine resource queries, since dashes won't work directly anyway (`obj.a-b-c` is invalid) and underscores will trigger TSLint (`obj.a_b_c` will be snake-case) so in both instances we would refer via index (`obj['a-b-c']` or `obj['a_b_c']`)

test/test-lib/resin-token.ts
richbayliss: Sure, I can take a look at that

automation/common.sh
richbayliss: Yes. If you make use of the `applications` fixture then it will make an S3 call to determine device-types. Without mocking the S3 calls, this is needed.

test/test-lib/fixtures.ts
richbayliss: I will do the `a_b_c` as this is how we reference the resources in Pine. Keeps it aligned 👍

automation/common.sh
richbayliss: I am going to mock the S3 calls 👍

test/test-lib/fixtures.ts
```diff
@@ -0,0 +1,695 @@
...
+			return res.text;
+		});
+
+	const user = (await balenaToken.parse(token)) as AnyObject & { id: number };
+	user.token = token;
+
+	user.actor = await supertest(app, user)
+		.get(`/resin/user(${user.id})`)
+		.expect(200)
+		.then(res => res.body.d[0].actor as number);
```
Page-: Should be
```ts
const res = await supertest(app, user)
...

File: test/test-lib/fixtures.ts [hunk 5]
```diff
@@ -0,0 +1,695 @@
...
+	}
+
+	const token = await supertest(app)
+		.post('/login_')
+		.send({
+			username: SUPERUSER_EMAIL,
+			password: SUPERUSER_PASSWORD,
+		})
+		.expect(200)
+		.then(res => {
```
Page-: Should be
```ts
const res = await supertest(app)
...

PR #147, opened 2019-07-08
Requirements for direct ssh authentication

wrboyce:


src/lib/config.ts
```diff
@@ -97,6 +97,8 @@ export const REGISTRY2_HOST = requiredVar('REGISTRY2_HOST');
 export const REGISTRY_HOST = requiredVar('REGISTRY_HOST');
 export const SENTRY_DSN = process.env.SENTRY_DSN;
 export const SUPERUSER_EMAIL = process.env.SUPERUSER_EMAIL || '';
+export const SUPERUSER_USERNAME =
+	process.env.SUPERUSER_EMAIL || SUPERUSER_EMAIL.split('@')[0];
```
richbayliss: I'm not sure where the requirement for this has come from, but I think we should keep using `root` as the default username.

src/platform/auth.ts
```diff
@@ -357,6 +359,9 @@ export const registerUser = (
 	},
 	tx: Tx,
 ): Promise<AnyObject> => {
+	if (USERNAME_BLACKLIST.includes(userData.username)) {
```
richbayliss: Why would we want to blacklist the `root` user?

src/resin.sbvr
```diff
@@ -55,6 +55,11 @@ Fact type: user has permission
 	Fact type: user permission has expiry date
 		Necessity: Each user permission has at most one expiry date.
 
+Term: public key
+	Concept Type: Text (Type)
+Fact type: user has public key
```
richbayliss: `user (Auth)` this should be

src/migrations/00010-rename-root-user.sql
```diff
@@ -0,0 +1 @@
+UPDATE "user" SET username='admin' WHERE username='root';
```
richbayliss: Again, not sure this is desirable.

wrboyce:
TLDR on the root username issue is that we need to use a non-root username for the device side of ssh auth (for ssh key lookups via the api).

init.ts
```diff
@@ -54,7 +56,7 @@ async function createSuperuser() {
 	const { ConflictError } = sbvrUtils;
 
 	const data = {
-		username: 'root',
+		username: SUPERUSER_USERNAME,
```
dfunckt: This will create a new user every time the username changes. I would hardcode `'admin'` right here, like we had `'root'` before.

wrboyce: Nice catch, will update.

src/lib/auth.ts
```diff
@@ -101,6 +104,7 @@ export const ROLES: {
 		"resin.service_instance.get?service_type eq 'vpn'",
 		'resin.service_label.all',
 		'resin.user.read',
+		`resin.user__has__public_key.all?${matchesActor}`,
```
richbayliss: `matchesActor` won't work here, as it only has a relationship to the user. It needs to look at the `user` then link to the `actor`

src/routes/auth.ts
```diff
@@ -0,0 +1,29 @@
+import { RequestHandler } from 'express';
+
+import { resinApi } from '../platform';
+import { handleHttpErrors } from '../platform/errors';
+
+export const getUserPublicKeys: RequestHandler = (req, res) => {
```
dfunckt: Public keys are, well, public. We might as well just drop the newly added permissions, which are far too liberal (grants the device access to all users on the platform), and have this endpoint use `root` privileges to fetch public keys of any user.

src/routes/auth.ts
```diff
@@ -0,0 +1,29 @@
...
+
+	if (username == null) {
+		return res.send(400);
+	}
+
+	return resinApi
+		.get({
+			resource: 'user__has__public_key',
+			options: {
+				$select: ['user', 'public_key'],
```
Page-: I can't see `user` being used in the response at all
```suggestion
				$select: 'public_key',
```

src/lib/auth.ts
```diff
@@ -101,6 +104,7 @@ export const ROLES: {
 		"resin.service_instance.get?service_type eq 'vpn'",
 		'resin.service_label.all',
 		'resin.user.read',
+		`resin.user__has__public_key.all?`,
```
Page-: This shouldn't have the `?` without a filter:
```suggestion
		`resin.user__has__public_key.all`,
```

src/routes/auth.ts
```diff
@@ -0,0 +1,36 @@
...
+	return resinApi
+		.get({
+			resource: 'user__has__public_key',
+			options: {
+				$select: 'public_key',
+				$filter: {
+					user: {
+						$any: {
+							$alias: 'u',
+							$expr: { username },
```
dfunckt: Apologies if I suggested it this way in my comment -- this should be instead:

```
$expr: {
  u: { username }
}
```

PR #152, opened 2019-07-19
Add ability for clients to register hooks for device state endpoint

dfunckt:
Change-type: minor

### DIFF
--- a/src/lib/device-state.ts
+++ b/src/lib/device-state.ts
@@ -47,6 +47,35 @@ export const serviceInstallFromImage = (
 export const formatImageLocation = (imageLocation: string) =>
 	imageLocation.toLowerCase();
 
+export type DeviceConfigHook = (
+	configVars: Dictionary<string>,
+	osVersion: string,
+) => void;
+
+export interface HookRegistry {
+	deviceConfig: DeviceConfigHook[];
+}
+
+const STATE_ENDPOINT_HOOKS: HookRegistry = {
+	deviceConfig: [],
+};
+
+export function registerHook<Key extends keyof HookRegistry>(
+	name: Key,
+	fn: typeof STATE_ENDPOINT_HOOKS[Key][number],
+) {
+	STATE_ENDPOINT_HOOKS[name].push(fn);
+}
+
+function runHook<Key extends keyof HookRegistry>(
+	name: Key,
+	args: Parameters<typeof STATE_ENDPOINT_HOOKS[Key][number]>,
+) {
+	STATE_ENDPOINT_HOOKS[name].forEach(hook =>
+		hook(...(args as Parameters<typeof hook>)),
+	);
+}
+
 // Some config vars cause issues with certain versions of resinOS.
 // This function will check the OS version against the config
 // vars and filter any which cause problems, returning a new map to
@@ -64,4 +93,6 @@ export const filterDeviceConfig = (
 	if (semver.gte(osVersion, '2.0.0')) {
 		delete configVars.RESIN_HOST_LOG_TO_DISPLAY;
 	}
+
+	runHook('deviceConfig', [configVars, osVersion]);
 };


src/lib/device-state.ts
```diff
@@ -97,6 +97,30 @@ export const serviceInstallFromImage = (
...
+
+export function registerHook<
+	K extends keyof HookRegistry,
+	Fn extends typeof STATE_ENDPOINT_HOOKS[K][number]
+>(name: K, fn: Fn) {
+	STATE_ENDPOINT_HOOKS[name].push(fn);
+}
+
+function runHook(name: keyof HookRegistry, args: any[]) {
+	STATE_ENDPOINT_HOOKS[name].forEach(hook => hook.apply(null, args));
```
Page-:
```suggestion
	STATE_ENDPOINT_HOOKS[name].forEach(hook => hook(...args));
```

src/lib/device-state.ts
```diff
@@ -97,6 +97,30 @@ export const serviceInstallFromImage = (
...
+};
+
+export function registerHook<
+	K extends keyof HookRegistry,
+	Fn extends typeof STATE_ENDPOINT_HOOKS[K][number]
+>(name: K, fn: Fn) {
+	STATE_ENDPOINT_HOOKS[name].push(fn);
+}
+
+function runHook(name: keyof HookRegistry, args: any[]) {
```
Page-: You should be able to get strict types with
```suggestion
function runHook<T extends keyof HookRegistry>(name: T, args: Parameters<HookRegistry[T][number]>) {
```

dfunckt: Ah nice, I searched for something like `Parameters` but didn't find it.

src/lib/device-state.ts
dfunckt: This unfortunately doesn't work, even after adapting the signature as you suggested below.

Page-: You can do
```suggestion
	STATE_ENDPOINT_HOOKS[name].forEach(hook => hook(...args as Parameters<typeof hook>));
```
instead and still be able to use `...` instead

src/lib/device-state.ts
```diff
@@ -97,6 +97,33 @@ export const serviceInstallFromImage = (
...
+export function registerHook<
+	Key extends keyof HookRegistry,
+	Fn extends typeof STATE_ENDPOINT_HOOKS[Key][number]
+>(name: Key, fn: Fn) {
+	STATE_ENDPOINT_HOOKS[name].push(fn);
+}
+
+function runHook<
+	Key extends keyof HookRegistry,
+	Args extends Parameters<typeof STATE_ENDPOINT_HOOKS[Key][number]>
```
Page-: This allows changing what `Args` is, which shouldn't be possible as it should always be precisely `Parameters<typeof STATE_ENDPOINT_HOOKS[Key][number]>`, hence why I used `args: Parameters<HookRegistry[T][number]>`

src/lib/device-state.ts
```diff
@@ -97,6 +97,33 @@ export const serviceInstallFromImage = (
...
+	deviceConfig: DeviceConfigHook[];
+}
+
+const STATE_ENDPOINT_HOOKS: HookRegistry = {
+	deviceConfig: [],
+};
+
+export function registerHook<
+	Key extends keyof HookRegistry,
+	Fn extends typeof STATE_ENDPOINT_HOOKS[Key][number]
```
Page-: Same here, this isn't really a generic and should instead be `fn: typeof STATE_ENDPOINT_HOOKS[Key][number]`

src/lib/device-state.ts
```diff
@@ -47,6 +47,35 @@ export const serviceInstallFromImage = (
...
+) {
+	STATE_ENDPOINT_HOOKS[name].push(fn);
+}
+
+function runHook<Key extends keyof HookRegistry>(
+	name: Key,
+	args: Parameters<typeof STATE_ENDPOINT_HOOKS[Key][number]>,
+) {
+	STATE_ENDPOINT_HOOKS[name].forEach(hook =>
+		hook(...(args as Parameters<typeof hook>)),
```
Page-: Is this `as` actually necessary?

dfunckt: It doesn't work without this. It was your suggestion actually, my initial implementation was doing `hook.apply` which would loose type info but didn't type coercion.

Page-: Yeah, I do prefer the explicit cast because it at least makes it clear, but wanted to check

PR #153, opened 2019-07-25
Use prepared queries for vpn client connect/disconnect

Page-:
Change-type: patch

### DIFF
--- a/src/routes/services.ts
+++ b/src/routes/services.ts
@@ -12,6 +12,36 @@ const authQuery = resinApi.prepare<{ uuid: string }>({
 		},
 	},
 });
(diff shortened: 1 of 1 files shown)


src/routes/services.ts
```diff
@@ -12,6 +12,38 @@ const authQuery = resinApi.prepare<{ uuid: string }>({
...
+});
+const clientDisconnectQuery = resinApi.prepare<{
+	uuid: string;
+	serviceId: number;
+}>({
+	method: 'PATCH',
+	resource: 'device',
+	options: {
+		$filter: {
+			$filter: {
```
dfunckt: There's two nested `$filter`s here.

PR #159, opened 2019-08-28
🗃 : Introduce renamed dt slug field for device and application resources

afitzek:
Introduce renamed dt slug field for device and application resources

### DIFF
--- a/src/hooks/common.ts
+++ b/src/hooks/common.ts
@@ -29,7 +29,7 @@ export const resolveDeviceType = (
 			}
 
 			return api.get({
-				resource: 'device_type',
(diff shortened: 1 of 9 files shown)


src/hooks/common.ts
```diff
@@ -29,7 +29,7 @@ export const resolveDeviceType = (
 			}
 
 			return api.get({
-				resource: 'device_type',
+				resource: 'device_type_table',
```
dfunckt: Is there a small window of time after the API starts and until device types are populated where this might return no results?

afitzek: I think it might.

PR #160, opened 2019-08-29
platform/auth: Type the result of findUser()

thgreasi:
Change-type: patch

### DIFF
--- a/src/platform/auth.ts
+++ b/src/platform/auth.ts
@@ -36,6 +36,16 @@ export const userHasPermission = (
 export const comparePassword = (password: string, hash: string) =>
 	sbvrUtils.sbvrTypes.Hashed.compare(password, hash);
 
+/**
+ * A known invalid comparisson to emulate a wrong password error.
+ * Used to prevent exposing information via timing attacks.
+ */
+export const runInvalidPasswordComparison = () =>
+	comparePassword(
+		'',
+		'$2b$10$Wj6ud7bYmcAw4B1uuORsnuYODUKSkrH6dVwG1zoUhDeTCjwsxlp5.',
+	);
+
 export const validatePassword = (password?: string) => {
 	if (!password) {
 		throw new Error('Password required.');
--- a/src/platform/auth.ts
+++ b/src/platform/auth.ts
@@ -1,4 +1,5 @@
 import * as _ from 'lodash';
+import { User as DbUser } from '../models';
 import { createJwt, SignOptions, User } from './jwt';
 import { retrieveAPIKey } from './api-keys';
 import { Tx, sbvrUtils, resinApi, root } from './index';
@@ -315,7 +316,7 @@ export function getUser(
 export const findUser = (
 	loginInfo: string,
 	tx?: Tx,
-): Promise<AnyObject | undefined> => {
+): Promise<Pick<DbUser, typeof $select[number]> | undefined> => {
 	if (!loginInfo) {
 		return Promise.resolve(undefined);
 	}
@@ -326,6 +327,8 @@ export const findUser = (
 	} else {
 		loginField = 'username';
 	}
+	const $select = ['id', 'actor', 'username', 'password'] as const;
+	type UserResult = Pick<DbUser, typeof $select[number]>;
 	return resinApi
 		.get({
 			resource: 'user',
@@ -344,10 +347,10 @@ export const findUser = (
 						},
 					],
 				},
-				$select: ['id', 'actor', 'username', 'password'],
+				$select: $select as Writable<typeof $select>,
 			},
 		})
-		.then(([user]: AnyObject[]) => user);
+		.then(([user]: [UserResult?]) => user);
 };
 
 export const registerUser = (
--- a/src/routes/session.ts
+++ b/src/routes/session.ts
@@ -3,6 +3,7 @@ import {
 	findUser,
 	getUser,
 	comparePassword,
+	runInvalidPasswordComparison,
 } from '../platform/auth';
 import { sbvrUtils } from '../platform';
 import { captureException, handleHttpErrors } from '../platform/errors';
@@ -41,7 +42,12 @@ export const login = (onLogin: SetupOptions['onLogin']): RequestHandler => (
 				throw new NotFoundError('User not found.');
 			}
 
-			return comparePassword(password, user.password)
+			const passwordComparison =
+				user.password == null
+					? runInvalidPasswordComparison()
+					: comparePassword(password, user.password);
+
+			return passwordComparison
 				.then(matches => {
 					if (!matches) {
 						throw new BadRequestError('Current password incorrect.');
--- a/src/models.ts
+++ b/src/models.ts
@@ -1,0 +1,18 @@
+interface WithId {
+	id: number;
+}
+
+interface WithActor {
+	actor: number;
+}
+
+interface PineResourceBase extends WithId {
+	created_at: Date;
+}
+
+export interface User extends PineResourceBase, WithActor {
+	username: string;
+	password?: string;
+	jwt_secret?: string;
+	email?: string;
+}
--- a/typings/common.d.ts
+++ b/typings/common.d.ts
@@ -6,4 +6,6 @@ declare global {
 	interface Dictionary<T> {
 		[key: string]: T;
 	}
+
+	type Writable<T> = { -readonly [K in keyof T]: T[K] };
 }


src/platform/auth.ts
```diff
@@ -315,7 +315,7 @@ export function getUser(
 export const findUser = (
 	loginInfo: string,
 	tx?: Tx,
-): Promise<AnyObject | undefined> => {
+): Promise<Pick<User, typeof $select[number]> | undefined> => {
```
thgreasi: Now that the result is typed, how about dropping the explicit return type declaration?

dfunckt: No, please don't do this.

Having explicit return types helps tremendously when reading code, because you can just anchor your attention on these and connect the dots in your head without having to *read the function body*. I find extremely annoying that TS allows implicit returns.

You also get some practical benefits:

- You rule out compiler regressions on type inference/propagation
- You rule out errors during authoring -- you typically know what you *want* the function to return but might not notice that what you return is of incorrect type and the compiler will happily accept it.
- You help the compiler get type inference in the function body right, so you don't end up having to annotate intermediate results.

thgreasi: @dfunckt just to be clear, are you fine with `Promise<Pick<User, typeof $select[number]> | undefined>`?

dfunckt: Not sure what you're asking @thgreasi ?

If it's about whether I'd rather we had the return type on the signature, then yes, I'm strongly in favour of keeping the return type and not dropping it as you suggest in your first comment, for all the reasons in my previous commnet.

If it's about whether I have any issue with that specific return type, I must admit it's quite smarter than I feel comfortable with by referencing a constant in the function body, but I'm okay with it.

src/platform/auth.ts
```diff
@@ -344,10 +346,10 @@ export const findUser = (
 						},
 					],
 				},
-				$select: ['id', 'actor', 'username', 'password'],
+				$select: $select as Writable<typeof $select>,
```
Page-: Is there a way to allow both arrays and read-only arrays in pinejs client to make this usage easier?

src/platform/jwt.ts
```diff
@@ -61,6 +61,7 @@ export interface User extends sbvrUtils.User {
 
 	twoFactorRequired?: boolean;
 	authTime?: number;
+	password?: string;
```
Page-: I don't think this should be here, this is about a req.user / jwt user which should definitely not include the password, not a pinejs user resource

src/platform/auth.ts
thgreasi: Yea, that's probably the best way to do it. I will take a look :+1:

src/platform/jwt.ts
thgreasi: @Page- thanks for the heads up. It looked weird to me as well.
Saw the `jwt_secret` in there and got confused.
I will create a new interface for the DbUser. Maybe actually this should be JwtUser and the DbUser should be just User... will check how it looks, but let me know what you think.

src/platform/auth.ts
thgreasi: @Page- I tried adding `| readonly string[]` or `| ReadonlyArray<string>` in pine client but that didn't work :(
TS gets completely crazy even after adding the above and regardless of whether we pass `$select` w/o casting or w/ `as readonly string[]` emits the following:
```
Type 'readonly ["id", "actor", "username", "password"]' is not assignable to type '{ [index: string]: "asc" | "desc"; }'.
Index signature is missing in type 'readonly ["id", "actor", "username", "password"]'.ts(2345)
```
After some testing, it looks like once it sees a readonly array it can't match it w/ `$select` any more and ends up trying it against the `[index: string]: undefined | ParameterAlias | string[] | Filter | Expand | OrderBy;` of the `ODataOptions`.
Hopefully this might get solved in a newer version.
Some alternatives that at work are:
* `$select: $select as readonly string[] as string[],`
* `$select: $select.slice(),`
* dropping the `as const` and manually typing `'id' | 'actor' | 'username' | 'password'` (but that's repeating almost the exact same text).

Give me a note on how you prefer to proceed on this.

thgreasi: An alternative that I thought is to make `findUser()` accept an optional `$select` parameter which would have the current values as the default. The typings feel better imo and also allow us to omit props as needed for each case.

src/routes/session.ts
```diff
@@ -41,6 +41,11 @@ export const login = (onLogin: SetupOptions['onLogin']): RequestHandler => (
 				throw new NotFoundError('User not found.');
 			}
 
+			if (user.password == null) {
```
Page-: @afitzek is this ok?  It'll presumably be detectable via timing

src/platform/auth.ts
```diff
@@ -344,10 +350,10 @@ export const findUser = (
 						},
 					],
 				},
-				$select: ['id', 'actor', 'username', 'password'],
+				$select,
 			},
 		})
-		.then(([user]: AnyObject[]) => user);
+		.then(([user]: [T] | []) => user);
```
Page-:
```suggestion
		.then(([user]: [T?]) => user);
```

src/platform/auth.ts
```diff
@@ -312,10 +313,14 @@ export function getUser(
...
-export const findUser = (
+export const findUser = <
+	T extends Pick<DbUser, TProps>,
+	TProps extends keyof DbUser
+>(
 	loginInfo: string,
-	tx?: Tx,
-): Promise<AnyObject | undefined> => {
+	tx: Tx | undefined = undefined,
+	$select: TProps[] = ['id', 'actor', 'username', 'password'] as TProps[],
```
Page-: The ` as TProps[]` should be unnecessary:
```suggestion
	$select: TProps[] = ['id', 'actor', 'username', 'password'],
```

thgreasi: Interestingly enough, having `const defaultFindUserProps: Array<keyof DbUser> = ['id', 'actor', 'username', 'password'];` right above would work, but TS errors w/o this.
Doing `findUser('thgreasi', undefined, ['id', 'actor', 'username', 'password'])` works fine.
That's also how we ended up typing such functions in the cloud API as well.

src/platform/auth.ts
thgreasi: I was hesitant about using this, since that's not 100% the same, but on second look I like that it's DRYer. Will change.

src/routes/session.ts
thgreasi: Does `comparePassword` also work with `null`/`undefined` values? If so, then we can definitely drop this in favor of updating the typings, rather than copying the same error message here as well.

src/platform/auth.ts
Page-: So the current typing can allow for an invalid return type, there's 2 ways of fixing this:
1. remove the $select parameter altogether since it wasn't there before and probably isn't necessary:
```ts
export const findUser = (
	loginInfo: string,
	tx: Tx | undefined = undefined,
): Promise<Pick<DbUser, typeof $select[number]> | undefined> => {
	if (!loginInfo) {
		return Promise.resolve(undefined);
	}
	const $select = ['id' as const, 'actor' as const, 'username' as const, 'password' as const];

	let loginField;
	if (loginInfo.includes('@')) {
		loginField = 'email';
	} else {
		loginField = 'username';
	}

	return resinApi
		.get({
			resource: 'user',
			passthrough: {
				req: root,
				tx,
			},
			options: {
				$filter: {
					$eq: [
						{
							$tolower: { $: loginField },
						},
						{
							$tolower: loginInfo,
						},
					],
				},
				$select,
			},
		})
		.then(([user]: [Pick<DbUser, typeof $select[number]>?]) => user);
};
```
2. use an overload
```ts
export function findUser(
	loginInfo: string,
	tx?: Tx | undefined,
	$select?: undefined,
): Promise<Pick<DbUser, 'username' | 'password' | 'id' | 'actor'> | undefined>;
export function findUser<
	T extends Pick<DbUser, TProps>,
	TProps extends keyof DbUser
>(
	loginInfo: string,
	tx: Tx | undefined,
	$select: TProps[],
): Promise<T | undefined>;
export function findUser(
	loginInfo: string,
	tx: Tx | undefined = undefined,
	$select = ['id' as const, 'actor' as const, 'username' as const, 'password' as const],
): Promise<Pick<DbUser, typeof $select[number]> | undefined> {
	if (!loginInfo) {
		return Promise.resolve(undefined);
	}

	let loginField;
	if (loginInfo.includes('@')) {
		loginField = 'email';
	} else {
		loginField = 'username';
	}

	return resinApi
		.get({
			resource: 'user',
			passthrough: {
				req: root,
				tx,
			},
			options: {
				$filter: {
					$eq: [
						{
							$tolower: { $: loginField },
						},
						{
							$tolower: loginInfo,
						},
					],
				},
				$select,
			},
		})
		.then(([user]: [Pick<DbUser, typeof $select[number]>?]) => user);
};
```

As a note I'd prefer `['id', 'actor', 'username', 'password'] as const` to `['id' as const, 'actor' as const, 'username' as const, 'password' as const]` but the trailing `as const` makes it readonly and fails to compile..

thgreasi: How about using `['id', 'actor', 'username', 'password'] as const` along with an extra `Writable` type modifier as in the initial commit, so that we can avoid repeating `as const` everywhere?
See: https://github.com/balena-io/open-balena-api/pull/160/commits/d2289415a935497d236123dde5e29250cdfef144

Page-: Yeah, I even searched for a type that was removing readonly as I'd sworn I'd seen it in use before! I guess now I know why I thought that.. :P

src/platform/auth.ts
```diff
@@ -35,6 +36,15 @@ export const userHasPermission = (
 export const comparePassword = (password: string, hash: string) =>
 	sbvrUtils.sbvrTypes.Hashed.compare(password, hash);
 
+/**
+ * A known invalid comparisson to emulate a wrong password error
```
Page-: Can you add to the comment that it's in order to prevent exposing information via timing please - the comment states *what* it does, but not *why* it's important

thgreasi: Added @Page- :+1: 
Wasn't sure if we were fine having such reasoning in the open-balena-api repo, otherwise I would have targeted for the why for sure, instead of the what.

PR #163, opened 2019-09-13
Convert application and device commit hashes to release references

Page-:
Change-type: minor

Depends on https://github.com/balena-io-modules/abstract-sql-compiler/pull/65


Page-:
This still needs a migration added before merging

PR #164, opened 2019-09-17
Add a contract JSON field to the image resource

CameronDiver:
We add a `contract` field on the `image` resource, which will be used to fill the container contract for this image at runtime. This is returned to the supervisor who will act upon it, denying or allowing the use of the image based on whether the device fulfils the requirements detailed by the contract.

Change-type: minor

### DIFF
--- a/src/resin.sbvr
+++ b/src/resin.sbvr
@@ -96,6 +96,9 @@ Term: composition
 Term: content hash
 	Concept Type: Short Text (Type)
 
+Term: contract
+	Concept Type: JSON (Type)
+
 Term: date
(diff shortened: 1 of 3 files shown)


src/resin.sbvr
```diff
@@ -90,6 +90,9 @@ Term: build log
 Term: commit
 	Concept Type: Short Text (Type)
 
+Term: contract
```
dfunckt: Nit: we generally try to maintain alphabetical order of term declarations to facilitate lookup, so this should go just after `content hash` a few lines below.

PR #167, opened 2019-09-19
🗃 💥: Renaming dt table

afitzek:
This brings the device type model into
it's final state.

Change-type: minor

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -7,7 +7,7 @@ export = {
 		{
 			username: 'guest',
 			password: ' ',
-			permissions: ['resin.device_type_table.get'],
+			permissions: ['resin.device_type.get'],
 		},
 	],
 } as Config;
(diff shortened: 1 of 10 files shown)


src/migrations/00013-add-device-types.sql
```diff
@@ -1,29 +1,28 @@
 -- create device types table
 
-CREATE TABLE IF NOT EXISTS "device type" (
+CREATE TABLE IF NOT EXISTS "device type table" (
```
dfunckt: Ouch. How did this slip though? Can you please make this a separate PR that is merged as soon as possible? The API is completely broken now.

afitzek: @dfunckt  
Here is the PR that fixes this: https://github.com/balena-io/open-balena-api/pull/179

I also realized that we have two 00014 migrations, one of them should become 00015.

PR #171, opened 2019-09-19
Allow scoped access tokens to use the whoami endpoint

### LINKED ISSUE #166: Allow scoped access token to retrieve user data from /user/v1/whoami
If a scoped access token has access to the user that corresponds to the `actor` property in the JWT, you should be able to retrieve the user details from the `/user/v1/whoami` endpoint.

LucianBuzzo:
Fixes #166

Change-type: minor
Signed-off-by: Lucian <lucian.buzzo@gmail.com>

### DIFF
--- a/src/routes/session.ts
+++ b/src/routes/session.ts
@@ -1,10 +1,12 @@
+import * as _ from 'lodash';
 import {
 	loginUserXHR,
 	findUser,
 	getUser,
 	comparePassword,
 } from '../platform/auth';
-import { sbvrUtils } from '../platform';
+import { User as DbUser } from '../models';
+import { resinApi, root, sbvrUtils } from '../platform';
 import { captureException, handleHttpErrors } from '../platform/errors';
 import { resetCounter } from '../lib/rate-limiting';
 import { RequestHandler } from 'express';
@@ -12,18 +14,64 @@ import { SetupOptions } from '..';
 
 const { BadRequestError, NotFoundError } = sbvrUtils;
 
-export const whoami: RequestHandler = (req, res) =>
-	getUser(req)
-		.then(user =>
-			res.send({ id: user.id, username: user.username, email: user.email }),
-		)
-		.catch(err => {
-			if (handleHttpErrors(req, res, err)) {
-				return;
+export const whoami: RequestHandler = async (req, res) => {
+	try {
+		const user = await getUser(req, true);
+
+		if (user.actor && user.email === undefined) {
+			// If we get to this point, then the request is most likely being made
+			// using a scoped access token.
+			// First we check to see if this token can access the user resource,
+			// then if it can, we retrieve user document that corresponds to the
+			// `actor` value on the token using root privileges.
+			const [userWithId] = (await resinApi.get({
+				resource: 'user',
+				passthrough: { req },
+				options: {
+					$filter: {
+						actor: user.actor,
+					},
+					$select: 'id',
+					$top: 1,
+				},
+			})) as Array<{ id: number }>;
+
+			// If the count is 0, then this token doesn't have access to the
+			// user resource and we should just continue with whatever was
+			// provided from the `getUser` call above.
+			if (userWithId) {
+				// If we reach this step, then the token has access and we can
+				// retrieve the user document in full
+				const { id, username, email } = (await resinApi.get({
+					resource: 'user',
+					passthrough: { req: root },
+					id: userWithId.id,
+					options: {
+						$select: ['id', 'username', 'email'],
+					},
+				})) as Pick<DbUser, 'id' | 'username' | 'email'>;
+
+				return res.send({
+					id,
+					username,
+					email,
+				});
 			}
-			captureException(err, 'Error while getting user info', { req });
-			res.sendStatus(500);
+		}
+
+		return res.send({
+			id: user.id,
+			username: user.username,
+			email: user.email,
 		});
+	} catch (err) {
+		if (handleHttpErrors(req, res, err)) {
+			return;
+		}
+		captureException(err, 'Error while getting user info', { req });
+		res.sendStatus(500);
+	}
+};
 
 export const login = (onLogin: SetupOptions['onLogin']): RequestHandler => (
 	req,
--- a/test/03-session.ts
+++ b/test/03-session.ts
@@ -1,0 +1,75 @@
+import 'mocha';
+import { expect } from 'chai';
+import { app } from '../init';
+import { SUPERUSER_EMAIL, SUPERUSER_PASSWORD } from '../src/lib/config';
+import { createScopedAccessToken } from '../src/platform/jwt';
+
+import supertest = require('./test-lib/supertest');
+
+describe('session', () => {
+	before(async function() {
+		this.token = (await supertest(app)
+			.post('/login_')
+			.send({
+				username: SUPERUSER_EMAIL,
+				password: SUPERUSER_PASSWORD,
+			})
+			.expect(200)).text;
+	});
+
+	it('/user/v1/whoami returns a user', async function() {
+		const user = (await supertest(app, this.token)
+			.get('/user/v1/whoami')
+			.expect(200)).body;
+
+		expect(user)
+			.to.have.property('id')
+			.that.is.a('number');
+		expect(user.username).to.equal('admin');
+		expect(user.email).to.equal(SUPERUSER_EMAIL);
+	});
+
+	it('/user/v1/whoami returns a user when using a correctly scoped access token', async function() {
+		const record = (await supertest(app, this.token)
+			.get("/v5/user?$filter=username eq 'admin'")
+			.expect(200)).body.d[0];
+
+		// Create a token that only has access to the granting users document
+		const accessToken = createScopedAccessToken({
+			actor: record.actor,
+			permissions: ['resin.user.get?actor eq @__ACTOR_ID'],
+			expiresIn: 60 * 10,
+		});
+
+		const user = (await supertest(app, accessToken)
+			.get('/user/v1/whoami')
+			.expect(200)).body;
+
+		expect(user)
+			.to.have.property('id')
+			.that.is.a('number');
+		expect(user.username).to.equal('admin');
+		expect(user.email).to.equal(SUPERUSER_EMAIL);
+	});
+
+	it('/user/v1/whoami returns a 401 error when using a scoped access token that does not have user permissions', async function() {
+		const record = (await supertest(app, this.token)
+			.get("/v5/user?$filter=username eq 'admin'")
+			.expect(200)).body.d[0];
+
+		const permissions = [
+			'resin.application.get?belongs_to__user/any(u:u/eq @__ACTOR_ID)',
+		];
+
+		// Create a token that only has access to the granting users applications
+		const accessToken = createScopedAccessToken({
+			actor: record.actor,
+			permissions,
+			expiresIn: 60 * 10,
+		});
+
+		await supertest(app, accessToken)
+			.get('/user/v1/whoami')
+			.expect(401);
+	});
+});
--- a/automation/common.sh
+++ b/automation/common.sh
@@ -81,6 +81,8 @@ runapi () {
 		-e VPN_PORT=5433 \
 		-e VPN_SERVICE_API_KEY=vpn_service_api_key \
 		-e VPN_SERVICE_CONNECTIONS=5 \
+		-e SUPERUSER_EMAIL=balena-admin@example.com\
+		-e SUPERUSER_PASSWORD=foobarbaz\
 		$extra_vol_args \
 		$extra_env_args \
 		$image_name


src/routes/session.ts
```diff
@@ -4,26 +4,69 @@ import {
...
+			// user resource and we should just continue with whatever was
+			// provided from the `getUser` call above.
+			if (count > 0) {
+				// If we reach this step, then the token has access and we can
+				// retrieve the user document in full
+				const [record] = (await resinApi.get({
+					resource: 'user',
+					passthrough: { req: root },
+					options: {
+						$filter: {
```
afitzek: Please add a `$select` here and only query the field from this list: https://github.com/balena-io/open-balena-api/blob/master/src/platform/auth.ts#L141 here. So that the object looks the same as from the `getUser()` function. There is no need to fetch the full row.

src/routes/session.ts
```diff
@@ -4,26 +4,69 @@ import {
...
+		// If the user email is available, then the record was successfully
+		// populated by the passport middleware
+		if (user.email == null) {
+			// If we get to this point, then the request is most likely being made
+			// using a scoped access token.
+			// First we check to see if this token can access the user resource,
+			// then if it can, we retrieve user document that corresponds to the
+			// `actor` value on the token using root privileges.
+			const count = await resinApi.get({
+				resource: 'user/$count',
```
afitzek: We could get rid fo the `$count` here and add a `$select: 'id'` into the options. So that we already have the id of the potential user.

src/routes/session.ts
afitzek: Instead of the `$filter` on the actor we could select the id in the first call and change this to something like this:
```
const [record] = (await resinApi.get({
	resource: 'user',
	passthrough: { req: root },
        id: user.id
	options: {
		$select: userFields
	}
```

src/routes/session.ts
LucianBuzzo: @afitzek this won't work if the scoped access token lets you see multiple users

src/routes/session.ts
LucianBuzzo: @afitzek as above, I don't think this will work if the scoped access token lets you see multiple users

src/routes/session.ts
afitzek: It will work, because in the request we would keep the filter against the actor. But we can save the filter against the actor in the second request and use the primary key. There cannot be two users with the same actor.

LucianBuzzo: aha I understand, I'll make the optimization :+1:

test/03-session.ts
```diff
@@ -0,0 +1,80 @@
...
+		expect(user)
+			.to.have.property('id')
+			.that.is.a('number');
+		expect(user.username).to.equal('admin');
+		expect(user.email).to.equal(SUPERUSER_EMAIL);
+	});
+
+	it('/user/v1/whoami returns a 401 error when using a scoped access token that does not have user permissions', async function() {
+		const record = await supertest(app, this.token)
+			.get('/v5/user?$filter=username%20eq%20%27admin%27')
```
Page-: I think you can just write this as straight
```suggestion
			.get("/v5/user?$filter=username eq 'admin'")
```
and it'll work, and be much more readable
(and similar for other cases)

test/03-session.ts
```diff
@@ -0,0 +1,80 @@
...
+		const accessToken = createScopedAccessToken({
+			actor: record.actor,
+			permissions: ['resin.user.get?actor eq @__ACTOR_ID'],
+			expiresIn: 60 * 10,
+		});
+
+		const user = await supertest(app, accessToken)
+			.get('/user/v1/whoami')
+			.expect(200)
+			.then(res => res.body);
```
Page-: There's no need for this `.then`:
```ts
		const { body: user } = await supertest(app, accessToken)
			.get('/user/v1/whoami')
			.expect(200)
```
or
```ts
		const user = (await supertest(app, accessToken)
			.get('/user/v1/whoami')
			.expect(200)).body
```
(and similar for other cases)

automation/fasttest.sh
```diff
@@ -61,7 +61,9 @@ if [ -z "$db_id" ] || [ -z "$redis_id" ] || [ -z "$api_id" ]; then
 	redis_id=$(runredis '-p 6378:6379')
 	echo $redis_id >>"$CONFIG_FILE"
 
-	api_id=$(runapi $IMAGE_NAME $db_id $redis_id " -p 9228:9229 -v $(pwd):/usr/src/app")
+	# Inject the SUPERUSER_EMAIL and SUPERUSER_PASSWORD env vars so that test
+	# suites have a set of user credentials to work with
+	api_id=$(runapi $IMAGE_NAME $db_id $redis_id " -e SUPERUSER_EMAIL=test@example.com -e SUPERUSER_PASSWORD=foobarbaz -p 9228:9229 -v $(pwd):/usr/src/app")
```
Page-: These options should be in `runapi` in common.sh instead

LucianBuzzo: @Page-  Is `runapi` ever used for non-testing purposes?

Page-: Nope, it's specifically for testing; `common.sh` is for things shared between ci testing in `test.sh` and local `fasttest.sh` (which just skips some steps like linting that aren't useful in the context of locally iterating on a test)

src/routes/session.ts
```diff
@@ -4,26 +4,69 @@ import {
...
+
+		// If the user email is available, then the record was successfully
+		// populated by the passport middleware
+		if (user.email == null) {
+			// If we get to this point, then the request is most likely being made
+			// using a scoped access token.
+			// First we check to see if this token can access the user resource,
+			// then if it can, we retrieve user document that corresponds to the
+			// `actor` value on the token using root privileges.
+			const results = await resinApi.get({
```
Page-: Why do we retrieve the matching resource with root privileges?  Doesn't the fact that this query succeeds mean that the token itself could fetch them directly, ie:
```suggestion
			const result = await resinApi.get({
				resource: 'user',
				passthrough: { req },
				options: {
					$filter: {
						actor: user.actor,
					},
					$select: ['id', 'username', 'email'],
				},
			});
			if (result) {
				user = result
			}

File: src/routes/session.ts [hunk 7]
```diff
@@ -4,26 +4,69 @@ import {
...
+				const id = results[0].id;
+
+				user = (await resinApi.get({
+					resource: 'user',
+					passthrough: { req: root },
+					id,
+					options: {
+						$select: ['id', 'username', 'email'],
+					},
+				})) as AnyObject[];
```
Page-: This should actually be
```suggestion
				})) as AnyObject;
```
but my other suggestion should remove the need for this query altogether

src/routes/session.ts
LucianBuzzo: @Page- we can't simulate the `/whoami` response for users with scoped access tokens without fetching by root if an implementation blacklists the email field, which is what we do on balenaCloud.

src/routes/session.ts
```diff
@@ -5,26 +5,69 @@ import {
...
+				const id = results[0].id;
+
+				user = (await resinApi.get({
+					resource: 'user',
+					passthrough: { req: root },
+					id,
+					options: {
+						$select: ['id', 'username', 'email'],
+					},
+				})) as AnyObject;
```
thgreasi:
```suggestion
				})) as Promise<Pick<DbUser, 'id' | 'username' | 'email'>>;
```
Where the DbUser is `import { User as DbUser } from '../platform/models';`

src/routes/session.ts
```diff
@@ -5,26 +5,69 @@ import {
...
+			// then if it can, we retrieve user document that corresponds to the
+			// `actor` value on the token using root privileges.
+			const results = await resinApi.get({
+				resource: 'user',
+				passthrough: { req },
+				options: {
+					$filter: {
+						actor: user.actor,
+					},
+					$select: 'id',
```
thgreasi: `
```suggestion
					$select: 'id',
					$top: 1,
```

src/routes/session.ts
```diff
@@ -5,26 +5,69 @@ import {
...
+
+		// If the user email is available, then the record was successfully
+		// populated by the passport middleware
+		if (user.email == null) {
+			// If we get to this point, then the request is most likely being made
+			// using a scoped access token.
+			// First we check to see if this token can access the user resource,
+			// then if it can, we retrieve user document that corresponds to the
+			// `actor` value on the token using root privileges.
+			const results = await resinApi.get({
```
thgreasi: How about:
```suggestion
			const [result] = await resinApi.get({
```
? Or even `userActor`?

src/routes/session.ts
```diff
@@ -3,28 +3,70 @@ import {
...
+
+		// If the user email is available, then the record was successfully
+		// populated by the passport middleware
+		if (user.email == null) {
+			// If we get to this point, then the request is most likely being made
+			// using a scoped access token.
+			// First we check to see if this token can access the user resource,
+			// then if it can, we retrieve user document that corresponds to the
+			// `actor` value on the token using root privileges.
+			const results = (await resinApi.get({
```
thgreasi: How about:
```suggestion
			const [result] = await resinApi.get({
```
?
```suggestion
			const [userWithId] = (await resinApi.get({
```

src/routes/session.ts
```diff
@@ -3,28 +3,70 @@ import {
...
-		.catch(err => {
-			if (handleHttpErrors(req, res, err)) {
-				return;
+export const whoami: RequestHandler = async (req, res) => {
+	try {
+		let user: AnyObject = await getUser(req);
+
+		// If the user email is available, then the record was successfully
+		// populated by the passport middleware
+		if (user.email == null) {
```
thgreasi: Keep in mind that in the cloud API we do have users with a `null` email.

Just wanted to clarify that the `most likely` part since I don't see anything wrong on running this extra code anyway, but I would prefer it if there was a different prop that we could check. Is there any chance this would be `undefined` when from a scoped api key instead of `null` when the user genuinely doesn't have an email set?

src/routes/session.ts
```diff
@@ -3,28 +3,70 @@ import {
...
+			const results = (await resinApi.get({
+				resource: 'user',
+				passthrough: { req },
+				options: {
+					$filter: {
+						actor: user.actor,
+					},
+					$select: 'id',
+				},
+			})) as AnyObject[];
```
thgreasi:
```suggestion
			})) as Array<{ id: number }?>;
```
Or a `Pick<>` just like below.

src/routes/session.ts
LucianBuzzo: Nice!

src/routes/session.ts
LucianBuzzo: @thgreasi thanks for the heads up. With a scoped access token, the `email` value will always be `undefined`, so we can just check for that

src/routes/session.ts
```diff
@@ -5,26 +5,71 @@ import {
...
-	getUser(req)
-		.then(user =>
-			res.send({ id: user.id, username: user.username, email: user.email }),
-		)
-		.catch(err => {
-			if (handleHttpErrors(req, res, err)) {
-				return;
+export const whoami: RequestHandler = async (req, res) => {
+	try {
+		let user: AnyObject = await getUser(req);
```
thgreasi: Would it work if we also change this to:
```suggestion
		let user: User & { email?: string } = await getUser(req);
```
Where User is `import { User } from '../platform/jwt';`
PS: we define `type Overwrite<T, U> = Pick<T, Exclude<keyof T, keyof U>> & U;` in common.d.ts and use it here.

src/routes/session.ts
```diff
@@ -1,30 +1,78 @@
...
-			res.send({ id: user.id, username: user.username, email: user.email }),
-		)
-		.catch(err => {
-			if (handleHttpErrors(req, res, err)) {
-				return;
+export const whoami: RequestHandler = async (req, res) => {
+	const userFields = ['id', 'username', 'email'];
+
+	try {
+		let user = _.pick(await getUser(req), userFields);
```
thgreasi: :grey_exclamation:  This `pick` actually omits the `actor` that is required later.

src/routes/session.ts
```diff
@@ -1,30 +1,78 @@
...
-export const whoami: RequestHandler = (req, res) =>
-	getUser(req)
-		.then(user =>
-			res.send({ id: user.id, username: user.username, email: user.email }),
-		)
-		.catch(err => {
-			if (handleHttpErrors(req, res, err)) {
-				return;
+export const whoami: RequestHandler = async (req, res) => {
+	const userFields = ['id', 'username', 'email'];
```
thgreasi: How about
```suggestion
	const userFields = ['id', 'username', 'email'] as const;
```
so that `pick` gives something better than just a `Partial<User>`?
PS: I'm not sure whether the `User` typings are correct on the `email` being non-nullable.

src/routes/session.ts
Page-: Right, and there's ways to bypass the whitelist for internal requests but open api wouldn't know how to do that

src/routes/session.ts
```diff
@@ -1,29 +1,79 @@
...
-		.catch(err => {
-			if (handleHttpErrors(req, res, err)) {
-				return;
+export const whoami: RequestHandler = async (req, res) => {
+	try {
+		const user = await getUser(req);
+
+		// If the user email is available, then the record was successfully
+		// populated by the passport middleware
+		if (user.email !== undefined) {
```
Page-: `user` can be undefined here, in which case this would throw an error of trying to access a property on undefined.  I get that the try/catch will technically handle that but I feel like that's by accident rather than design - if you want to ensure it's not undefined then you can use
```ts
const user = await getUser(req, true);

File: src/routes/session.ts [hunk 18]
```diff
@@ -1,29 +1,79 @@
...
+			// using a scoped access token.
+			// First we check to see if this token can access the user resource,
+			// then if it can, we retrieve user document that corresponds to the
+			// `actor` value on the token using root privileges.
+			const [userWithId] = (await resinApi.get({
+				resource: 'user',
+				passthrough: { req },
+				options: {
+					$filter: {
+						actor: user.actor,
```
Page-: I'm pretty sure `user.actor` can be undefined so imo we should check that it's not before running this query

PR #172, opened 2019-09-23
Fix password comparisons with null hash

thgreasi:
Further typing `findUser()` calls revealed more places that we were comparing the password with a potentially null hash.
I now think that backing the null hash handling in `comparePassword` is more scalable and covers all possibly unhanded cases in the cloud API as well.
Also updated the typings to use `null` instead of `undefined`, to correctly match pine's results.

Change-type: patch
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/g2iRNTO3hHxm8Z9pQqsyYzaA8Sk

### DIFF
--- a/src/platform/auth.ts
+++ b/src/platform/auth.ts
@@ -33,18 +33,20 @@ export const userHasPermission = (
 	return user.permissions.includes(permission);
 };
 
-export const comparePassword = (password: string, hash: string) =>
-	sbvrUtils.sbvrTypes.Hashed.compare(password, hash);
-
 /**
  * A known invalid comparisson to emulate a wrong password error.
  * Used to prevent exposing information via timing attacks.
  */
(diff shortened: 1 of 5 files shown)


src/platform/auth.ts
```diff
@@ -86,10 +88,11 @@ export const updatePasswordIfNeeded = (
 	newPassword: string,
 	tx?: Tx,
 ): Promise<boolean> => {
-	return findUser(usernameOrEmail, tx).then((user: AnyObject | undefined) => {
+	return findUser(usernameOrEmail, tx).then(user => {
```
thgreasi: Many places that you see typing changes (either adding explicit types or removing `AnyObject`) it revealed that we were possibly checking a null password and as a result I decided to embed the null check in the `comparePassword` method.

PR #173, opened 2019-09-24
Dont send a null contract from the state endpoint

CameronDiver:
Change-type: patch


CameronDiver:
@balena-ci retest

nazrhom:
@balena-ci retest

Issue #174, opened 2019-09-24
Fix device type alias resolution


afitzek:

thgreasi:
Fixed by #https://github.com/balena-io/open-balena-api/pull/175

PR #175, opened 2019-09-25
🐛: All device type slug lookups

afitzek:
Change-type: patch
Connects-to: #174


src/lib/device-types/index.ts
```diff
@@ -342,26 +342,70 @@ function getDeviceTypes(): Promise<Dictionary<DeviceTypeInfo>> {
...
-			options,
+	return (
+		Promise.try(() => {
+			if (slugs) {
+				// We need to translate incomming slug aliases to non alias slugs to
+				// search the database
+				const slugSet = new Set(slugs);
+				return getDeviceTypes()
+					.then(deviceTypesInfos => {
+						return _(deviceTypesInfos)
```
thgreasi: I think this can be simpler:
```ts
const nonAliasSlugs = slugs.reduce((acc, s) => {
  const deviceTypesInfo = deviceTypesInfos[s];
  if (deviceTypesInfo) {
    acc.push(deviceTypesInfo.latest.deviceType.slug)
  }
  return acc;
}, []);
```

thgreasi:
As noted in FD, maybe we could just change `findDeviceTypeInfoBySlug` to something like:
```ts
export const findDeviceTypeInfoBySlug = (
    api: PinejsClient,
    slug: string,
): Promise<DeviceTypeInfo> =>
        getDeviceTypes()
        .then(deviceTypeInfos => {
            // the slug can be an alias,
            // since the Dictionary also has props for the aliases
            const deviceTypeInfo = deviceTypeInfos[slug];
            if (!deviceTypeInfo || !deviceTypeInfo.latest) {
                throw new UnknownDeviceTypeError(slug);
            }

            return getAccessibleSlugs(api, [deviceTypeInfo.latest.deviceType.slug])
            .then(([acessibleSlug]) => {
                if (!acessibleSlug) {
                    // We cannot access the device type
                    throw new UnknownDeviceTypeError(slug);
                }
            })
            .return(deviceTypeInfo);
        });
```
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/ONaFKPYrxWlaua9Lum6HNEGNuhH

This way we also keep `getAccessibleSlugs` simpler & not depending on `getDeviceTypes`.

PR #182, opened 2019-10-03
🐛: Add explaining error for failed application deletions

afitzek:
If devices where moved between applications, but are still
running a release that belongs to the old application,
the old application cannot be deleted because of a foreign
key relation, between the device and the release `is running-release`.

Change-type: patch

### DIFF
--- a/src/hooks/resources/application.ts
+++ b/src/hooks/resources/application.ts
@@ -149,19 +149,73 @@ sbvrUtils.addPureHook('DELETE', 'resin', 'application', {
 				}
 			}
 			if (appIds.length > 0) {
-				// We need to null `should_be_running__release` or otherwise we have a circular dependency and cannot delete either
+				// find devices which are
+				// not part of any of the applications that are about to be deleted
+				// but run a release that belongs to any of the applications that
+				// is about to be deleted
 				return args.api
-					.patch({
-						resource: 'application',
+					.get({
+						resource: 'device',
+						passthrough: {
+							req: root,
+						},
 						options: {
+							$select: ['uuid'],
 							$filter: {
-								id: { $in: appIds },
-								should_be_running__release: { $ne: null },
+								$not: {
+									belongs_to__application: {
+										$in: appIds,
+									},
+								},
+								is_running__release: {
+									$any: {
+										$alias: 'r',
+										$expr: {
+											r: {
+												belongs_to__application: {
+													$in: appIds,
+												},
+											},
+										},
+									},
+								},
 							},
 						},
-						body: { should_be_running__release: null },
 					})
-					.return();
+					.then((devices: AnyObject[]) => {
+						if (devices.length === 0) {
+							return;
+						}
+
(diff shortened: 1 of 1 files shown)


src/hooks/resources/application.ts
```diff
@@ -149,19 +149,69 @@ sbvrUtils.addPureHook('DELETE', 'resin', 'application', {
...
+							return;
+						}
+
+						throw new BadRequestError('updateRequired', {
+							error: 'updateRequired',
+							message: `Can't delete applications ${_.join(
+								appIds,
+								', ',
+							)}! The following devices where moved from these applications and are still running releases that belong to these applications: ${_.join(
+								devices.map(({ uuid }) => uuid),
```
Page-: Could you do this `.map` just the once:
```ts
const uuids = devices.map(({ uuid }) => uuid);
						throw new BadRequestError('updateRequired', {
							error: 'updateRequired',
							message: `Can't delete applications ${_.join(
								appIds,
								', ',
							)}! The following devices where moved from these applications and are still running releases that belong to these applications: ${uuids.join(', ')}`,
							appids: appIds,
							uuids,
						});

File: src/hooks/resources/application.ts [hunk 2]
```diff
@@ -149,19 +149,73 @@ sbvrUtils.addPureHook('DELETE', 'resin', 'application', {
...
-					.return();
+					.then((devices: AnyObject[]) => {
+						if (devices.length === 0) {
+							return;
+						}
+
+						const uuids = devices.map(({ uuid }) => uuid);
+						throw new BadRequestError('updateRequired', {
+							error: 'updateRequired',
+							message: `Can't delete applications ${_.join(
```
dfunckt:
```suggestion
							message: `Can't delete application(s) ${_.join(
```

src/hooks/resources/application.ts
```diff
@@ -149,19 +149,73 @@ sbvrUtils.addPureHook('DELETE', 'resin', 'application', {
...
+								appIds,
+								', ',
+							)}! The following devices where moved from these applications and are still running releases that belong to these applications: ${_.join(
```
dfunckt:
```suggestion
							)} because the following devices are still running releases that belong to these applications: ${_.join(
```

src/hooks/resources/application.ts
```diff
@@ -149,19 +149,73 @@ sbvrUtils.addPureHook('DELETE', 'resin', 'application', {
...
+							$select: ['uuid'],
 							$filter: {
-								id: { $in: appIds },
-								should_be_running__release: { $ne: null },
+								belongs_to__application: {
+									$not: {
+										$in: appIds,
+									},
+								},
+								is_running__release: {
```
dfunckt: You need to `$or` `should_be_running__release` as well, right?

afitzek: `should_be_running__release` is cleared when a device is moved.

PR #188, opened 2019-10-08
✨: Use GH contracts repo to find device types

afitzek:
Change-type: minor


src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,46 @@
...
+	return Promise.try(() => {
+		return octokit.repos.getContents(realOptions).then(dtSlugs => {
+			if (_.isArray(dtSlugs.data)) {
+				dtSlugs.data.forEach(element => {
+					if (element.type == 'dir') {
+						dts.push(element.name);
+					}
+				});
+			} else {
+				console.error(
```
richbayliss: Is this the correct way to handle such an error; do we not want to `captureException()`?

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,46 @@
...
+	const octokit = new Octokit();
+	const dts: string[] = [];
+	return Promise.try(() => {
+		return octokit.repos.getContents(realOptions).then(dtSlugs => {
+			if (_.isArray(dtSlugs.data)) {
+				dtSlugs.data.forEach(element => {
+					if (element.type == 'dir') {
+						dts.push(element.name);
+					}
+				});
```
richbayliss: Could do this in a filter?

```suggestion
    dts = dtSlugs.data.filter(e => e.type == 'dir').map(e => e.name);
```

src/lib/device-types/index.ts
```diff
@@ -124,52 +125,74 @@ function fetchDeviceTypes(): Promise<Dictionary<DeviceTypeInfo>> {
...
-					if (_.isEmpty(builds)) {
-						return;
-					}
-
-					const sortedBuilds = sortBuildIds(builds);
-					return getFirstValidBuild(slug, sortedBuilds).then(
-						latestBuildInfo => {
-							if (!latestBuildInfo) {
+	// fetch device types from GH
+	return fetchExternalDeviceTypes()
```
thgreasi: How about a `Promise.all`?

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,42 @@
...
+	owner: string;
+	repo: string;
+	path: string;
+	ref?: string;
+}
+
+export const fetchExternalDeviceTypes = (
+	options?: FetchDeviceTypeOptions,
+): Promise<string[]> => {
+	const realOptions = options || {
```
thgreasi: This could be a default value in the argument, except if you were meant to do `_.defaults(options || {}, {...})` :thinking:

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,42 @@
...
+	return Promise.try(() => {
+		return octokit.repos.getContents(realOptions).then(dtSlugs => {
+			if (_.isArray(dtSlugs.data)) {
+				return dtSlugs.data.filter(e => e.type == 'dir').map(e => e.name);
+			}
+			const err = new Error(
+				`Unexpected response from GH: ${realOptions.owner}/${realOptions.repo}/${realOptions.path} should be a directory!`,
+			);
+			captureException(err, err.message);
+			return [];
```
thgreasi: Is it fine that we don't throw the error here, and we resolve the promise instead?

test/test-lib/github-mock.ts
```diff
@@ -0,0 +1,52 @@
+import * as Octokit from '@octokit/rest';
+import * as Promise from 'bluebird';
```
thgreasi: Does `Octokit` return a Bluebird enhanced promises?
If not then we should better remove the Bluebird import, to have more accurate test cases about the Promise types used.

test/test-lib/github-mock.ts
```diff
@@ -0,0 +1,52 @@
...
+
+			if (!_.has(getContent, contentKey)) {
+				return Promise.reject(
+					new Error(
+						`gh mock: getContents could not find a mock for ${contentKey}`,
+					),
+				);
+			}
+
+			const response: Octokit.Response<Octokit.ReposGetContentsResponse> = {
```
thgreasi: Maybe we could just type it as 
```suggestion
			const response: Pick<Octokit.Response<Octokit.ReposGetContentsResponse>, 'data' | 'status' | 'headers'> = {
```
and
* add a comment that these are the only props that we need from the mock
* drop the extra `as Octokit.Response<Octokit.ReposGetContentsResponse>`

test/test-lib/github-mock.ts
```diff
@@ -0,0 +1,52 @@
...
+					'x-ratelimit-reset': '1570549993',
+					'x-Octokit-request-id': '123345',
+					'x-Octokit-media-type': 'github.v3; format=json',
+					link: '',
+					'last-modified': 'Wed, 02 Oct 2019 12:09:34 GMT',
+					etag: '63b1e55e670df084b122deba00768fe52ad424b7',
+					status: '',
+				},
+			} as Octokit.Response<Octokit.ReposGetContentsResponse>;
+			return Promise.resolve(response);
```
thgreasi: If we are going to drop the bluebird import, then how about changing this to just `return response` and marking the method as `async`?

test/test-lib/github-mock.ts
```diff
@@ -0,0 +1,52 @@
...
+		},
+	};
+}
+
+export const GHSdkMock = GHMock;
```
thgreasi: Do we really need this export?

src/lib/device-types/storage/gh.ts
afitzek: Yes, it will catch up later, but we might hit GH rate limits or similar.

src/lib/device-types/index.ts
```diff
@@ -124,52 +125,74 @@ function fetchDeviceTypes(): Promise<Dictionary<DeviceTypeInfo>> {
...
+				})
+				.then((dbSlugs: Array<AnyObject>) => {
+					// merge them into a unique list of slugs
+					return _.uniq(_.concat(slugs, _.map(dbSlugs, 'slug')));
+				});
+		})
+		.then((dtList: string[]) => {
+			return Promise.resolve(dtList)
+				.map(slug => {
+					return Promise.try(() => listFolders(getImageKey(slug)))
```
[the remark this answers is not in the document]
afitzek: No without the try the system can fail in listFolders.

test/test-lib/github-mock.ts
afitzek: That's why I wrap the response in a Promise.try to have Bluebird promises to work with.

src/lib/device-types/index.ts
afitzek: You might mean `Promise.join`?

afitzek:
@balena-ci retest

afitzek:
@balena-ci retest

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,123 @@
...
+): Promise<string[]> => {
+	const octokit = new Octokit({
+		// ToDo: use current version string in UA
+		userAgent: `OpenBalenaApi`,
+	});
+	options.archive_format = 'tarball';
+	try {
+		const branch = await octokit.repos.getBranch({
+			owner: options.owner!,
+			repo: options.repo!,
```
Page-: owner/repo are specified as required so the `!` is redundant:
```suggestion
			owner: options.owner,
			repo: options.repo,
```

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,123 @@
...
+			branch: options.ref!,
+		});
+
+		if (branch.data.commit.sha !== currentHash) {
+			const archiveLinkResponse: AnyObject = await octokit.repos.getArchiveLink(
+				{
+					owner: options.owner!,
+					repo: options.repo!,
```
Page-:
```suggestion
					owner: options.owner,
					repo: options.repo,
```

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,123 @@
...
+	const octokit = new Octokit({
+		// ToDo: use current version string in UA
+		userAgent: `OpenBalenaApi`,
+	});
+	options.archive_format = 'tarball';
+	try {
+		const branch = await octokit.repos.getBranch({
+			owner: options.owner!,
+			repo: options.repo!,
+			branch: options.ref!,
```
Page-: As far as I can see `options.ref` can definitely be null based on the typings so the `!` looks like a mistake/bug?

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,123 @@
...
+				{
+					owner: options.owner!,
+					repo: options.repo!,
+					ref: options.ref!,
+					archive_format: 'tarball',
+				},
+			);
+
+			const archiveUrl = archiveLinkResponse.url;
+			const extractor = tar.extract();
```
Page-: Could you move this inside the `new Bluebird` block please to make it clear it's only for use in there

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,123 @@
...
+					.pipe(extractor);
+			}).timeout(20 * 1000);
+
+			currentHash = branch.data.commit.sha;
+			cachedDeviceTypes = deviceTypes;
+		}
+	} catch (e) {
+		captureException(e, 'Failed to sync device types from GH repository');
+	}
+	return cachedDeviceTypes.map(({ slug }) => slug);
```
Page-: Since we always convert the device types to an array of slugs it would make sense to cache it in that form in order to reduce memory usage for the cache and also reduce work slightly if the refetch starts failing and we fall back to using the cached device types
```suggestion
			cachedDeviceTypes = deviceTypes.map(({ slug }) => slug);
		}
	} catch (e) {
		captureException(e, 'Failed to sync device types from GH repository');
	}
	return cachedDeviceTypes;
```

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,123 @@
...
+						e,
+						'Failed to unzip stream from GH contract repositority',
+					);
+				});
+
+				request
+					.get(archiveUrl)
+					.pipe(gunzipper)
+					.pipe(extractor);
+			}).timeout(20 * 1000);
```
Page-: Can you move this timeout value to a constant please

Page-: If we time out it would be good to cancel the in-progress request

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,123 @@
...
+			const dtRegex = /.*\/contracts\/hw\.device-type\/.*\/contract\.json/;
+			const deviceTypes: DTContract[] = [];
+
+			await new Bluebird(resolve => {
+				extractor.on(`entry`, (headers, stream, next) => {
+					const chunks: Buffer[] = [];
+					if (dtRegex.test(headers.name)) {
+						stream.on('end', function() {
+							const buf = Buffer.concat(chunks);
+							const dtContract = JSON.parse(buf.toString('UTF-8'));
```
Page-: There should definitely be a try/catch around this, if we get invalid JSON back then right now it would crash the entire process

src/lib/device-types/storage/gh.ts
```diff
@@ -0,0 +1,123 @@
...
+	options.archive_format = 'tarball';
+	try {
+		const branch = await octokit.repos.getBranch({
+			owner: options.owner!,
+			repo: options.repo!,
+			branch: options.ref!,
+		});
+
+		if (branch.data.commit.sha !== currentHash) {
+			const archiveLinkResponse: AnyObject = await octokit.repos.getArchiveLink(
```
Page-: Doesn't octokit have proper typings?
```suggestion
			const archiveLinkResponse = await octokit.repos.getArchiveLink(
```

src/lib/device-types/index.ts
```diff
@@ -230,11 +252,10 @@ async function updateDTModel(
 				$filter: filter,
 			},
 		});
-		return;
 	}
 }
 
-function syncDataModel(
+async function syncDataModel(
```
Page-:
```suggestion
function syncDataModel(
```

src/lib/device-types/index.ts
```diff
@@ -125,7 +126,28 @@ async function fetchDeviceTypes(): Promise<Dictionary<DeviceTypeInfo>> {
...
+			captureException(err, 'Failed to fetch external device types');
+		}
+
+		const dbSlugs = (await api.resin.get({
+			resource: 'device_type',
+			passthrough: { req: root },
+			options: {
+				$select: ['slug'],
+			},
+		})) as AnyObject[];
```
Page-:
```suggestion
		})) as Array<{ slug: string }>;
```

sradevski:
Closing in favor of https://github.com/balena-io/open-balena-api/pull/473 as a simpler, more generic solution

PR #190, opened 2019-10-09
fix: API heartbeat allow for poll jitter

richbayliss:
The supervisor introduced a "jitter" to the supplied poll interval
value of between 0.5x and 1.5x the configured value. The state mechanims
will now always account for the 1.5x case, so devices will be considered
online for the max possible time.

For Reference: https://github.com/balena-io/balena-supervisor/pull/934


src/lib/device-online-state.ts
```diff
@@ -86,25 +86,32 @@ const getPollIntervalForParentApplication = resinApi.prepare<{ uuid: string }>({
...
-				DEFAULT_SUPERVISOR_POLL_INTERVAL,
-			);
-		});
+				return Math.max(
+					parseInt(pollIntervals[0].value, 10) || 0,
+					DEFAULT_SUPERVISOR_POLL_INTERVAL,
+				);
+			})
+			// adjust the value for the jitter in the Supervisor...
+			.then(pollInterval => pollInterval * POLL_JITTER_FACTOR)
```
richbayliss: This line caused prettier to indent 🤷‍♂

src/lib/device-online-state.ts
```diff
@@ -338,7 +345,7 @@ class DeviceOnlineStateManager {
 						uuid,
 						DeviceOnlineStates.Online,
 						DeviceOnlineStates.Timeout,
-						timeoutSeconds,
+						Math.ceil(timeoutSeconds), // always make this a whole number of seconds, and round up to make sure we dont expire too soon...
```
richbayliss: Redis keys can only expire in whole seconds, so we round up to make sure we don't expire too soon.

test/03-device-state-v2.ts
```diff
@@ -172,6 +172,9 @@ describe('Device State v2', () => {
 		});
 
 		describe('Event Tracking', () => {
+			const devicePollInterval =
```
richbayliss: Calculate the value to make the maths easier below...

PR #194, opened 2019-10-18
Fixes and optimisations for moving devices

Page-:
This fixes the case of an attempted move affected no devices and also
correctly waits for all of the image install patches to complete

Change-type: patch

### DIFF
--- a/src/hooks/resources/device.ts
+++ b/src/hooks/resources/device.ts
@@ -232,28 +232,32 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
 							throw new InaccessibleAppError();
 						}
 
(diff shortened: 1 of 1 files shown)


src/hooks/resources/device.ts
```diff
@@ -418,32 +422,24 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
 				}),
 			);
 
-			// Also mark all image installs which are part of the current
-			// application as deleted.
-			waitPromises.push(
-				(args.request.custom.devices as Promise<AnyObject[]>).map(device => {
```
Page-: This was actually an `AnyObject[]` so whilst the `.map` worked it did not wait on any of the returned promises

PR #198, opened 2019-10-24
Async/await device logs

Page-:


src/lib/device-logs/struct.ts
```diff
@@ -61,9 +60,9 @@ export type AnySupervisorLog = OldSupervisorLog | SupervisorLog;
 export type Subscription = (log: DeviceLog) => void;
 
 export interface DeviceLogsBackend {
-	history(ctx: LogContext, count: number): Promise<DeviceLog[]>;
+	history(ctx: LogContext, count: number): PromiseLike<DeviceLog[]>;
```
thgreasi: TBH, I would prefer to see `Promise` (since `hustory()` is marked as `async`) instead of `PromiseLike`, since it makes things simpler to read (ppl often are not that used to see `PromiseLike` and need more time to process it), and in case we even need to use `.catch` we will have to change this at that point.

PR #206, opened 2019-11-01
metrics: Emit events for key points in the API heartbeat flow

richbayliss:
- `change` whenever a device state should change with start/end times from the call to the DB confirm.
- `stats` every 10sec to log the current RSMQ queue stats

Change-type: patch

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -4,6 +4,7 @@ import { setup } from './src';
 import config = require('./config');
 import { version } from './package.json';
 import { sbvrUtils } from '@resin/pinejs';
+import * as deviceOnlineState from './src/lib/device-online-state';
 
 async function onInitMiddleware(app: express.Application) {
 	const { forwardRequests } = await import('./src/platform/versions');
@@ -105,8 +106,10 @@ setup(app, {
 	onInitHooks,
 })
 	.tap(createSuperuser)
-	.then(({ startServer }) => {
-		return startServer(process.env.PORT || 1337).return();
+	.then(({ startServer }) => startServer(process.env.PORT || 1337).return())
+	.then(() => {
+		console.log('Starting API heartbeat state worker...');
+		deviceOnlineState.getInstance().start();
 	})
 	.then(() => {
 		if (doRunTests) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -125,12 +125,15 @@ export function setup(app: _express.Application, options: SetupOptions) {
 		.then(() => import('./routes'))
 		.then(routes => routes.setup(app, options.onLogin))
 		.then(runSetupFunction(app, options.onInitRoutes))
-		.then(() => app.use(Raven.errorHandler()))
-		.return({
-			app,
-			startServer: _.partial(startServer, app),
-			runCommand: _.partial(runCommand, app),
-			runFromCommandLine: _.partial(runFromCommandLine, app),
+		.then(() => {
+			app.use(Raven.errorHandler());
+
+			return {
+				app,
+				startServer: _.partial(startServer, app),
+				runCommand: _.partial(runCommand, app),
+				runFromCommandLine: _.partial(runFromCommandLine, app),
+			};
 		});
 }
 
--- a/src/lib/device-online-state.ts
+++ b/src/lib/device-online-state.ts
@@ -1,4 +1,5 @@
 import * as Bluebird from 'bluebird';
+import * as _ from 'lodash';
 import { DEFAULT_SUPERVISOR_POLL_INTERVAL } from './env-vars';
 import { noop } from 'lodash';
 import { sbvrUtils } from '@resin/pinejs';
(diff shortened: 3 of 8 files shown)


src/lib/device-online-state.ts
```diff
@@ -181,6 +245,10 @@ class DeviceOnlineStateManager {
 					req: root,
 				},
 			})
+			.tap(() => {
+				const endAt = Date.now();
+				this.emit('change', { startAt, endAt, uuid, newState });
```
Page-: We should also emit on failure, but with the extra info that it failed (so we can track failure rate alongside success rate)

src/lib/device-online-state.ts
```diff
@@ -157,16 +202,35 @@ class DeviceOnlineStateManager {
 				}
 			})
 			.then(() => {
+				this.queueStatsInterval = setInterval(async () => {
```
Page-: There's no need to actually assign this anywhere, you can do `setInterval(...).unref()`

src/lib/device-online-state.ts
```diff
@@ -14,6 +14,7 @@ import {
 	API_HEARTBEAT_STATE_ENABLED,
 	API_HEARTBEAT_STATE_TIMEOUT_SECONDS,
 } from './config';
+import * as events from 'events';
```
Page-: We normally use the `eventemitter3` library

src/lib/device-online-state.ts
richbayliss: 👍

src/lib/device-online-state.ts
```diff
@@ -157,6 +202,23 @@ class DeviceOnlineStateManager {
 				}
 			})
 			.then(() => {
+				setInterval(async () => {
```
Page-: One problem with this being a `setInterval` is that if for some reason the `rsmq.getQueueAttributesAsync` starts to be slow we could pile them up with no backoff

init.ts
```diff
@@ -106,7 +107,15 @@ setup(app, {
 })
 	.tap(createSuperuser)
 	.then(({ startServer }) => {
-		return startServer(process.env.PORT || 1337).return();
+		startServer(process.env.PORT || 1337).return();
```
richbayliss: revert

Page-:
```suggestion
		return startServer(process.env.PORT || 1337).return();
```
(or drop the `{}` for implicit return)

src/index.ts
```diff
@@ -126,11 +126,13 @@ export function setup(app: _express.Application, options: SetupOptions) {
 		.then(routes => routes.setup(app, options.onLogin))
 		.then(runSetupFunction(app, options.onInitRoutes))
 		.then(() => app.use(Raven.errorHandler()))
-		.return({
-			app,
-			startServer: _.partial(startServer, app),
-			runCommand: _.partial(runCommand, app),
-			runFromCommandLine: _.partial(runFromCommandLine, app),
+		.then(() => {
```
richbayliss: revert

Page-: Can you join this with the previous `.then` please

src/lib/device-online-state.ts
richbayliss: I will implement this 👍

src/lib/device-online-state.ts
```diff
@@ -126,17 +128,65 @@ export const enum DeviceOnlineStates {
...
+}
+
+export class DeviceOnlineStateManager extends events.EventEmitter {
 	private static readonly REDIS_NAMESPACE = 'device-online-state';
 	private static readonly EXPIRED_QUEUE = 'expired';
 	private static readonly RSMQ_READ_TIMEOUT = 30;
+	private static readonly QUEUE_STATS_INTERVAL_MSEC = 10000;
+
+	private readonly featureIsEnabled: boolean;
+	private resinApi?: sbvrUtils.PinejsClient;
```
Page-: Can you remove this and use `api.resin...` instead please

src/lib/device-online-state.ts
```diff
@@ -356,4 +439,11 @@ class DeviceOnlineStateManager {
 	}
 }
 
-export const manager = new DeviceOnlineStateManager();
+let instance: DeviceOnlineStateManager | undefined = undefined;
+export const getInstance = () => {
```
Page-:
```suggestion
export const getInstance = _.once(() => new DeviceOnlineStateManager())
```

src/lib/device-online-state.ts
```diff
@@ -14,41 +15,44 @@ import {
...
-	resource: 'device_config_variable',
-	passthrough: { req: root },
-	options: {
-		$select: ['name', 'value'],
-		$top: 1,
-		$expand: {
-			device: {
-				$filter: { uuid: { '@': 'uuid' } },
+export const getPollInterval = async (uuid: string) => {
+	const { root, api } = sbvrUtils;
```
Page-: You can move this to the module level and avoid duplicating it below

src/lib/device-online-state.ts
```diff
@@ -160,45 +210,76 @@ class DeviceOnlineStateManager {
...
+				this.setupQueueStatsEmitter(
+					DeviceOnlineStateManager.QUEUE_STATS_INTERVAL_MSEC,
+				),
+			);
+	}
+
+	private async setupQueueStatsEmitter(interval: number) {
+		return setTimeout(async () => {
+			const startAt = Date.now();
+			const queueAttributes = await this.rsmq.getQueueAttributesAsync({
```
Page-: Can this fail? It's probably worth adding a try/catch with a captureException to be safe in that case, and put the `this.setupQueueStatsEmitter(interval)` in a `finally`

test/03-device-state-v2.ts
```diff
@@ -248,8 +293,16 @@ describe('Device State v2', () => {
...
 				TIMEOUT_MSEC) /
-				1000} seconds (plus 1 second to be sure)`, async () => {
-				await Bluebird.delay(devicePollInterval + TIMEOUT_MSEC + 1000);
+				1000} seconds`, async () => {
+				stateChangeEventSpy.resetHistory();
+
+				await Bluebird.delay(devicePollInterval + TIMEOUT_MSEC);
+
+				// it will be called for TIMEOUT and OFFLINE...
+				while (!stateChangeEventSpy.calledTwice) {
```
Page-: It'd be good to have a max number of loops here rather than waiting forever on failure (and in the other while cases)

test/03-device-state-v2.ts
```diff
@@ -248,8 +299,18 @@ describe('Device State v2', () => {
...
 				TIMEOUT_MSEC) /
-				1000} seconds (plus 1 second to be sure)`, async () => {
-				await Bluebird.delay(devicePollInterval + TIMEOUT_MSEC + 1000);
+				1000} seconds`, async () => {
+				stateChangeEventSpy.resetHistory();
+
+				await Bluebird.delay(devicePollInterval + TIMEOUT_MSEC);
+
+				// it will be called for TIMEOUT and OFFLINE...
+				let testLimit = 10000 / 50;
```
Page-: Could you dedupe this as
```ts
const waitFor = async (fn: () => boolean) => {
	let testLimit = 10000 / 50;
	while (!fn() && testLimit > 0) {
		await Bluebird.delay(50);
		testLimit--;
	}
}
...
await waitFor(() => stateChangeEventSpy.calledTwice)
```

richbayliss: I could

PR #207, opened 2019-11-04
Use enum definitions instead of rules

Page-:
Change-type: minor


thgreasi:
That's awesome @Page- :+1:

PR #208, opened 2019-11-06
Update typescript to 3.7

Page-:
Change-type: patch

### DIFF
--- a/src/lib/device-logs/backends/redis.ts
+++ b/src/lib/device-logs/backends/redis.ts
@@ -11,7 +11,7 @@ import {
 	Subscription,
 } from '../struct';
 import { captureException } from '../../../platform/errors';
-import { sbvrUtils } from '../../../platform';
+import { sbvrUtils } from '@resin/pinejs';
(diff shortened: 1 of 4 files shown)


src/lib/device-logs/backends/redis.ts
```diff
@@ -35,16 +35,6 @@ const schema = avro.Type.forSchema({
 	],
 });
 
-export interface DeviceLog {
```
thgreasi: I can't believe that we did hit this error, just like the announcement mentioned...

src/routes/registry.ts
```diff
@@ -194,7 +194,7 @@ const resolveAccess = async (
 const authorizeRequest = (
 	req: Request,
 	scopes: string[],
-): Resolvable<Access[]> => {
+): Access[] | PromiseLike<Access[]> => {
```
Page-: This is a weird bug where downstream a `Promise.all` type was suddenly adding `undefined` as a possible value

src/lib/device-logs/backends/redis.ts
Page-: Yep

PR #214, opened 2019-11-19
Convert routes/devices to async/await

Page-:
Change-type: patch

### DIFF
--- a/src/routes/devices.ts
+++ b/src/routes/devices.ts
@@ -1,5 +1,5 @@
 import * as _ from 'lodash';
-import * as Promise from 'bluebird';
+import * as Bluebird from 'bluebird';
 
 import {
 	captureException,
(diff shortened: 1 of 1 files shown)


src/routes/devices.ts
```diff
@@ -804,169 +786,159 @@ export const statePatch: RequestHandler = (req, res) => {
...
+					$filter: { uuid },
+				},
+			})) as AnyObject[];
+
+			if (device == null) {
+				throw new UnauthorizedError();
+			}
+
+			if (local.is_on__commit === null) {
+				deviceBody!.is_running__release = null;
```
Page-: I was hoping to get rid of this `!` and the equivalent one below, but turns out it's still required because the check is outside of the `sbvrUtils.db.transaction` and this is in the callback :(

PR #215, opened 2019-11-19
Only set the running release in state patch if it's valid

Page-:
Change-type: patch


Page-:
@CameronDiver when an application is moved between apps it might continue trying to report the old release which it no longer has access to

PR #217, opened 2019-11-19
Some async await conversion

Page-:


src/routes/auth.ts
```diff
@@ -27,12 +26,15 @@ export const getUserPublicKeys: RequestHandler = (req, res) => {
 				},
 			},
 			passthrough: { req },
-		})
-		.then((data: Array<{ public_key: string }>) => {
-			const authorizedKeys = data.map(e => e.public_key).join('\n');
-			res.status(200).send(authorizedKeys);
-		})
-		.catch(err => {
-			handleHttpErrors(req, res, err);
```
Page-: I also fixed a bug here where it was possible that we'd never respond to the request if the error wasn't a `HttpError`

src/routes/access.ts
```diff
@@ -10,92 +9,89 @@ const { UnauthorizedError, api } = sbvrUtils;
...
-						allowedDevices.d == null ||
-						!_.isArray(allowedDevices.d) ||
-						allowedDevices.d.length !== 1
-					) {
-						res.sendStatus(401);
-						return;
-					}
+		if (
+			allowedDevices.d == null ||
+			!_.isArray(allowedDevices.d) ||
```
dfunckt: You can drop the check above, since isArray will take care of it being null-ish.

src/routes/applications.ts
```diff
@@ -15,32 +13,31 @@ import { RequestHandler, Request } from 'express';
...
-			// Check that the current user has access to this application.
-			if (app == null) {
-				throw new UnauthorizedError(
-					'You do not have permission to access this application',
-				);
-			}
-			return app;
-		});
+		},
+	})) as AnyObject;
```
dfunckt: We really need to sort the types, the need to cast every single query to AnyObject is unreasonable.

Page-: It's actually the same as before in terms of casting, just the cast was non-obvious before because we don't enforce `strictFunctionTypes` and so can do it via `(x: cast) =>` in function params, and did because the true cast in promise world gets much messier than this much quicker. And ideally we will switch to enforcing `strictFunctionTypes` because the ability to implicitly cast is a potential source of non-obvious errors.
That said we do need a way to handle better types in pinejs-client, it's just the `'/$count'` stuff that makes it really hard to to type it in a way it narrows to the single return type it can be

dfunckt: The problematic case is just “.get” and we don’t *have* to overload it. We could have “get” that requires an “id” and fetches zero or one results, “find<T>” that takes no “id” and requires a “$filter” and returns a “T[]”, and finally a “count” that does a count based on given “$filter” and solves the problematic case you mention. Let’s discuss this some more in FD

PR #225, opened 2019-11-22
Add missing `$select`s

Page-:
Change-type: patch

### DIFF
--- a/src/platform/permissions.ts
+++ b/src/platform/permissions.ts
@@ -41,7 +41,7 @@ export const assignUserPermission = (
 
 // api key helpers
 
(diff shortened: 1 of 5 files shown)


src/platform/permissions.ts
```diff
@@ -41,7 +41,7 @@ export const assignUserPermission = (
 
 // api key helpers
 
-export const getOrInsertApiKey = (
+const getOrInsertApiKey = (
```
Page-: This is not actually used externally anywhere (I had to search for any usage in order to work out the $select) so I removed it to make that clearer

Issue #231, opened 2019-12-06
Service installs resources are not always created retroactivelly for new devices


CameronDiver:
When a device is added to an application, we need to create a service install for every service that is part of that application. Currently this doesn't always work.

Another option is to have the service_install resources created on-demand, but prescriptively should be enough too.

dfunckt:
#232 fixes this.

PR #232, opened 2019-12-11
Ensure a device has service_install records after being un-pinned from a release

richbayliss:
If a device is pinned to a release, and a new application release is
created, that device will now be able to run the new release if
un-pinned.

Change-type: patch

### DIFF
--- a/src/hooks/resources/device.ts
+++ b/src/hooks/resources/device.ts
@@ -442,7 +442,43 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
 		}
 
 		if (args.request.values.should_be_running__release !== undefined) {
-			// If the device has been pinned, we should alert the supervisor
+			// If the device was preloaded, and then pinned, service_installs do not exist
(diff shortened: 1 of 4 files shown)


roman-mazur:
@richbayliss would you remind why we need a service install? Do we have some logic on the backend that can select what services should not be installed on a device? Because otherwise, if all services are always installed, this DB record does not make much sense (a release reference would be enough)...

richbayliss:
@roman-mazur the `service_install` resource is used to tie services to devices, and to associate the environment variables. I have looked at dropping it out, but there would be significant impact and translating the resource to lower API levels would mean changing IDs; something we don't want.

roman-mazur:
@richbayliss do you mean there are clients that query service installs? What is the use case?

richbayliss:
@roman-mazur adding env vars for example - so yes, it's called directly in those cases to determine the ID to use 👍

roman-mazur:
@richbayliss should we close #231?

PR #235, opened 2019-12-17
Async await

Page-:


src/lib/device-types/index.ts
```diff
@@ -125,43 +125,40 @@ function fetchDeviceTypes(): Promise<Dictionary<DeviceTypeInfo>> {
...
 	getIsIgnored.clear();
 	getDeviceTypeJson.clear();
-	return listFolders(IMAGE_STORAGE_PREFIX)
-		.map(slug => {
-			return listFolders(getImageKey(slug))
-				.then(builds => {
-					if (_.isEmpty(builds)) {
+	return Promise.map(listFolders(IMAGE_STORAGE_PREFIX), slug => {
+		return listFolders(getImageKey(slug))
+			.then(builds => {
```
dfunckt: How come you didn't go all the way converting this one?

src/platform/auth.ts
```diff
@@ -179,76 +183,77 @@ export function setUserTokenDataCallback(fn: GetUserTokenDataFn) {
...
+			resource: 'user',
+			id: userId,
+			passthrough: { req: root },
+			options: {
+				$select: tokenFields,
+			},
+		}) as Promise<AnyObject>,
+		sbvrUtils.getUserPermissions(userId),
+	]);
+	// (userData: AnyObject, permissionData) => {
```
dfunckt: Stray comment.

src/platform/auth.ts
```diff
@@ -179,76 +183,77 @@ export function setUserTokenDataCallback(fn: GetUserTokenDataFn) {
...
+): Promise<void> => {
+	const token = await createSessionToken(userId, extraParams);
+	sendXHRToken(res, token, statusCode);
+};
+
+export const updateUserXHR = async (
+	res: Response,
+	req: Request,
+): Promise<void> => {
+	await getUser(req, false);
```
dfunckt: Let's add a comment here that we care about `getUser`'s side-effects. Otherwise, the fact we're passing `false` for `required` suggests we don't care if this fails which makes one wonder why this call is useful here.

src/platform/auth.ts
```diff
@@ -288,57 +293,55 @@ export function getUser(
...
 		creds?: User;
 	},
 	required = true,
 ): Promise<User | undefined> {
-	return retrieveAPIKey(req).then(() => {
-		// This shouldn't happen but it does for some internal PineJS requests
-		if (req.user && !req.creds) {
-			req.creds = req.user;
-		}
+	await retrieveAPIKey(req);
```
dfunckt: Same here.

src/platform/permissions.ts
```diff
@@ -206,168 +190,162 @@ export function createAll(
...
-										{
-											is_of__role: {
-												$any: {
-													$alias: 'rhp',
-													$expr: { rhp: { id: { $ne: null } } },
-												},
-											},
+	).tap(async roles => {
+		// Assign user roles
+		await Bluebird.all(
```
dfunckt:
```suggestion
		await Bluebird.map(_.toPairs(userMap), async ([userEmails, roleName]) => {
```

src/lib/device-types/index.ts
Page-: This was just making the minimal changes here to accomodate the `src/lib/device-types/storage/s3.ts` conversion, I'll eventually convert this file as well I'm sure

src/platform/permissions.ts
Page-: The current form will be quicker because it avoids creating unnecessarily an unnecessary intermediate

PR #237, opened 2019-12-18
Convert multicontainer check from a hook to a rule

Page-:
Change-type: patch


src/balena.sbvr
```diff
@@ -638,3 +638,5 @@ Rule: It is necessary that each release1 that has a status that is equal to "suc
 Rule: It is necessary that each device that should be running a release, should be running a release that belongs to an application that the device belongs to.
 Rule: It is necessary that each application that should be running a release, should be running a release that has a status that is equal to "success" and belongs to the application.
 --Rule: It is necessary that each application that should be running a release1, owns a release that has a status that is equal to "success".
+
+Rule: It is necessary that each release that contains at least 2 images, belongs to an application that has an application type that supports multicontainer.
```
thgreasi: Oh my, this rule sounds like poetry :sparkles:
I can't stand seeing the old hook code after seeing this.

dfunckt:
Looks great indeed, just need to confirm the rule won't apply the moment this is deployed, let's do a check before merging @Page- .

thgreasi:
Right Aki!
I would expect this to be a major in balena cloud, and to include an empty migration that would throw if there are still offenders (helps debugging such issues on our BoBs).

Page-:
It will run immediately and I'll make sure any violators are dealt with (the previous check should stop them but the rule is a better guarantee and far more straightforward)

PR #242, opened 2019-12-27
Async await

Page-:


src/lib/device-types/index.ts
```diff
@@ -291,61 +248,59 @@ function syncDataModel(
...
+		return null;
+	});
 
 	// if the cache is still empty, use this promise so that
 	// we do not start a second set of requests to s3
 	// in case another api request comes before the first completes
 	if (!deviceTypesCache) {
 		deviceTypesCache = promise;
-	} else {
-		updateDeviceTypesCache(promise);
```
Page-: I removed the `updateDeviceTypesCache` function altogether because all it does is mutate the contents of the previous cache with the updated info but as we don't keep the previously resolved cache contents around anywhere there's no need to mutate it and so we can simplify this and just rely on the
```ts
// when the promise gets resolved, cache it
deviceTypesCache = promise;
```
below to update the cache

PR #243, opened 2019-12-27
Export ExtraParams interface in auth

iamsolankiamit:
Change-type: patch
Signed-off-by: Amit Solanki <amit@balena.io>


iamsolankiamit:
Need it to help write types for wrapping functions for [loginUserXHR](https://github.com/balena-io/open-balena-api/blob/master/src/platform/auth.ts#L233).

PR #247, opened 2019-12-31
Fix migrations/00022-drop-vpn-host-port.sql

Page-:
Change-type: patch


src/migrations/00022-drop-vpn-host-port.sql
```diff
@@ -1,3 +1,3 @@
-ALTER TABLE "device"
+ALTER TABLE "application"
 DROP COLUMN IF EXISTS "VPN_port",
```
dfunckt: The fields are named "VPN port" and "VPN host".

PR #264, opened 2020-01-24
Populate the logoUrl in the device-type endpoint results

thgreasi:
This is a Hack Friday proof of concept, so it's the easiest way of achieving this.

Change-type: minor
HQ: https://github.com/balena-io/balena/issues/1818
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/DaHFH8vet7FsqzK6brh1ywBR-BD


src/lib/config.ts
```diff
@@ -46,6 +46,13 @@ export const requiredVar = (varName: string): string => {
 	return s;
 };
 
+export const optionalVar = <R>(
+	varName: string,
+	defaultValue: R,
+): string | R => {
```
dfunckt: There's no point making this generic -- it can always return `string`. If the value is supposed to be an integer, you'll have to parseInt anyway.

src/lib/device-types/build-info-facade.ts
```diff
@@ -21,6 +21,9 @@ export const getLogoUrl = memoizee(
 		buildId: string,
 	): Promise<string | undefined> => {
 		try {
+			if (!FILES_HOST) {
+				return;
+			}
```
dfunckt: This can be outside the try/catch block.

src/lib/config.ts
thgreasi: My point was just to avoid repeating `process.env.`, which made prettier to add line breaks and made the code less consistent and harder to quickly scan. Will revert.

src/lib/device-types/build-info-facade.ts
thgreasi: Sure

src/lib/config.ts
thgreasi: @dfunckt even though I liked the consistency of not having to access `process.env` directly anywhere in the config vars section, I reverted this as per your comment :disappointed:

dfunckt: @thgreasi I didn't mean to remove the function, read my comment again. What I'm saying is that the *function* need not be generic over some `R` -- it could just have the signature:

```
export const optionalVar = (varName: string, defaultValue: string): string
```

thgreasi: Oooooh, my bad :sweat_smile: 
@dfunckt isn't there a case that we indeed want an envvar to have `undefined` as a default value instead of an empty string?

dfunckt: There is, but what's the function you wrote can't be used for it -- what's the point of providing a default if undefined is permitted? You get zero benefit out of the function.

If you literally just want to avoid having to write process.env, just drop the `defaultValue` argument and type the return value as `string | undefined`.

thgreasi: Updated @dfunckt with func call override, so that it's `string | undefined` when no default value is provided, or `string` when one is.

src/lib/device-types/build-info-facade.ts
```diff
@@ -14,6 +15,34 @@ export const getIsIgnored = memoizee(
...
+		try {
+			const logoKey = getImageKey(normalizedSlug, buildId, 'logo.svg');
+			const hasLogo = await fileExists(logoKey);
+			if (!hasLogo) {
+				return;
+			}
+
+			const encodedLogoPath = getImageKey(
+				normalizedSlug,
+				encodeURIComponent(buildId),
```
dfunckt: Why do you URL-encode here but not above? Also, why only the `buildId` and not the slug?

src/lib/device-types/build-info-facade.ts
```diff
@@ -14,6 +15,34 @@ export const getIsIgnored = memoizee(
...
+				'logo.svg',
+			);
+			return `https://${FILES_HOST}/${encodedLogoPath}`;
```
dfunckt: Shouldn't this be a signed URL? What if the bucket or object is private?

src/lib/config.ts
```diff
@@ -46,6 +46,18 @@ export const requiredVar = (varName: string): string => {
 	return s;
 };
 
+export function optionalVar(varName: string, defaultValue: string): string;
+export function optionalVar(
+	varName: string,
+	defaultValue?: string,
+): string | undefined;
```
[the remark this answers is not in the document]
thgreasi: Unfortunately I need to repeat this (even though it's exactly the same as the implementation) since only the overload signature definitions are used while consuming the function.
The implementation's param types are not used anywhere else except the func implementation itself.

src/lib/device-types/build-info-facade.ts
thgreasi: > Why do you URL-encode here but not above?

AWS SDK allows us to pass plain string values and ignore url encoding, (probably that's done internally).
On the other hand since the logoUrl will be a url (:smiley:), I needed to do this since otherwise versions/`buildIds`s like `2.46.1+rev3.prod` would be parsed as `2.46.1 rev3.prod` on the files.balena-cloud.com side, which would return a 404.

> Also, why only the buildId and not the slug?

Atm only the `version` has the `+` issue, but indeed I should have used url encode on all of them :+1:

src/lib/device-types/build-info-facade.ts
thgreasi: This PR indeed doesn't handle private device types.
Need to re-discuss this with @afitzek but I would like to unblock this for the public ones at least, so that we don't have to manually add DT logos to the dashboard sooner than later.
See: https://github.com/balena-io/balenacloud-dashboard/issues/194 (since 2016)

src/lib/device-types/build-info-facade.ts
```diff
@@ -14,6 +15,33 @@ export const getIsIgnored = memoizee(
...
+		try {
+			const logoKey = getImageKey(normalizedSlug, buildId, 'logo.svg');
+			const hasLogo = await fileExists(logoKey);
+			if (!hasLogo) {
+				return;
+			}
+
+			// url encode since the buildId can contain a `+`
+			const encodedLogoPath = getImageKey(
+				...[normalizedSlug, buildId, 'logo.svg'].map(encodeURIComponent),
```
dfunckt: At the start of the function:

```
const pathComponents = [normalizedSlug, buildId, 'logo.svg']
```

and then:

```suggestion
				...pathComponents.map(encodeURIComponent),
```

And also above:

```
const logoKey = getImageKey(... pathComponents);
```

PR #266, opened 2020-01-30
Always use asynchronous `crypto.pseudoRandomBytes`

Page-:
Change-type: patch

### DIFF
--- a/src/hooks/resources/device.ts
+++ b/src/hooks/resources/device.ts
@@ -1,6 +1,5 @@
 import * as _ from 'lodash';
 import * as Bluebird from 'bluebird';
-import * as crypto from 'crypto';
 
 import {
 	checkDevicesCanHaveDeviceURL,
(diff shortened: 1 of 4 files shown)


dfunckt:
@Page- I would expect the change to be effectively a noop -- ie. even though the previous sync call is synchronous the event loop doesn't block and promisifying it won't offer any benefit. What do I miss?

Page-:
What do you mean the event loop doesn't block @dfunckt ?

dfunckt:
@Page- I mean, would the server block waiting on the sync `pseudoRandomBytes` call? I would expect it won't since the event loop (in libuv) will just switch to servicing other callbacks while it waits for the syscall to complete. Is that correct? If so, then switching to async won't make any difference effectively.

Page-:
No, it will block @dfunckt, that's the whole premise of a synchronous call; it blocks waiting for completion. An asynchronous call cedes the event loop for other work to continue whilst it does slow work (usually io) in the background. Blocking the event loop is why synchronous calls are so bad for performance

dfunckt:
@Page- I know what a synchronous call is. The point I'm trying to make, and perhaps the cause of my confusion, is that I'd expect Node to *always* perform I/O in a non-blocking fashion (which it does) and invoke syscalls in a threadpool in order to not block its event loop. That's the whole premise of an evented system. That is, even a call such as `fs.readSync` is only blocking from the perspective of JS, but the event loop keeps servicing I/O in the background etc.

PR #269, opened 2020-01-31
Linting fixes for resin-lint 3.x

Page-:
Change-type: patch


src/lib/device-online-state.ts
```diff
@@ -176,9 +176,9 @@ export class DeviceOnlineStateManager extends events.EventEmitter {
 
 	private readonly featureIsEnabled: boolean;
 
-	isConsuming: boolean = false;
-	rsmq: RedisSMQ;
-	redis: PromisifedRedisClient;
+	public isConsuming: boolean = false;
+	public rsmq: RedisSMQ;
+	public redis: PromisifedRedisClient;
```
dfunckt: I think these were meant to be `private` -- the access modifier was probably forgotten.

richbayliss: I see no reason they would need to be `public` 👍

PR #270, opened 2020-01-31
Throw specific http errors rather than generic errors

Page-:
Change-type: minor


Page-:
Yeah, it's why I marked it as a minor (ie major for 0.x versions), and it will change from a 404 to more specific errors

dfunckt:
I meant it's breaking change from an OData API version perspective.

PR #271, opened 2020-01-31
Grant device API keys read access to release and application tags

dfunckt:
Change-type: patch
Refs: https://github.com/balena-io/balena/issues/1843


thgreasi:
Feels like a `Change-type: minor` to me :P
Is `Refs:` new? I though we used `HQ:`

src/lib/auth.ts
```diff
@@ -56,22 +56,23 @@ export const DEVICE_API_KEY_PERMISSIONS = [
...
+	`resin.release_tag.read?release/any(r:r/should_be_running_on__device/any(d:d/${matchesActor}) or r/${belongsToApplicationOrIsManagedByDevice})`,
 	`resin.device_environment_variable.get?device/any(d:d/belongs_to__application/any(a:a/depends_on__application/any(da:da/${ownsDevice})))`,
 	`resin.device_environment_variable.all?device/any(d:d/${matchesActor})`,
 	`resin.application_environment_variable.get?${applicationControlsDevice}`,
 
 	// Dependent device permissions
-	`resin.device.create?belongs_to__application/any(da:da/depends_on__application/any(a:a/${ownsDevice}))`,
-	`resin.device.read?belongs_to__application/any(da:da/depends_on__application/any(a:a/${ownsDevice}))`,
-	`resin.device.update?belongs_to__application/any(da:da/depends_on__application/any(a:a/${ownsDevice}))`,
+	`resin.device.create?belongs_to__application/any(a:a/depends_on__application/any(da:da/${ownsDevice}))`,
```
Page-: I assume you've changed this because you think the second one is the dependent application? That's not the case though because you have `dependentApp depends on mainApp` so the aliases are now wrong/confusing

dfunckt: I didn't put any thought on this. The reasons I changed the aliases are:

1. other permissions a few lines above (eg. the permission for `device_environment_variable`) use the `da` alias for `depends_on__application `.
2. reusing the same alias both for `belongs_to__application` and `depends_on__application` felt confusing to me.

So the point of this change was to get consistency among the permissions. Happy to change all to `a:a/...` instead.

PR #273, opened 2020-02-03
Update resin-lint to 3.x

Page-:
Change-type: patch

### DIFF
--- a/src/lib/device-types/storage/s3.ts
+++ b/src/lib/device-types/storage/s3.ts
@@ -12,24 +12,24 @@ import {
 export const getKey = (...parts: string[]): string => parts.join('/');
 
 class UnauthenticatedS3Facade {
-	constructor(private s3Client: AWS.S3) {}
+	constructor(private s3: AWS.S3) {}
 
 	public headObject(
 		params: AWS.S3.Types.HeadObjectRequest,
 	): ReturnType<AWS.S3['headObject']> {
-		return this.s3Client.makeUnauthenticatedRequest('headObject', params);
+		return this.s3.makeUnauthenticatedRequest('headObject', params);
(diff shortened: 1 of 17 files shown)


src/lib/haiku-name.ts
```diff
@@ -137,5 +137,6 @@ const bigNumber = Math.pow(2, 12);
 
 export const generate = () => {
 	const rnd = Math.floor(Math.random() * bigNumber);
+	// tslint:disable-next-line:no-bitwise
```
Page-: This is the only case I had to just disable the rule and I don't want to touch the code here, but I do think the rule makes sense in general

src/lib/device-types/storage/s3.ts
```diff
@@ -12,24 +12,24 @@ import {
 export const getKey = (...parts: string[]): string => parts.join('/');
 
 class UnauthenticatedS3Facade {
-	constructor(private s3Client: AWS.S3) {}
+	constructor(private s3: AWS.S3) {}
```
dfunckt: Nit: I personally prefer `client` over `s3`, as it more accurately describes it.

Page-: I prefer s3 and went with it because `client` is pretty generic; it could refer to a pinejs client, a redis client, etc, whereas `s3` makes it clear what it talks to, eg `s3.makeUnauthenticatedRequest` is obvious where it's going whereas `client.makeUnauthenticatedRequest` is not. For this specific case the code is short enough that you can see all the context at once so it doesn't matter much either way, but if it was larger it might, particularly if we mixed multiple types of client.

That said I'd prefer `s3Client` but there was no easy refactoring to make the name available

PR #274, opened 2020-02-04
Device logs improvements

Page-:
This removes an unnecessary transaction, shrinks the potentially long living ctx data structure, and makes it easier to gc objects we only need at the start of the long lived requests


src/routes/device-logs.ts
```diff
@@ -357,7 +354,7 @@ async function getWriteContext(
 	req: Request,
 ): Promise<LogWriteContext> {
 	const { uuid } = req.params;
-	const [ctx] = (await resinApi.get({
+	const device = (await resinApi.get({
```
richbayliss: Why isn't this going to be `[device]`?

Page-: Good spot, it should be - fixed

PR #279, opened 2020-02-11
Update dependencies

Page-:
Change-type: patch


src/lib/certs.ts
```diff
@@ -6,8 +6,38 @@ import {
...
+	'PS256',
+	'PS384',
+	'PS512',
+	'none',
+] as const;
+const isAllowedAlgorithm = (
+	algo: any,
+): algo is typeof allowedAlgorithms[number] => allowedAlgorithms.includes(algo);
+
+if (!isAllowedAlgorithm(TOKEN_AUTH_JWT_ALGO)) {
```
Page-: This is to satisfy the updates jsonwebtoken typings

package.json
```diff
@@ -69,7 +69,7 @@
     "passport-jwt": "^4.0.0",
     "randomstring": "^1.1.5",
     "raven": "^2.6.4",
-    "redis": "^2.8.0",
+    "redis": "^3.0.2",
```
Page-: https://github.com/NodeRedis/node-redis/blob/v3.0.2/CHANGELOG.md#v300---09-feb-2020

We're not affected by the breaking changes and it mentions some performance benefits which would be nice

src/lib/certs.ts
```diff
@@ -6,8 +6,38 @@ import {
...
+	'RS256',
+	'RS384',
+	'RS512',
+	'ES256',
+	'ES384',
+	'ES512',
+	'PS256',
+	'PS384',
+	'PS512',
+	'none',
```
afitzek: We don't want to support the `none` algorithm, because this would mean one could provide a JWT which is not signed and we still accept it as valid. The other ones should be fine. The number is the size of the hash function 256-512 and the prefix defines the actual signature algorithm.

Page-: Removed 'none'

PR #288, opened 2020-02-26
Minor test suite improvements

dfunckt:
Change-type: patch

### DIFF
--- a/test/scenarios/unpin-device-after-release.ts
+++ b/test/scenarios/unpin-device-after-release.ts
@@ -3,10 +3,10 @@ import 'mocha';
 import * as _ from 'lodash';
 
 import { app } from '../../init';
-import { Admin, getAdminUser } from '../test-lib/api-helpers';
+import { getAdminUser, User } from '../test-lib/api-helpers';
 import { expect } from '../test-lib/chai';
 import * as fakeDevice from '../test-lib/fake-device';
-import { default as supertest, SupertestUser } from '../test-lib/supertest';
+import supertest, { SupertestUser } from '../test-lib/supertest';
 
 interface MockReleaseParams {
 	belongs_to__application: number;
@@ -92,7 +92,7 @@ const addImageToRelease = async (
 };
 
 describe('Device with missing service installs', () => {
-	let admin: Admin;
+	let admin: User;
 	let applicationId: number = 0;
 	let device: fakeDevice.Device;
 	const releases: _.Dictionary<number> = {};
--- a/test/scenarios/unpin-device-after-release.ts
+++ b/test/scenarios/unpin-device-after-release.ts
@@ -3,11 +3,10 @@ import 'mocha';
 import * as _ from 'lodash';
 
 import { app } from '../../init';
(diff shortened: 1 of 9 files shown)


test/test-lib/api-helpers.ts
```diff
@@ -1,9 +1,10 @@
 import { expect } from './chai';
-import supertest = require('./supertest');
+import supertest from './supertest';
 
 import { app } from '../../init';
 import { SUPERUSER_EMAIL, SUPERUSER_PASSWORD } from '../../src/lib/config';
 import { User } from '../../src/platform/jwt';
+export { User as Admin } from '../../src/platform/jwt';
```
thgreasi: What's the benefit of exporting this as `Admin` instead of just re-exporting the `User`?

NIT: How about
```
import { User as Admin } from '../../src/platform/jwt';
```
so the return type is consistent w/ the `let admin: Admin;` that we have?

test/scenarios/unpin-device-after-release.ts
```diff
@@ -3,11 +3,10 @@ import 'mocha';
 import * as _ from 'lodash';
 
 import { app } from '../../init';
+import { Admin, getAdminUser } from '../test-lib/api-helpers';
 import { expect } from '../test-lib/chai';
 import * as fakeDevice from '../test-lib/fake-device';
-import supertest = require('../test-lib/supertest');
-
-import { SUPERUSER_EMAIL, SUPERUSER_PASSWORD } from '../../src/lib/config';
+import { default as supertest, SupertestUser } from '../test-lib/supertest';
```
thgreasi:
```suggestion
import supertest, { SupertestUser } from '../test-lib/supertest';
```

test/test-lib/fake-device.ts
```diff
@@ -5,7 +5,7 @@ import * as uuid from 'uuid';
 
 import { app } from '../../init';
 
-import supertest = require('./supertest');
+import { default as supertest, SupertestUser } from './supertest';
```
thgreasi:
```suggestion
import supertest, { SupertestUser } from './supertest';
```

test/test-lib/api-helpers.ts
dfunckt: Changed it to `User`, I'm not convinced anymore `Admin` provides any value -- admins are users after all.

dfunckt: Changed it to `User`, I'm not convinced anymore `Admin` provides any value -- admins are users after all.

test/test-lib/supertest.ts
```diff
@@ -3,9 +3,9 @@ import * as _ from 'lodash';
 import * as supertest from 'supertest';
 import { User } from '../../src/platform/jwt';
 
-type UserObjectParam = Partial<User & { token: string }>;
+export type SupertestUser = string | Partial<User & { token: string }>;
```
Page-: I'd rather break it down to `ApiKey | Token` so test code can specify just one if it wants, with token being `{ token: string } & AnyObject`. It's also more accurate to what the code actually does

dfunckt:
Superseded by #298

PR #290, opened 2020-02-28
Don't allow downgrading a supervisor release for a device

CameronDiver:
Change-type: patch

### DIFF
--- a/src/hooks/resources/device.ts
+++ b/src/hooks/resources/device.ts
@@ -1,3 +1,4 @@
+import * as balenaSemver from 'balena-semver';
 import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
 
@@ -326,6 +327,28 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
 			);
 		}
 
+		if (request.values.should_be_managed_by__supervisor_release) {
+			if (
+				!Number.isInteger(
+					request.values.should_be_managed_by__supervisor_release,
+				)
+			) {
+				throw new BadRequestError('Expected an ID for the supervisor_release');
+			}
+
+			// Ensure that we don't ever downgrade the supervisor
+			// from its current version
+			waitPromises.push(
+				getCurrentRequestAffectedIds(args).then(ids =>
+					checkSupervisorReleaseUpgrades(
+						args.api,
+						ids,
+						request.values.should_be_managed_by__supervisor_release,
+					),
+				),
+			);
+		}
+
 		return Bluebird.all(waitPromises);
 	},
 	POSTRUN: args => {
@@ -494,3 +517,55 @@ addDeleteHookForDependents('device', [
 	['service_install', 'device'],
 	['gateway_download', 'is_downloaded_by__device'],
 ]);
+
+async function checkSupervisorReleaseUpgrades(
+	api: PinejsClient,
+	deviceIds: number[],
+	newSupervisorReleaseId: number,
+) {
+	const newSupervisorRelease = (await api.get({
+		resource: 'supervisor_release',
+		id: newSupervisorReleaseId,
+		options: {
+			$select: 'supervisor_version',
+		},
+	})) as AnyObject;
+
+	if (newSupervisorRelease == null) {
+		throw new BadRequestError(
+			`Could not find a supervisor release with this ID ${newSupervisorReleaseId}`,
+		);
+	}
+
+	const newSupervisorVersion = newSupervisorRelease.supervisor_version;
+
+	const releases = (await api.get({
+		resource: 'supervisor_release',
+		options: {
+			$select: ['supervisor_version'],
+			$filter: {
+				should_manage__device: {
+					$any: {
+						$alias: 'd',
+						$expr: {
+							d: {
+								id: {
+									$in: deviceIds,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	})) as AnyObject[];
+
+	for (const release of releases) {
+		const oldVersion = release.supervisor_version;
+		if (balenaSemver.lt(newSupervisorVersion, oldVersion)) {
+			throw new BadRequestError(
+				`Attempt to downgrade supervisor, which is not allowed`,
+			);
+		}
+	}
+}
--- a/test/05-supervisor-release.ts
+++ b/test/05-supervisor-release.ts
@@ -4,6 +4,8 @@ import { supertest, UserObjectParam } from './test-lib/supertest';
 
 import { app } from '../init';
 
+import * as fakeDevice from './test-lib/fake-device';
+
 describe('supervisor release', function() {
 	let admin: UserObjectParam;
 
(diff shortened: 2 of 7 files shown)


dfunckt:
@CameronDiver can you please convert the check to async/await please?

src/hooks/resources/device.ts
```diff
@@ -494,3 +509,64 @@ addDeleteHookForDependents('device', [
...
+) {
+	return api
+		.get({
+			resource: 'supervisor_release',
+			id: newSupervisorReleaseId,
+			options: {
+				$select: 'supervisor_version',
+			},
+		})
+		.then((newSupervisorRelease: AnyObject) => {
```
dfunckt: Async/await.

src/hooks/resources/device.ts
```diff
@@ -494,3 +509,64 @@ addDeleteHookForDependents('device', [
...
+			if (newSupervisorRelease == null) {
+				throw new BadRequestError(
+					`Could not find a supervisor release with this ID ${newSupervisorReleaseId}`,
+				);
+			}
+			const newSupervisorVersion = newSupervisorRelease.supervisor_version;
+			return Bluebird.map(deviceIds, deviceId => {
+				return api
```
dfunckt: That can be a single request fetching all affected devices and you can then map over the results.

src/hooks/resources/device.ts
```diff
@@ -494,3 +509,64 @@ addDeleteHookForDependents('device', [
...
+			$expand: {
+				should_be_managed_by__supervisor_release: {
+					$select: ['supervisor_version'],
+				},
+			},
+		},
+	})) as AnyObject[];
+
+	for (const { should_be_managed_by__supervisor_release } of devices) {
+		if (!_.isArray(should_be_managed_by__supervisor_release)) {
```
Page-:
```suggestion
		if (!Array.isArray(should_be_managed_by__supervisor_release)) {
```

src/hooks/resources/device.ts
```diff
@@ -494,3 +509,64 @@ addDeleteHookForDependents('device', [
...
+			// Something has gone wrong here, we should
+			// have received an empty array
+
+			// TODO: Is there a specific error I should use here
```
Page-: You should use one of the http errors, probably `InternalRequestError`

src/hooks/resources/device.ts
```diff
@@ -494,3 +509,64 @@ addDeleteHookForDependents('device', [
...
+			throw new Error(`Expected an array for supervisor releases from device`);
+		}
+		// So we know it's an array, if it's empty
+		// we're happy to return and allow the upgrade
+		// otherwise, we compare versions
+		if (_.isEmpty(should_be_managed_by__supervisor_release)) {
```
Page-:
```suggestion
		if (should_be_managed_by__supervisor_release === 0) {
```

src/hooks/resources/device.ts
```diff
@@ -494,3 +509,64 @@ addDeleteHookForDependents('device', [
...
+					$in: deviceIds,
+				},
+			},
+			$expand: {
+				should_be_managed_by__supervisor_release: {
+					$select: ['supervisor_version'],
+				},
+			},
+		},
+	})) as AnyObject[];
```
Page-: It would be better to invert this as
```suggestion
	const devices = (await api.get({
		resource: 'supervisor_release',
		options: {
			$select: ['supervisor_version'],
			$filter: {
				should_manage__device: {
					$any: {
						$alias: 'd',
						$expr: {
							d: {
								id: {
									$in: deviceIds,
								},
							}.
						},
					},
				},
			},
		},
	})) as AnyObject[];
```
and avoid fetching a load of wasteful data (it'll also simplify a bunch of the checks below)

src/hooks/resources/device.ts
```diff
@@ -1,5 +1,6 @@
 import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
+import * as resinSemver from 'resin-semver';
```
Page-: I have a PR in https://github.com/balena-io/open-balena-api/pull/291 that'll update this to balena-semver btw

test/05-supervisor-release.ts
```diff
@@ -52,4 +52,110 @@ describe('supervisor release', function() {
...
+		let small: number;
+		let middle: number;
+		let large: number;
+		before(async () => {
+			// create a new test application...
+			const { body: application } = await supertest(app, admin)
+				.post('/resin/application')
+				.send({
+					device_type: 'intel-nuc',
+					app_name: 'supervisor_release_tes_app',
```
Page-:
```suggestion
					app_name: 'supervisor_release_test_app',
```

test/05-supervisor-release.ts
```diff
@@ -52,4 +52,110 @@ describe('supervisor release', function() {
...
+				.get(`/resin/device_type?$select=id&$filter=slug eq 'intel-nuc'`)
+				.expect(200);
+
+			expect(res.body)
+				.to.have.nested.property('d[0].id')
+				.that.is.a('number');
+
+			middle = (
+				await supertest(app, admin)
+					.post(`/resin/supervisor_release`)
```
Page-: These look like they should be in the fixtures

CameronDiver: Yup and I looked at adding them but it didn't look like the infrastructure is there yet?

test/05-supervisor-release.ts
```diff
@@ -52,4 +52,110 @@ describe('supervisor release', function() {
...
+		it('should not allow downgrading a supervisor version', async () => {
+			await supertest(app, admin)
+				.patch(`/resin/device(${device.id})`)
+				.send({
+					should_be_managed_by__supervisor_release: middle,
+				})
+				.expect(400);
+		});
+
+		it('should correctly determine logstream values', async () => {
```
richbayliss: This test is just doing a downgrade check when the target contains `_logstream` -- it would be good if we could add a test to make sure that `vX.Y.Z` can be upgraded to `vX.Y.Z_logstream`

test/05-supervisor-release.ts
```diff
@@ -51,4 +53,74 @@ describe('supervisor release', function() {
...
+		let fx: fixtures.Fixtures;
+		before(async () => {
+			fx = await fixtures.load('05-supervisor-release');
+			admin = fx.users.admin;
+			applicationId = fx.applications.app1.id;
+
+			device = await fakeDevice.provisionDevice(admin, applicationId);
+		});
+
+		it('should allow setting a device to a supervisor release', async () => {
```
Page-:
```suggestion
		});

		after(async () => {
			await fixtures.clean(fx);
		});

		it('should allow setting a device to a supervisor release', async () => {
```

test/05-supervisor-release.ts
```diff
@@ -51,4 +53,74 @@ describe('supervisor release', function() {
 			.delete(`/resin/supervisor_release(${supervisorReleaseId})`)
 			.expect(200);
 	});
+
+	describe('Devices running supervisor releases', () => {
```
Page-: Is there any reason for this to be nested? I'd rather it be a separate top-level describe

test/05-supervisor-release.ts
```diff
@@ -51,4 +53,74 @@ describe('supervisor release', function() {
...
+
+			device = await fakeDevice.provisionDevice(admin, applicationId);
+		});
+
+		it('should allow setting a device to a supervisor release', async () => {
+			await supertest(app, admin)
+				.patch(`/resin/device(${device.id})`)
+				.send({
+					should_be_managed_by__supervisor_release:
+						fx['supervisor-release']['5.0.1'].id,
```
Page-: Could you do something like
```ts
let supervisorReleases: Dictionary<{ id: number }>
...
supervisorReleases = fx['supervisor-release']
```
please so we get some level of type enforcement (I would like to add types to the fixtures directly but that's far from straightforward, especially for `defaultFixtures` which are dynamically added)

test/05-supervisor-release.ts
CameronDiver: Only that it's directly related to supervisor releases, which is what the above describe is named

Page-: Could you un-nest it please and rename the other describe to `supervisor release management` as that's more accurate to what it is

xginn8:
ref https://github.com/balena-io/balena/issues/2108

PR #296, opened 2020-03-10
Tests: guarantee the mocks have initialized before continuing

Page-:
Change-type: patch

PR #298, opened 2020-03-10
Tests: add initial fixtures mechanism and use it for the admin user

Page-:
Depends on #296


test/test-lib/init_mocks.ts
```diff
@@ -1,7 +1,51 @@
...
 
-export {};
+	// override the interval used to emit the queue stats event...
+	const { DeviceOnlineStateManager } = await import(
+		'../../src/lib/device-online-state'
+	);
+	(DeviceOnlineStateManager as any)['QUEUE_STATS_INTERVAL_MSEC'] = 1000;
+};
+
+const getAdminUser = async () => {
```
Page-: I moved this to be private in here as it should always be fetched via the fixtures - there's no reason to call it directly anymore

test/test-lib/init_mocks.ts
```diff
@@ -1,7 +1,51 @@
...
+	// any user we try to create will be the superuser...
+	const { SUPERUSER_EMAIL, SUPERUSER_PASSWORD } = await import(
+		'../../src/lib/config'
+	);
+
+	if (!SUPERUSER_EMAIL || !SUPERUSER_PASSWORD) {
+		return;
```
dfunckt: Can you please change this so it crashes if these values aren't set? It's going to surprise people if the admin user doesn't exist when it's referenced in tests.

test/03-device-state-v2.ts
```diff
@@ -64,22 +63,13 @@ const waitFor = async (fn: () => boolean, timeout: number = 10000) => {
 };
 
 describe('Device State v2', () => {
-	let admin: string;
+	let admin: UserObjectParam;
```
dfunckt: BTW, I really like this pattern here @Page-, where we declare the test case "state" as *typed* local variables instead of stuffing everything untyped in `this` or `ctx`. As much as I don't (personally) like keeping state between tests, I don't see us moving away from doing so (at least in the short term), so I think we should adopt this pattern in other repos as well. This at least gives some visibility over what exactly is local state, and also allows refactoring because of the static typing, which is a big improvement over what we currently have.

init.ts
```diff
@@ -99,13 +99,17 @@ async function createSuperuser() {
 export const app = express();
 app.enable('trust proxy');
 
-const doRunTests = (process.env.RUN_TESTS || '').trim() === '1';
 const init = async () => {
 	try {
+		const doRunTests =
+			(process.env.RUN_TESTS || '').trim() === '1'
+				? await import('./test/test-lib/init_mocks')
```
dfunckt: I think it's worth renaming that file as it isn't really just initializing mocks anymore (and it'll only diverge more and more over time).

```suggestion
				? await import('./test/test-lib/init_tests')
```

test/03-device-state-v2.ts
Page-: The `this` pattern started in coffeescript where it makes a lot more sense btw and it's still in the typescript version of most converted tests because the focus is getting to typescript ASAP to avoid writing new code in coffeescript and getting benefits for most of the code.  Long-term it should all be standardized with proper typing but it'll be a much more gradual process

dfunckt: Yeah, I know, I didn't mean to suggest we're not making progress. This was more of a heads up to you to keep an eye out in review in order to suggest this as an improvement if needed.

test/test-lib/init-tests.ts
```diff
@@ -0,0 +1,54 @@
...
+import { supertest, UserObjectParam } from './supertest';
+
+export const preInit = async () => {
+	await import('./aws-mock');
+
+	// override the interval used to emit the queue stats event...
+	const { DeviceOnlineStateManager } = await import(
+		'../../src/lib/device-online-state'
+	);
+	(DeviceOnlineStateManager as any)['QUEUE_STATS_INTERVAL_MSEC'] = 1000;
```
dfunckt: Ugh, this is a bit ugly -- that class should get that config value from its environment and tests should override it appropriately instead of having to patch it like this in tests (though I'm not suggesting it should be changed with this PR).

test/test-lib/supertest.ts
```diff
@@ -1,23 +1,26 @@
 import * as express from 'express';
 import * as _ from 'lodash';
-import * as supertest from 'supertest';
+import * as $supertest from 'supertest';
```
CameronDiver: Is this a standard pattern? (adding a `$` in same named modules)

Page-: Honestly it's not something we do often enough to really have a pattern for, that said we do use `$` for a bunch of other places where we want a private/public version with the same name, eg for memoized functions

PR #299, opened 2020-03-11
Tests: use fixtures mechanism for applications

Page-:
Change-type: patch

### DIFF
--- a/test/test-lib/fixtures.ts
+++ b/test/test-lib/fixtures.ts
@@ -1,9 +1,135 @@
+import { sbvrUtils } from '@resin/pinejs';
 import * as Bluebird from 'bluebird';
+import * as fs from 'fs';
 import * as _ from 'lodash';
+import * as path from 'path';
 
-type Fixtures = Dictionary<PromiseLike<Dictionary<PromiseLike<any>>>>;
+import { Tx } from '@resin/pinejs/out/database-layer/db';
+import { Headers } from 'request';
+import { API_HOST } from '../../src/lib/config';
+import { requestAsync } from '../../src/lib/request';
 
-const defaultFixtures: Fixtures = {};
+const { api, root } = sbvrUtils;
+
+type PendingFixtures = Dictionary<PromiseLike<Dictionary<PromiseLike<any>>>>;
+type PartiallyAppliedFixtures = Dictionary<Dictionary<PromiseLike<any>>>;
+export type Fixtures = Dictionary<Dictionary<any>>;
+
+type LoaderFunc = (
+	jsonData: AnyObject,
+	fixtures: PartiallyAppliedFixtures,
+	tx: Tx,
+) => PromiseLike<any>;
+
+const logErrorAndThrow = (message: string, ...args: any[]) => {
+	console.error(message, ...args);
+	throw new Error(message);
+};
+
+const createResource = async (args: {
+	resource: string;
+	method?: string;
+	body?: AnyObject;
+	user?: { token: string };
+}) => {
+	const { resource, method = 'POST', body = {}, user } = args;
+	const headers: Headers = { 'X-Forwarded-Proto': 'https' };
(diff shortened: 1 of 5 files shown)


test/test-lib/fixtures.ts
```diff
@@ -1,7 +1,148 @@
...
+			`Failed to create: ${requestOpts.resource}`,
+			response.statusCode,
+			response.body,
+		);
+	}
+
+	return body;
+};
+
+const loaders: Dictionary<LoaderFunc> = {
```
Page-: I'd like to get rid of this `Dictionary<LoaderFunc>` and instead have the specific keys listed but I haven't worked out a nice way yet

test/test-lib/fixtures.ts
```diff
@@ -1,7 +1,148 @@
...
+	url: string;
+	resource: string;
+	headers: Headers;
+	method: string;
+	json: boolean;
+	body: AnyObject;
+}
 
 type Fixtures = Dictionary<PromiseLike<Dictionary<PromiseLike<any>>>>;
+type SemiAwaitedFixtures = Dictionary<Dictionary<PromiseLike<any>>>;
```
dfunckt:
```suggestion
type PendingFixtures = Dictionary<Dictionary<PromiseLike<any>>>;
```

?

test/test-lib/fixtures.ts
```diff
@@ -14,7 +155,34 @@ export const setDefaultFixtures = (
 
 export type FixtureData = Dictionary<Dictionary<any>>;
```
dfunckt: Nit, but might worth renaming this to `Fixtures`, since it's exported and is what tests will use, and the `Fixtures` registry above to `FixtureData` (or `FixturesRegistry`, or `FixturesByModel`, etc.).

test/test-lib/fixtures.ts
```diff
@@ -14,7 +155,34 @@ export const setDefaultFixtures = (
...
 
-export const load = async (): Promise<FixtureData> => {
-	const fixtures = { ...defaultFixtures };
-	return Bluebird.props(_.mapValues(fixtures, fx => Bluebird.props(fx)));
+export const load = async (fixtureName?: string): Promise<FixtureData> => {
```
dfunckt: I'd much rather:

```suggestion
export const load = async (fixtureName: string): Promise<FixtureData> => {
```

and then exporting a separate function:

```js
export const loadDefault = async (): Promise<FixtureData> => {
```

These two internally could either use this one (with the nullable argument), or `load(...)` could use `loadDefault()`.

test/test-lib/fixtures.ts
Page-: I renamed this one to `PartiallyAppliedFixtures` and the one above to `PendingFixtures` (since they're slightly more accurate names)

test/test-lib/fixtures.ts
Page-: I'd rather not do this as having a separate `loadDefault` implies that `load` wouldn't those defaults

test/test-lib/fixtures.ts
dfunckt: Maybe try something like this? https://github.com/balena-io/open-balena-api/blob/master/src/lib/device-state.ts#L63

Regardless, it's definitely worth putting this down as a `FIXME` comment, unless you plan to fix it in this PR.

test/test-lib/fixtures.ts
dfunckt: Makes sense. My concern is that for pretty much all uses of `load()`, the argument will be needed and the exported API merely suggests it -- does not enforce it. Maybe add a doc comment that not passing an argument will only load the default fixtures? VScode is pretty good making doc comments visible.

Page-: Done

PR #305, opened 2020-03-13
env-vars: Add DTB config vars for nvidia boards

thgreasi:
That's on top of #309 since it uses the new structure for the definitions.

Change-type: minor
See: https://github.com/balena-io/balenacloud-dashboard/issues/3413
HQ: https://github.com/balena-io/balena/issues/2166
Depends-on: https://github.com/balena-io/balenacloud-dashboard/issues/3486
Depends-on: https://github.com/balena-io/balena-supervisor/issues/1207
See: https://www.flowdock.com/app/rulemotion/r-supervisor/threads/54GM3Zqzk75loM4ELl77Zt1w7Vt
See: https://www.flowdock.com/app/rulemotion/resin-devices/threads/h4rhav-nPtFC6vROZyub47lVoRB

### DIFF
--- a/src/features/vars-schema/env-vars.ts
+++ b/src/features/vars-schema/env-vars.ts
@@ -182,6 +182,29 @@ export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
 		},
 	},
 	{
+		capableDeviceTypes: [
+			'astro-tx2',
+			'blackboard-tx2',
+			'jetson-tx2',
+			'n310-tx2',
+			'n510-tx2',
+			'orbitty-tx2',
+			'spacely-tx2',
+			'srd3-tx2',
+			'jetson-nano',
+			'jetson-nano-emmc',
+			'jn30b-nano',
+			'photon-nano',
+		],
+		properties: {
+			RESIN_HOST_EXTLINUX_fdt: {
+				type: 'string',
+				description:
+					'Define the file name of the DTB to be used. Only supported by supervisor versions >= v11.14.2.',
+			},
+		},
+	},
+	{
 		capableDeviceTypes: ['up-board'],
 		properties: {
 			RESIN_HOST_CONFIGFS_ssdt: {
--- a/test/01-basic.ts
+++ b/test/01-basic.ts
@@ -117,8 +117,15 @@ describe('Basic', () => {
 				],
 			},
 			{
+				deviceType: 'jetson-nano',
+				extraConfigVarSchemaProperties: ['RESIN_HOST_EXTLINUX_fdt'],
+			},
+			{
 				deviceType: 'jetson-tx2',
-				extraConfigVarSchemaProperties: ['RESIN_HOST_ODMDATA_configuration'],
+				extraConfigVarSchemaProperties: [
+					'RESIN_HOST_EXTLINUX_fdt',
+					'RESIN_HOST_ODMDATA_configuration',
+				],
 			},
 			{
 				deviceType: 'up-board',


src/lib/env-vars.ts
```diff
@@ -153,6 +153,26 @@ export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
...
+		capableDeviceTypes: [
+			'blackboard-tx2',
+			'jetson-nano',
+			'jetson-tx2',
+			'jn30b-nano',
+			'n510-tx2',
+			'orbitty-tx2',
+			'spacely-tx2',
+			'srd3-tx2',
+		],
```
20k-ultra, line 166:
```suggestion
		capableDeviceTypes: [
			'astro-tx2',
			'blackboard-tx2',
			'jetson-tx2',
			'n310-tx2',
			'n510-tx2',
			'orbitty-tx2',																	
			'spacely-tx2',
			'srd3-tx2',
			'jetson-nano',
			'jetson-nano-emmc',
			'jn30b-nano',
			'photon-nano',
		],
```

src/lib/env-vars.ts
```diff
@@ -153,6 +153,26 @@ export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
...
+		properties: {
+			RESIN_HOST_EXTLINUX_fdt: {
+				type: 'string',
+				description:
+					'Define the DTB to be used. Only supported by supervisor versions >= v11.11.0.',
+				examples: ['/path/to/file'],
```
20k-ultra, line 172:
```suggestion
				examples: ['default'],
```

src/lib/env-vars.ts
20k-ultra: This is an updated list from @acostach.

src/lib/env-vars.ts
thgreasi: @20k-ultra Isn't this supposed to be a file path or a filename? If it's supposed to be a path as we said, then imo this doesn't communicate that good enough.

20k-ultra: I totally understand. @acostach brings up some good points on this which convinced me it was a good idea: 


_I believe we could "tegra186-quill-p3310-1000-c03-00-base.dtb" or simply "default", because on the jetson nano there is another name. Using "default" also covers the case where nvidia decides to rename the dtb name in a future release. Using default also shows that the dtb from the raw emmc  partition is used._

Perhaps we can use `Default` and communicate it's a file path by changing the description. What about something like:

```Define a file path for the DTB to be used. Only supported by supervisor versions >= v11.11.0.'```

File: src/features/vars-schema/env-vars.ts [hunk 3]
```diff
@@ -153,6 +153,29 @@ export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
...
+			'jetson-nano',
+			'jetson-nano-emmc',
+			'jn30b-nano',
+			'photon-nano',
+		],
+		properties: {
+			RESIN_HOST_EXTLINUX_fdt: {
+				type: 'string',
+				description:
+					'Define the file name of the DTB to be used. Only supported by supervisor versions >= v11.11.0.',
```
20k-ultra, line 175:
```suggestion
					'Define the file name of the DTB to be used. Only supported by supervisor versions >= v11.14.2',
```

This had to be bumped because https://github.com/balena-io/balena-supervisor/pull/1423 didn't get merged until now.

20k-ultra (changes requested):
LGTM. This feature is technically in the Supervisor before the PR I linked above but it is fully functional at the version I suggested so let's advertise that being the version needed to use this feature.

20k-ultra (approved):
Good additions to the testing as well!

PR #307, opened 2020-03-13
Apply linting to typings files

Page-:
Change-type: patch


src/lib/device-config.ts
```diff
@@ -95,7 +95,7 @@ export const generateConfig = async (
 		{
 			appUpdatePollInterval:
 				_.parseInt(req.param('appUpdatePollInterval')) * 60 * 1000,
-			network: req.param('network'),
+			network: req.body.network ?? req.query.network,
```
Page-: The type error here was flagged up after the change from `Array<>` to `any[]` in typings, so I've converted this away from the deprecated `req.param` to fix it

PR #308, opened 2020-03-18
Add is_invalidated boolean term to the release resource

sradevski:
We currently need the flag to mark host OS releases as invalid
so they can be filtered out from the list of available versions.
The future plan is for the supervisor to automatically rollback
invalidated releases, but there are no dedicated resources to
work on that just yet.

This was already discussed at an arch call: https://app.frontapp.com/open/msg_ad91xc3

Change-type: minor

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -516,6 +516,7 @@ Fact type: release has source
 	Necessity: each release has exactly one source.
 Fact type: release has build log
 	Necessity: each release has at most one build log.
+Fact type: release is invalidated
 Fact type: release has start timestamp
 	Necessity: each release has exactly one start timestamp.
 Fact type: release has end timestamp
--- a/src/migrations/00025-release-is-invalidated.sql
+++ b/src/migrations/00025-release-is-invalidated.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "release" ADD COLUMN "is invalidated" INTEGER DEFAULT 0 NOT NULL;


src/balena.sbvr
```diff
@@ -516,6 +516,7 @@ Fact type: release has source
 	Necessity: each release has exactly one source.
 Fact type: release has build log
 	Necessity: each release has at most one build log.
+Fact type: release is invalidated
```
sradevski: We are open to other names if you have any better suggestions, this is just what we came up with at the arch call.

dfunckt: `is obsolete`?

sradevski: @dfunckt to me `obsolete` implies that it's out of date, but it's still fine to use it, which is not the case here, not sure if it's just me though :)

thgreasi:
Was this also supposed to filter out such releases from the ones that the device state endpoint takes considers?

sradevski:
@thgreasi for now there won't be any additional functionalities built on top of it, but there is a plan to do so, and expose it in the UI so users can easily mark a release as invalid.

nazrhom:
@balena-ci rebase

src/balena.sbvr
hedss: Yeah, 'obsolete' means 'out of date but still usable', 'invalidated' means 'rejected due to error'.

dfunckt: Okay gotcha -- I agree `is obsolete` isn't at all correct. I was more concerned with "is invalidated" meaning it's something it can be replaced with a new version (ie. "cache invalidation"), which in the case of a single release doesn't make sense to me, but if seen as a whole might :)

sradevski:
@balena-ci rebase

sradevski:
@dfunckt This is also what we came up with in the architecture call, and it is the most accurate description of all the options we considered.

sradevski:
@balena-ci rebase

PR #304, opened 2020-03-13
env-vars: Add SSDT overlays host configfs variable for the up-board

thgreasi:
That's on top of #309 since it uses the new structure for the definitions.

Change-type: minor
See: https://github.com/balena-io/balena-supervisor/issues/1113
See: https://github.com/balena-io/balena-supervisor/pull/1187
See: https://github.com/balena-os/meta-balena/issues/1845
See: https://github.com/balena-os/balena-up-board/issues/74
See: https://www.flowdock.com/app/rulemotion/r-supervisor/threads/tBuZeGU8N3aqK2IOidtHiOQIdBf
Signed-off-by: Thodoris Greasidis <thodoris@balena.io>

PR #309, opened 2020-03-18
Refactor how device type specific config vars are synthesized

thgreasi:
Gives us a more flexible way to define per device env var schemas, like in #304 #305 #306 

Change-type: patch

### DIFF
--- a/src/lib/env-vars.ts
+++ b/src/lib/env-vars.ts
@@ -109,42 +109,44 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
 	},
 };
 
-export const HOST_CONFIG_VAR_PROPERTIES: {
-	[k: string]: JSONSchema6Definition;
-} = {
-	RESIN_HOST_CONFIG_disable_splash: {
-		enum: ['0', '1'],
-		description: 'Enable / Disable the rainbow splash screen',
-		default: '1',
+export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
+	capableDeviceTypes: string[];
+	properties: Dictionary<JSONSchema6Definition>;
+}> = [
+	{
+		capableDeviceTypes: [
+			'raspberry-pi',
+			'raspberry-pi2',
+			'raspberrypi3-64',
+			'raspberrypi3',
+			'raspberrypi4-64',
+			'fincm3',
+			'revpi-core-3',
+			'npe-x500-m3',
+		],
+		properties: {
+			RESIN_HOST_CONFIG_disable_splash: {
+				enum: ['0', '1'],
+				description: 'Enable / Disable the rainbow splash screen',
+				default: '1',
+			},
+			RESIN_HOST_CONFIG_dtparam: {
+				type: 'string',
+				description: 'Define DT parameters',
+				default: '"i2c_arm=on","spi=on","audio=on"',
+			},
+			RESIN_HOST_CONFIG_enable_uart: {
+				enum: ['0', '1'],
+				description: 'Enable / Disable UART',
+				default: '1',
+			},
+			RESIN_HOST_CONFIG_gpu_mem: {
+				type: 'integer',
+				description: 'Define device GPU memory in megabytes.',
+				default: 16,
+			},
+		},
 	},
-	RESIN_HOST_CONFIG_dtparam: {
-		type: 'string',
-		description: 'Define DT parameters',
-		default: '"i2c_arm=on","spi=on","audio=on"',
-	},
-	RESIN_HOST_CONFIG_enable_uart: {
-		enum: ['0', '1'],
-		description: 'Enable / Disable UART',
-		default: '1',
-	},
-	RESIN_HOST_CONFIG_gpu_mem: {
-		type: 'integer',
-		description: 'Define device GPU memory in megabytes.',
-		default: 16,
-	},
-};
-
-// the namespace RESIN_HOST_CONFIG_ is only applicable for raspberrypis
-// as it supports the config.txt file
-export const RESIN_HOST_CONFIG_CAPABLE_DEVICE_TYPES = [
-	'raspberry-pi',
-	'raspberry-pi2',
-	'raspberrypi3-64',
-	'raspberrypi3',
-	'raspberrypi4-64',
-	'fincm3',
-	'revpi-core-3',
-	'npe-x500-m3',
 ];
 
 const startsWithAny = (ns: string[], name: string) => {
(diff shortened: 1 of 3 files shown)


src/routes/config.ts
```diff
@@ -21,13 +20,13 @@ export const vars: RequestHandler = (req, res) => {
...
-				: {}),
-		},
+		properties: Object.assign(
+			{},
+			SUPERVISOR_CONFIG_VAR_PROPERTIES,
+			...DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES.filter(config =>
+				config.capableDeviceTypes.has(req.query.deviceType),
+			).map(config => config.properties),
+		),
 	};
```
Page-:
```suggestion
		properties: {
			...SUPERVISOR_CONFIG_VAR_PROPERTIES,
			...DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES.filter(config =>
				config.capableDeviceTypes.has(req.query.deviceType),
			).map(config => config.properties),
		},
	};
```

src/lib/env-vars.ts
```diff
@@ -109,42 +109,44 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
...
-	RESIN_HOST_CONFIG_enable_uart: {
-		enum: ['0', '1'],
-		description: 'Enable / Disable UART',
-		default: '1',
+export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
+	capableDeviceTypes: Set<string>;
+	properties: Dictionary<JSONSchema6Definition>;
+}> = [
+	{
+		capableDeviceTypes: new Set([
```
Page-: Why a `Set`?

test/01-basic.ts
```diff
@@ -4,6 +4,110 @@ import { app } from '../init';
...
+	expect(configVarSchemaKeys).to.deep.equal(
+		[
+			'RESIN_SUPERVISOR_CONNECTIVITY_CHECK',
+			'RESIN_SUPERVISOR_INSTANT_UPDATE_TRIGGER',
+			'RESIN_SUPERVISOR_LOG_CONTROL',
+			'RESIN_SUPERVISOR_PERSISTENT_LOGGING',
+			'RESIN_SUPERVISOR_POLL_INTERVAL',
+			'RESIN_SUPERVISOR_VPN_CONTROL',
+		]
+			.concat(extraConfigVarSchemaProperties)
```
Page-:
```suggestion
		[
			'RESIN_SUPERVISOR_CONNECTIVITY_CHECK',
			'RESIN_SUPERVISOR_INSTANT_UPDATE_TRIGGER',
			'RESIN_SUPERVISOR_LOG_CONTROL',
			'RESIN_SUPERVISOR_PERSISTENT_LOGGING',
			'RESIN_SUPERVISOR_POLL_INTERVAL',
			'RESIN_SUPERVISOR_VPN_CONTROL',
			...extraConfigVarSchemaProperties,
		]
```

src/lib/env-vars.ts
thgreasi: Thought it would make searching a bit more performant (O(1) vs O(n) on arrays).
Can go with arrays if you prefer having primitives.

src/routes/config.ts
thgreasi: That doesn't work @Page- , since `SUPERVISOR_CONFIG_VAR_PROPERTIES` is an object, while `DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES.filter(...).map(...)` is an array of objects and that array get destructured as "rest arguments" of the `Object.assign`.
I don't think that we can do `{...x}` destructuring of unknown number of objects w/o a loop, which would be bad, since it would create tons of intermediate short lived objects.

src/lib/env-vars.ts
Page-: I'd say an array is fine, particularly at this size

Page-: Alternatively you could do:
```ts
_([...]).keyBy().mapValues(() => true).value()
```
or just directly do
```ts
{
	'raspberry-pi': true,
	...
}
```
and have an object lookup

thgreasi: Done :+1:

src/routes/config.ts
```diff
@@ -21,13 +20,13 @@ export const vars: RequestHandler = (req, res) => {
...
-			...(RESIN_HOST_CONFIG_CAPABLE_DEVICE_TYPES.includes(req.query.deviceType)
-				? HOST_CONFIG_VAR_PROPERTIES
-				: {}),
-		},
+		properties: Object.assign(
+			{},
+			SUPERVISOR_CONFIG_VAR_PROPERTIES,
+			...DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES.filter(config =>
+				config.capableDeviceTypes.includes(req.query.deviceType),
+			).map(config => config.properties),
```
Page-: This could be memoized if you're worried about performance

src/lib/env-vars.ts
thgreasi: Just to be clear, I went with the array.
Imo though a Set sounds more appropriate than `keyBy` or `{ 'raspberry-pi': true, ... }`, since the spec describes this as a native hashmap, it sounds weird to try replicate exactly that w/ a different way.

src/routes/config.ts
thgreasi: I guess that we don't need to, since this isn't a hot endpoint, so we can save the extra memory consumption. Let me know if you have a strong opinion on this though.

Page-: Memoizing would cause less allocations due to the filtering/mapping and so might actually be beneficial for memory - the baseline would be slightly higher but if there's multiple requests for the same device type concurrently then you'd only have one copy, and no need to worry about allocating for each request

src/routes/config.ts
```diff
@@ -1,33 +1,47 @@
...
+		return Object.assign(
+			{},
+			SUPERVISOR_CONFIG_VAR_PROPERTIES,
+			...DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES.filter(config =>
+				config.capableDeviceTypes.includes(deviceType),
+			).map(config => config.properties),
+		);
+	},
+	{
+		maxAge: 5 * 60 * 1000,
```
richbayliss: should this be configurable?

thgreasi: Reverted the memoization since that endpoint isn't hot at all (not even warm), so let's keep it simple.
See: https://www.flowdock.com/app/rulemotion/p-opensource/messages/48253

PR #311, opened 2020-03-23
Allow string ids for supervisor_release resources

CameronDiver:
Change-type: patch

### DIFF
--- a/src/hooks/resources/device.ts
+++ b/src/hooks/resources/device.ts
@@ -328,6 +328,15 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
 		}
 
 		if (request.values.should_be_managed_by__supervisor_release) {
+			// First try to coerce the value to an integer for
+			// moving forward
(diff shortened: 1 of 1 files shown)


src/hooks/resources/device.ts
```diff
@@ -330,7 +330,7 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
 		if (request.values.should_be_managed_by__supervisor_release) {
 			if (
 				!Number.isInteger(
-					request.values.should_be_managed_by__supervisor_release,
+					parseInt(request.values.should_be_managed_by__supervisor_release, 10),
```
Page-: Can you make this
```ts
if (request.values.should_be_managed_by__supervisor_release != null) {
	request.values.should_be_managed_by__supervisor_release = parseInt(request.values.should_be_managed_by__supervisor_release, 10);
	if (!Number.isInteger(request.values.should_be_managed_by__supervisor_release))
```
so it's also a number when we pass it forwards

PR #312, opened 2020-03-25
Cleanup imports and exports

dfunckt:
A few insignificant commits, except one that removes PineJS types that were re-exported by the API

Also, the first commit groups imports based on their origin:

- First are third-party libraries
- Then, first-party libraries
- Then, imports from “./platform”
- Finally, project imports

We don't enforce this (nor can we), but it'll hopefully encourage our future selves to preserve this separation.

Commits
=======

- Group imports by origin
- Remove unused lodash imports
- Remove underscored imports: Underscored imports where used to import only types and guard against using it for other reasons, but this can be enforced with type-only imports now where needed.
- Consistently import balena-semver as “semver”
- Stop re-exporting PineJS types: Back in the open sourcing days, it was thought that open-balena-api would completely wrap PineJS and expose higher level abstractions for clients, such as the cloud API. For this reason, it re-exported a few common PineJS types. This is not where we’re going however, so remove those exports, and have clients import them from PineJS.

### DIFF
--- a/src/abstract-sql-utils.ts
+++ b/src/abstract-sql-utils.ts
@@ -1,9 +1,9 @@
-import { Definition } from '@resin/odata-to-abstract-sql';
-
-import { AbstractSqlModel } from '@resin/abstract-sql-compiler';
-import { sbvrUtils } from '@resin/pinejs';
 import { readFileSync } from 'fs';
 import * as _ from 'lodash';
+
+import { AbstractSqlModel } from '@resin/abstract-sql-compiler';
+import { Definition } from '@resin/odata-to-abstract-sql';
+import { sbvrUtils } from '@resin/pinejs';
 
 export const generateAbstractSqlModel = (
 	seModelPath: string,
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,19 +1,15 @@
-import { Server } from 'http';
-import * as path from 'path';
-
-import * as _express from 'express';
-import * as _ from 'lodash';
-
 import * as Bluebird from 'bluebird';
-import * as Raven from 'raven';
-
 import * as bodyParser from 'body-parser';
 import * as compression from 'compression';
 import * as cookieParser from 'cookie-parser';
+import cookieSession = require('cookie-session');
+import * as _express from 'express';
+import { Server } from 'http';
+import * as _ from 'lodash';
 import * as methodOverride from 'method-override';
 import * as passport from 'passport';
-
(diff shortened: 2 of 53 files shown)


dfunckt:
You can completely ignore the first commit, which is the bulk of the diff, when reviewing -- it's a noop.

src/abstract-sql-utils.ts
```diff
@@ -1,9 +1,9 @@
-import { Definition } from '@resin/odata-to-abstract-sql';
+import { readFileSync } from 'fs';
+import * as _ from 'lodash';
 
 import { AbstractSqlModel } from '@resin/abstract-sql-compiler';
+import { Definition } from '@resin/odata-to-abstract-sql';
 import { sbvrUtils } from '@resin/pinejs';
-import { readFileSync } from 'fs';
-import * as _ from 'lodash';
```
Page-: You've reorganized this "by type" but you've grouped type-only imports with "real" imports which doesn't make sense imo (and those type-only imports should be `import type` but that wasn't possible until very recently)

src/index.ts
```diff
@@ -1,19 +1,15 @@
-import { Server } from 'http';
-import * as path from 'path';
-
-import * as _express from 'express';
-import * as _ from 'lodash';
```
Page-: These are also type-only imports that made sense to be separate (and I expect there'll probably be a bunch more so I'll stop commenting on them now)

src/index.ts
```diff
@@ -3,7 +3,7 @@ import * as bodyParser from 'body-parser';
 import * as compression from 'compression';
 import * as cookieParser from 'cookie-parser';
 import cookieSession = require('cookie-session');
-import * as _express from 'express';
+import * as express from 'express';
```
Page-: Typescript was always clever enough to elide unused types (or certainly for a very long time), this was also about making sure they weren't used for anything other than types unintionally)

src/abstract-sql-utils.ts
dfunckt: Yeah, I meant by "origin" (I mention "origin" everywhere except the commit title). I don't think separating type-only imports makes sense in this light.

src/index.ts
dfunckt: You're right, I'll fix the commit message and PR description.

src/abstract-sql-utils.ts
dfunckt: Marking type-only imports on #314

PR #313, opened 2020-03-25
Update dependencies

Page-:
All formatting courtesy of balena-lint

Update pinejs from 10.3.0 to 10.3.1

Change-type: patch


index.js
```diff
@@ -1,4 +1,3 @@
-// tslint:disable:no-console
```
Page-: Updated balena-lint inherits the ts rules for js which means we don't need to manually disable this for js anymore

typings/device-types.d.ts
```diff
@@ -12,12 +12,12 @@ declare module '@resin.io/device-types' {
 		instructions?: string[] | DeviceTypeInstructions;
 		gettingStartedLink?: string | DeviceTypeGettingStartedLink;
 		stateInstructions?: { [key: string]: string[] };
-		options?: Array<DeviceTypeOptions | DeviceTypeOptionsGroup>;
+		options?: (DeviceTypeOptions | DeviceTypeOptionsGroup)[];
```
Page-: This is unexpected @wrboyce

index.js
dfunckt: Seems I'm getting errors locally without this:

```
index.js:10:3
ERROR: 10:3  no-console  Calls to 'console.log' are not allowed.
ERROR: 12:4  no-console  Calls to 'console.log' are not allowed.
ERROR: 17:4  no-console  Calls to 'console.log' are not allowed.
```

PR #315, opened 2020-03-25
Use `canAccess` to simplify device api key permissions

Page-:
Change-type: patch

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -10,6 +10,12 @@ import {
 	VPN_SERVICE_API_KEY,
 } from './config';
 
+export const writePerms = (resource: string, filter: string): string[] => [
+	`${resource}.create?${filter}`,
+	`${resource}.update?${filter}`,
+	`${resource}.delete?${filter}`,
+];
+
 const matchesActor = 'actor eq @__ACTOR_ID';
 const matchesUser = `user/any(u:u/${matchesActor})`;
 const ownsDevice = `owns__device/any(d:d/${matchesActor})`;
@@ -33,7 +39,7 @@ export const ROLES: {
 		'resin.application_tag.all',
 		'resin.application_type.all',
 		'resin.device.all',
-		`resin.device.tunnel-22222`,
+		'resin.device.tunnel-22222',
 		'resin.device_config_variable.all',
 		'resin.device_environment_variable.all',
 		'resin.device_tag.all',
@@ -49,7 +55,7 @@ export const ROLES: {
 		'resin.service.all',
 		'resin.service_environment_variable.all',
 		'resin.service_install.all',
-		"resin.service_instance.get?service_type eq 'vpn'",
+		"resin.service_instance.read?service_type eq 'vpn'",
 		'resin.service_label.all',
 		'resin.supervisor_release.all',
 		'resin.user.read',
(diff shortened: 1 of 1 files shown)


src/lib/auth.ts
```diff
@@ -117,8 +140,8 @@ export const DEVICE_API_KEY_PERMISSIONS = [
 
 ROLES['device-api-key'] = [
 	...DEVICE_API_KEY_PERMISSIONS,
-	`resin.user.get`,
-	`resin.user__has__public_key.get`,
+	`resin.user.read`,
+	`resin.user__has__public_key.read`,
```
Page-: Not technically necessary, but standardizing on `.read` is nice, and made my grepping far easier ;)

src/lib/auth.ts
```diff
@@ -57,58 +63,75 @@ export const ROLES: {
 };
 
 export const DEVICE_API_KEY_PERMISSIONS = [
-	`resin.device_type.get?describes_device/any(d:d/${matchesActor})`,
-	`resin.device.get?${matchesActor}`,
+	`resin.device_type.read?describes_device/canAccess()`,
```
Page-: This technically expands the permissions to also be able to read device types of dependent devices, but I think that's actually correct

src/lib/auth.ts
```diff
@@ -57,58 +63,75 @@ export const ROLES: {
...
+	`resin.device.read?${matchesActor}`,
 	`resin.device.update?${matchesActor}`,
-	`resin.supervisor_release.get?should_manage__device/any(d:d/${matchesActor})`,
-	`resin.application.read?${ownsDevice} or depends_on__application/any(da:da/${ownsDevice})`,
-	`resin.application_tag.read?application/any(a:a/${ownsDevice} or a/depends_on__application/any(da:da/${ownsDevice}))`,
-	`resin.device_config_variable.get?device/any(d:d/${matchesActor} or d/belongs_to__application/any(a:a/depends_on__application/any(da:da/${ownsDevice})))`,
+	`resin.supervisor_release.read?should_manage__device/canAccess()`,
```
Page-: This technically expands the permissions to also be able to read supervisor releases of dependent devices, which is a bit redundant but I think it's the right thing to do

src/lib/auth.ts
```diff
@@ -57,58 +63,75 @@ export const ROLES: {
...
+	`resin.device_tag.read?device/canAccess()`,
 	`resin.device_tag.set?device/any(d:d/${matchesActor})`,
-	`resin.application_config_variable.get?${applicationControlsDevice}`,
-	`resin.release.read?should_be_running_on__device/any(d:d/${matchesActor}) or ${belongsToApplicationOrIsManagedByDevice}`,
-	`resin.release_tag.read?release/any(r:r/should_be_running_on__device/any(d:d/${matchesActor}) or r/${belongsToApplicationOrIsManagedByDevice})`,
-	`resin.device_environment_variable.get?device/any(d:d/belongs_to__application/any(a:a/depends_on__application/any(da:da/${ownsDevice})))`,
-	`resin.device_environment_variable.all?device/any(d:d/${matchesActor})`,
-	`resin.application_environment_variable.get?${applicationControlsDevice}`,
+	`resin.application_config_variable.read?application/canAccess()`,
+	`resin.release.read?should_be_running_on__device/canAccess() or ${belongsToApplicationOrIsManagedByDevice}`,
```
afitzek: `${belongsToApplicationOrIsManagedByDevice}` could become `belongs_to__application/canAccess()` I think?

Page-: Done, good spot - I was checking for reads that didn't already use `canAccess` which missed this one that did.. but only for one half

src/lib/auth.ts
```diff
@@ -58,58 +64,75 @@ export const ROLES: {
...
+	'resin.release_tag.read?release/canAccess()',
+	'resin.device_environment_variable.read?device/canAccess()',
+	...writePerms(
+		'resin.device_environment_variable',
+		`device/any(d:d/${matchesActor})`,
+	),
+	'resin.application_environment_variable.read?application/canAccess()',
 
 	// Dependent device permissions
+	'resin.device.read?belongs_to__application/any(a:a/depends_on__application/canAccess())',
```
Page-: Updated this as well, @afitzek just to double check it's fine to have a nested `canAccess` like this right?

afitzek: I think this should work as intended: You can read the device resource, if there exists a `belongs_to__application` that has a `depends_on__application` that is readable by the device.

PR #318, opened 2020-03-26
Add is_archived term to the application resource

sradevski:
The original purpose of the flag is to be able to
discontinue host applications, but the same concept
can be used by users to mark applications that are not
supported/developed anymore. This was discussed at a
product call and we decided to go with is_archived
based on Github's terminology as the meaning of both
are close.

Change-type: minor

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -388,6 +388,7 @@ Fact type: application1 depends on application2
 Fact type: application has application type
 	Necessity: each application has exactly one application type.
 Fact type: application is host
+Fact type: application is archived
 
 
 -- service instance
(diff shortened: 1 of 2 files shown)


dfunckt:
@sradevski shouldn't the API respect that flag? For example, should it not prevent new releases on an archived app? Otherwise, the flag is just random client metadata that I'd personally prefer was stored in a metadata-specific part of the model, for example app tags.

sradevski:
@dfunckt It should, but what we discussed was to just add the flag for the host apps use-case we have now and build functionality on top of it later when we want to put the resources in it. We didn't go with tags as this is something that we will want to be part of the model sooner or later.

PR #323, opened 2020-03-31
Avoid using res.send

Page-:
Change-type: patch

### DIFF
--- a/src/routes/devices.ts
+++ b/src/routes/devices.ts
@@ -381,7 +381,7 @@ const stateQuery = api.resin.prepare<{ uuid: string }>({
 export const state: RequestHandler = async (req, res) => {
 	const uuid = req.param('uuid');
 	if (!uuid) {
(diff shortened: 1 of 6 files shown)


src/routes/devices.ts
```diff
@@ -381,7 +381,7 @@ const stateQuery = api.resin.prepare<{ uuid: string }>({
 export const state: RequestHandler = async (req, res) => {
 	const uuid = req.param('uuid');
 	if (!uuid) {
-		return res.send(400);
+		return res.status(400).send();
```
dfunckt: It's not clear from the documentation what an empty `send()` does, so I'm not 100% sure on what is the expected result here, but if I understand this correctly I think `end()` conveys intent better.

Page-: It handles whatever extra stuff express does and then ends the response, I've switched to .end though

PR #324, opened 2020-04-07
Improve error handling

Page-:


src/platform/errors.ts
```diff
@@ -10,10 +11,19 @@ export const translateError = (err: Error | number | string): string => {
 	if (err instanceof InternalRequestError) {
 		return 'Server Error';
 	}
-	if (err instanceof Error) {
-		return err.message;
+	let message;
+	if (err instanceof HttpError) {
+		captureException(
+			new Error(),
+			'Translating a HttpError, this should go via handleHttpErrors instead',
```
dfunckt, line 18:
```suggestion
			'Translating an HttpError, this should go via handleHttpErrors instead',
```

src/routes/applications.ts
```diff
@@ -64,8 +64,7 @@ export const downloadImageConfig: RequestHandler = async (req, res) => {
 		res.json(config);
 	} catch (err) {
 		if (err instanceof UnauthorizedError) {
-			res.status(404).send(err.message);
-			return;
+			err = new BadRequestError(err);
```
dfunckt, line 67: Why change the response? Shouldn't this be NotFoundError instead?

src/platform/errors.ts
Page-: It depends on your accent @dfunckt, if you say `haitch` it's `a`, if you say `aitch` it's `an` - I'd use `a`

src/platform/errors.ts
```diff
@@ -10,10 +11,19 @@ export const translateError = (err: Error | number | string): string => {
...
+		);
+		message = err.message;
+	} else if (err instanceof Error) {
+		message = err.message;
+	} else {
+		message = `${err}`;
 	}
-	return `${err}`;
+	return escapeHtml(message);
```
thgreasi, line 26: I hope that this doesn't make the API generated errors look weird in the dashboard.

src/routes/applications.ts
Page-: oh ffs yes

Page-: fixed

src/platform/errors.ts
dfunckt: Okay, I think I'll trust you on this :)

src/platform/errors.ts
Page-: I would expect not, but if so we can handle those as we spot them :)

src/platform/errors.ts
Page-: Yeah it really comes down to accent, it's `an` before vowel sounds and `a` before non-vowel sounds, and the letter `H` can be said with either so the correct one depends on the reader.  You can also get vowels with non-vowel sounds like `a ukulele`

PR #330, opened 2020-04-10
track device logs bytes written (information lost when rotating at retention limit)

dt-rush:
This key is read by a prometheus metrics exporter to produce a histogram of the rate of logs use over all devices, see: https://github.com/balena-io/balena-redis-exporter/blob/first-build/device-logs-bytes-written-exporter/src/sampler.ts#L43-L65

Writing these per-device keys to redis and analyzing them in a batch job (see the code linked above) is necessary since merely exporting a bunch of time series labelled by device UUID would exceed the safe cardinality for a prometheus timeseries. The exporter above processes this high cardinality data into a small histogram (cardinality on the order of 46 histogram bins rather than 55k devices)

Change-type: minor
Signed-off-by: dt-rush <nickp@balena.io>

### DIFF
--- a/src/lib/device-logs/backends/redis.ts
+++ b/src/lib/device-logs/backends/redis.ts
@@ -77,6 +77,7 @@ export class RedisBackend implements DeviceLogsBackend {
 
 		const limit = ctx.retention_limit || 0;
 		const key = this.getKey(ctx);
+		const bytesWrittenKey = this.getKey(ctx, 'logBytesWritten');
 		// Create a Redis transaction
 		const tx = this.cmds.multi();
 		// Add the logs to the List structure
@@ -84,11 +85,16 @@ export class RedisBackend implements DeviceLogsBackend {
 		// Trim it to the retention limit
 		tx.ltrim(key, -limit, -1);
 		// Publish each log using Redis PubSub
+		let bytesWritten = 0;
 		for (const rLog of redisLogs) {
 			tx.publish(key, rLog);
+			bytesWritten += rLog.length;
 		}
+		// Increment log bytes written total
+		tx.incrby(bytesWrittenKey, bytesWritten);
 		// Devices with no new logs eventually expire
 		tx.pexpire(key, KEY_EXPIRATION);
+		tx.pexpire(bytesWrittenKey, KEY_EXPIRATION);
 		return Bluebird.fromCallback((callback) => {
 			tx.exec(callback);
 		});
@@ -134,8 +140,8 @@ export class RedisBackend implements DeviceLogsBackend {
 		return this.cmds.connected && this.pubSub.connected;
 	}
 
-	private getKey(ctx: LogContext) {
-		return `device:${ctx.id}:logs`;
+	private getKey(ctx: LogContext, suffix = 'logs') {
+		return `device:${ctx.id}:${suffix}`;
 	}
 
 	private handleMessage(key: string, payload: string) {


src/lib/device-logs/backends/redis.ts
```diff
@@ -76,19 +76,25 @@ export class RedisBackend implements DeviceLogsBackend {
 		const redisLogs = logs.map(this.toRedisLog, this);
 
 		const limit = ctx.retention_limit || 0;
-		const key = this.getKey(ctx);
+		const logKey = this.getKey(ctx);
```
dt-rush, line 79: renamed key for clarity when dealing in the same block of code with another redis key

src/lib/device-logs/backends/redis.ts
```diff
@@ -76,19 +76,25 @@ export class RedisBackend implements DeviceLogsBackend {
...
+		const logBytesWrittenKey = this.getLogBytesWrittenKey(ctx);
```
dt-rush, line 80: this is the new key we add

src/lib/device-logs/backends/redis.ts
```diff
@@ -76,19 +76,25 @@ export class RedisBackend implements DeviceLogsBackend {
...
-			tx.publish(key, rLog);
+			tx.publish(logKey, rLog);
+			logBytesWritten += rLog.length;
 		}
+		// Increment log bytes written total
+		tx.incrby(logBytesWrittenKey, logBytesWritten);
 		// Devices with no new logs eventually expire
-		tx.pexpire(key, KEY_EXPIRATION);
+		tx.pexpire(logKey, KEY_EXPIRATION);
+		tx.pexpire(logBytesWrittenKey, KEY_EXPIRATION);
```
dt-rush, line 97: both keys expire together (they represent the same device)

src/lib/device-logs/backends/redis.ts
```diff
@@ -76,19 +76,25 @@ export class RedisBackend implements DeviceLogsBackend {
...
+		tx.rpush(logKey, redisLogs);
 		// Trim it to the retention limit
-		tx.ltrim(key, -limit, -1);
+		tx.ltrim(logKey, -limit, -1);
 		// Publish each log using Redis PubSub
+		let logBytesWritten = 0;
 		for (const rLog of redisLogs) {
-			tx.publish(key, rLog);
+			tx.publish(logKey, rLog);
+			logBytesWritten += rLog.length;
```
Page-, line 91: This specifically is the size of the avsc serialized binary data which is fairly transformed vs the actual logs we receive - that may be fine and will still give an idea of the relative log usage between devices but it should be used as guidance rather than precise figures.

src/lib/device-logs/backends/redis.ts
Page-: Why do we expire the data about logs written? It's not an issue to do so imo but I'm wondering if there's a reason we want to

Page-: I also wonder if this will cause an issue where it looks like a device suddenly has a massive negative delta?

dt-rush: A1: If a device is no longer logging, we can safely expire the tally of bytes it has written as well (a counter associated with a process (logs writing) which no longer exists)

dt-rush: A2: deltas are computed based on this expression:

```
for (const device of Object.keys(nowSample.data)) {
	const deltaBytes = nowSample.data[device] - lastSample.data?.[device] ?? 0;
	const bytesPerMin = 60 * (deltaBytes / (msSince(lastSample.t) / 1e3));
	metrics.histogram('device_logs_bytes_written_per_min_hist', bytesPerMin);
}
```

Where the `nowSample` consists of data from `device:*:logBytesWritten` keys which are *currently* present (sampled "just now"). So if we sampled (and computed the delta) from a certain amount (say, `x`) of logBytesWritten last minute (eg. there is a `device:D3ADB33F:logBytesWritten: x` in `lastSample.data` map), but this minute there is no longer an entry (ie. an expiry time less than a minute, something we will never see in practice), it would still not result in a negative delta since `device:D3ADB33F:logBytesWritten` is not a key of `nowSample.data`.

([source link in context of `sampler.ts`](https://github.com/balena-io/balena-redis-exporter/blob/first-build/device-logs-bytes-written-exporter/src/sampler.ts#L76))

src/lib/device-logs/backends/redis.ts
dt-rush: @Page- is there an `O(1)` storage complexity difference between serialized and raw logs received?

dt-rush: And if so, if that constant factor is small enough, it may be alright to just look at serialized size. But I think we are also interested in the actual *network transfer bytes*, represented by the data seen at `POST /device_logs`, vs seen here which is more representative of *bytes written to disk*.

dt-rush: @petrosagg @Page- brings a good point here. 

Given we are interested in this data from a capacity planning perspective, we might want metrics on both the serialized and raw logs (the former represents storage space, the latter network transfer). 

For the former (serialized logs sent to redis), this PR suffices. For the latter, (raw logs bytes seen from devices at the soonest ingestion point, `POST /device_logs`) I can easily add `redis.incrby()` calls to the `/device_logs` endpoint as well. 

My question is thus, summarized: would we like both (given that it's trivial to implement both, and then scan both in our exporter and have two histograms). There may be some useful difference to tracking both. Given the work done so far, this would be like a ~10 line change to both the API and the exporter to add that ability.

dt-rush: In fact I'm working this out as a PR now.

dt-rush: In the branch I'm working out now (soon posted), we will track:

- `device:*:logBytesWritten` for serialized size in redis keystore
- `device:*:logBytesReceived` for bytes received at the `/device_logs` endpoint

dt-rush: In researching this, I've found even this PR doesn't handle streaming writes. My first order of business is to make sure this PR includes that, as we want at least an accurate measure of bytes written to REDIS.

dt-rush: This PR at any rate, to keep track of bytes written, is ready to go. I had a slight worry that it didn't handle streaming writes, but since the instrumentation is added in the `publish()` method, which is used by both `/device/v2/:uuid/logs` and `/device/v2/:uuid/log-stream` (in the `handleStreamingWrite()` method) it's good, this will track both regular and streaming logs bytes written.

src/lib/device-logs/backends/redis.ts
```diff
@@ -134,8 +140,8 @@ export class RedisBackend implements DeviceLogsBackend {
 		return this.cmds.connected && this.pubSub.connected;
 	}
 
-	private getKey(ctx: LogContext) {
-		return `device:${ctx.id}:logs`;
+	private getKey(ctx: LogContext, suffix?: string) {
+		return `device:${ctx.id}:${suffix ?? 'logs'}`;
```
Page-, line 144: Can you change to this before merging please as it'll improve automated docs when type hinting and is clearer that the param is always defaulted
```suggestion
	private getKey(ctx: LogContext, suffix = 'logs') {
		return `device:${ctx.id}:${suffix}`;
```

dt-rush: sure thing, pushing now

PR #337, opened 2020-04-17
Update balena-device-config and re-add user id

Page-:
Change-type: patch


Page-:
I don't think we need to revert it there @dfunckt as it's not used in open-balena-api for anything

PR #342, opened 2020-04-28
Add new logging backend LokiBackend

codewithcheese:
This PR is the first step towards migrating device logging from Redis to Loki. The logs will be written to both Redis and Loki but Redis will still be used for reads until Loki is ready for reads.

- New logging backend `LokiBackend`
- Add `LOKI_HOST` and `LOKI_PORT` config 
- Logs only published to `LokiBackend` if `LOKI_HOST` is defined.
- Add `loki` docker setup to test automation
- Add test for loki publish

Loki has a multi-tenant model that partitions logs by the tenant id. The tenant id can be specified using the `X-Scope-OrgID` header. In this PR the header is set to the application id as follows `'X-Scope-OrgID': ctx.belongs_to__application`.

@Page- Is `device.belongs_to__application` an appropriate way to access the app id. Do all devices belong to an application?

Change-type: minor


src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,188 @@
...
+
+	private getKey(ctx: LogContext, suffix = 'logs') {
+		return `device:${ctx.id}:${suffix}`;
+	}
+
+	private fromLokiStreamsToDeviceLogs(streams: LokiStream[]): DeviceLog[] {
+		return _.concat(
+			[],
+			...streams.map((stream: LokiStream) => this.fromLokiStream(stream)),
+		);
```
zvin, line 142: why not just 
```typescript
return streams.map((stream: LokiStream) => this.fromLokiStream(stream))
```
?

src/routes/device-logs.ts
```diff
@@ -310,13 +318,25 @@ function handleStreamingWrite(
...
+				}
 				// Clear the buffer
 				buffer.length = 0;
 				// Resume in case it was paused due to buffering
 				if (req.isPaused()) {
 					req.resume();
 				}
 				await promise;
+				try {
+					if (promise2) await promise2;
```
Page-, line 336: I think this can be just
```suggestion
					await promise2;
```
and aside from that we always use `{}` for if's - I'll have to check if it's possible to make prettier enforce them (I thought it did tbh)

src/routes/device-logs.ts
```diff
@@ -217,6 +220,11 @@ export const store: RequestHandler = async (req: Request, res: Response) => {
 		const logs: DeviceLog[] = supervisor.convertLogs(ctx, body);
 		if (logs.length) {
 			await getBackend(ctx).publish(ctx, logs);
+			try {
+				if (LOKI_HOST) await loki.publish(ctx, logs);
```
Page-, line 224: braces please

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,188 @@
...
+				stream: {
+					device_id: String(ctx.id), // write only device id, to reduce cardinality do not also write device uuid
+					service_id: String(log.serviceId),
+					is_system: String(log.isSystem),
+					is_std_err: String(log.isStdErr),
+				},
+				values: [
+					[
+						`${log.timestamp.toString()}000000`, // convert timestamp to nanosecond string
+						JSON.stringify({ version: VERSION, ...log }), // store log as JSON string to retain any meta not writen as label
```
Page-, line 180: I think doing
```
log.version = VERSION
JSON.stringify(log)
```
will have much better performance

src/lib/device-logs/backends/loki.ts
Page-, line 142: I think what you actually want here is
```suggestion
		return streams.flatMap((stream: LokiStream) => this.fromLokiStream(stream));
```
?
And you can probably simplify further with
```suggestion
		return streams.flatMap(this.fromLokiStream);
```

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,188 @@
...
+			const tailUri =
+				`ws://${LOKI_HOST}:${LOKI_PORT}/loki/api/v1/tail` +
+				`?query=${encodeURIComponent(this.getDeviceQuery(ctx))}&start=${String(
+					Date.now(),
+				)}000000`;
+			const ws = new WebSocket(tailUri, {
+				headers: { 'X-Scope-OrgID': ctx.belongs_to__application },
+			});
+			ws.on('message', (data: string) => {
+				this.fromLokiStreamsToDeviceLogs(JSON.parse(data).streams).map(
```
Page-, line 107: If this `JSON.parse(data)` throws, or parses as something that `.streams` will throw on then I think it'll crash the entire nodejs process

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,188 @@
...
+
+	public async publish(ctx: LogWriteContext, logs: DeviceLog[]): Promise<any> {
+		return requestAsync({
+			url: `http://${LOKI_HOST}:${LOKI_PORT}/loki/api/v1/push`,
+			headers: {
+				'X-Scope-OrgID': ctx.belongs_to__application,
+			},
+			method: 'POST',
+			json: {
+				streams: logs.map((log) => this.toLokiStream(ctx, log), this),
```
Page-, line 90: What's the `, this` for?

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,188 @@
...
+	public async history(ctx: LogContext, count: number): Promise<DeviceLog[]> {
+		const [response, body] = await requestAsync({
+			url: `http://${LOKI_HOST}:${LOKI_PORT}/loki/api/v1/query_range`,
+			headers: {
+				'X-Scope-OrgID': ctx.belongs_to__application,
+			},
+			method: 'GET',
+			qs: {
+				query: this.getDeviceQuery(ctx),
+				limit: _.isFinite(count) ? count : 1000,
```
Page-, line 71:
```suggestion
				limit: Number.isFinite(count) ? count : 1000,
```

Page-, line 71: When will `count` not be finite? We have typings throughout the codebase that should enforce it unless we're casting

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,188 @@
...
+		return _.concat(
+			[],
+			...streams.map((stream: LokiStream) => this.fromLokiStream(stream)),
+		);
+	}
+
+	private fromLokiStream(stream: LokiStream): DeviceLog[] {
+		try {
+			return stream.values.map((value) => {
+				const [, message] = value;
```
Page-, line 148:
```suggestion
			return stream.values.map(([, message]) => {
```

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,188 @@
+import * as _ from 'lodash';
+import * as WebSocket from 'ws';
+import { EventEmitter } from 'events';
+import * as skhema from 'skhema';
+import { JSONSchema6 } from 'skhema/node_modules/@types/json-schema';
```
Page-, line 5: This needs to be changed - the fact that this path works at all is pure chance and we should instead import `@types/json-schema` directly

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,188 @@
...
+			});
+		} catch (err) {
+			captureException(err, `Failed to deserialize a Loki log: ${stream}`);
+			return [];
+		}
+	}
+
+	private toLokiStream(ctx: LogContext, log: DeviceLog): LokiStream {
+		try {
+			skhema.validate(schema, log);
```
Page-, line 169: I think skhema has performance issues and this could potentially bring the api down - I'd want the performance to be tested/validated before merging rather than waiting for production

src/routes/device-logs.ts
Page-: https://github.com/balena-io-modules/node-balena-lint/pull/64 - once that's merged and updated to then it'll automatically add the braces for you so you can type it in whichever format you want and not worry :)

src/lib/device-logs/backends/loki.ts
codewithcheese: `this.fromLokiStream(stream)` returns an array so mapping that would result in array of array. I have re-implemented it using array.flatMap, that should be easier to parse.

src/routes/device-logs.ts
codewithcheese: Indeed `await undefined` does not throw. Thanks for the lint PR.

src/routes/device-logs.ts
codewithcheese: :+1: updated.

src/lib/device-logs/backends/loki.ts
codewithcheese: TS complains `TS2339: Property 'version' does not exist on type 'DeviceLog'.` do you recommend that I use `// @ts-ignore`?

src/lib/device-logs/backends/loki.ts
codewithcheese: Indeed much nicer! Thanks

src/lib/device-logs/backends/loki.ts
codewithcheese: Hmm yeah, I've added a conditional to check `streams` property exists and try/catch with captureException.

src/lib/device-logs/backends/loki.ts
codewithcheese: hmm good point, it snuck in there somehow :grimacing: removed.

src/lib/device-logs/backends/loki.ts
codewithcheese: Updated

codewithcheese: In `RedisBackend@61` `count === Infinity ? 0 : -count` it supports count equals `Infinity`. I implemented with `isFinite` instead. Which approach do you prefer?

src/lib/device-logs/backends/loki.ts
codewithcheese: :+1: updated

src/lib/device-logs/backends/loki.ts
codewithcheese: I will re-implement this using `io-ts` instead.

src/lib/device-logs/backends/loki.ts
codewithcheese: This was done intentionally. The version of `json-schema` types used by `skhema` is different from the version of `json-schema` in `dependencies` causing a TS error.

src/lib/device-logs/backends/loki.ts
codewithcheese: I thought `io-ts` might be slow aswell. So instead I removed `skema` and did not add any validation replacement. Do you think validation is necessary here?

src/lib/device-logs/backends/loki.ts
codewithcheese: This was removed when I remove `skhema` usage

src/routes/device-logs.ts
```diff
@@ -216,6 +219,13 @@ export const store: RequestHandler = async (req: Request, res: Response) => {
 		const logs: DeviceLog[] = supervisor.convertLogs(ctx, body);
 		if (logs.length) {
 			await getBackend(ctx).publish(ctx, logs);
+			try {
+				if (LOKI_HOST) {
+					await loki.publish(ctx, logs);
+				}
+			} catch (err) {
+				captureException(err, 'Failed to publish log to Loki');
+			}
```
Page-, line 228: I think it makes sense to put the entire try/catch inside the if, ie
```suggestion
			if (LOKI_HOST) {
				try {
					await loki.publish(ctx, logs);
				} catch (err) {
					captureException(err, 'Failed to publish log to Loki');
				}
			}
```

src/lib/device-logs/backends/loki.ts
Page-: You should be able to do
```ts
	private toLokiStream(ctx: LogContext, log: DeviceLog & { version?: number}): LokiStream {
```
and type it properly

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	public subscribe(ctx: LogContext, subscription: Subscription) {
+		const key = this.getKey(ctx);
+		if (!this.webSockets.has(key)) {
+			const tailUri =
+				`ws://${LOKI_HOST}:${LOKI_PORT}/loki/api/v1/tail` +
+				`?query=${encodeURIComponent(this.getDeviceQuery(ctx))}&start=${String(
+					Date.now(),
+				)}000000`;
```
Page-, line 84: There's no need for the `String()`
```suggestion
				`?query=${encodeURIComponent(this.getDeviceQuery(ctx))}&start=${
					Date.now()
				}000000`;
```

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,172 @@
...
+			const ws = new WebSocket(tailUri, {
+				headers: { 'X-Scope-OrgID': ctx.belongs_to__application },
+			});
+			ws.on('message', (data: string) => {
+				try {
+					const messageObj = JSON.parse(data);
+					if (messageObj.streams) {
+						this.fromLokiStreamsToDeviceLogs(messageObj.streams).map(
```
Page-, line 92: We don't need the mapped results so switching to
```suggestion
						this.fromLokiStreamsToDeviceLogs(messageObj.streams).forEach(
```
will reduce the amount of unnecessary arrays needed to be allocated/filled and then garbage collected

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,172 @@
...
+		}
+		this.subscriptions.on(key, subscription);
+	}
+
+	public unsubscribe(ctx: LogContext, subscription: Subscription) {
+		const key = this.getKey(ctx);
+		this.subscriptions.removeListener(key, subscription);
+		if (!this.webSockets.has(key)) {
+			const ws = this.webSockets.get(key);
+			if (ws) ws.close();
```
Page-, line 115: This can be
```suggestion
			ws?.close();
```

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,172 @@
...
+			return stream.values.map(([, message]) => {
+				const log = JSON.parse(message);
+				if (log.version !== VERSION) {
+					throw new Error(
+						`Invalid Redis serialization version: ${JSON.stringify(log)}`,
+					);
+				}
+				if (log.serviceId === null) {
+					delete log.serviceId;
+				}
```
Page-, line 142: When will `log.serviceId` be null? It'd make more sense to me to do this deletion when receiving the log instead

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	private toLokiStream(ctx: LogContext, log: DeviceLog): LokiStream {
+		try {
+			return {
+				stream: {
+					device_id: String(ctx.id), // write only device id, to reduce cardinality do not also write device uuid
+					service_id: String(log.serviceId),
+					is_system: String(log.isSystem),
+					is_std_err: String(log.isStdErr),
```
Page-, line 159: Does loki only accept strings?

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,172 @@
...
+	public get available(): boolean {
+		// TODO: check /ready endpoint?
+		return true;
+	}
+
+	public async history(ctx: LogContext, count: number): Promise<DeviceLog[]> {
+		const [response, body] = await requestAsync({
+			url: `http://${LOKI_HOST}:${LOKI_PORT}/loki/api/v1/query_range`,
+			headers: {
+				'X-Scope-OrgID': ctx.belongs_to__application,
```
Page-, line 48: Am I right in thinking a device will lose all of its logs if it's transferred between applications?

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,172 @@
...
+				query: this.getDeviceQuery(ctx),
+				limit: Number.isFinite(count) ? count : 1000,
+			},
+			json: true,
+		});
+		if (response.statusCode === 200) {
+			return this.fromLokiStreamsToDeviceLogs(body.data.result);
+		} else {
+			throw new BadRequestError('Query failed');
+		}
```
Page-, line 61: Would it be better to add some logging for this case so we can see something's gone wrong/fix it?

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,172 @@
...
+			this.webSockets.set(key, ws);
+		}
+		this.subscriptions.on(key, subscription);
+	}
+
+	public unsubscribe(ctx: LogContext, subscription: Subscription) {
+		const key = this.getKey(ctx);
+		this.subscriptions.removeListener(key, subscription);
+		if (!this.webSockets.has(key)) {
+			const ws = this.webSockets.get(key);
```
Page-, line 114: If we don't have the key we try to get it? I'm assuming this needs changing

src/lib/device-logs/backends/loki.ts
Page-: We use avro for the redis logs and so I'd be happy to use that without additional performance testing since it's already validated to be fast enough, otherwise I'm also happy with other validation I just want it to have performance tested as there's a lot of logs coming into the system. And yes I think it makes sense to have validation of the logs rather than potentially letting garbage in which could cause issues of its own

codewithcheese: Updated with a custom function for basic validation and speed

src/routes/device-logs.ts
codewithcheese: updated

src/lib/device-logs/backends/loki.ts
codewithcheese: removed

src/lib/device-logs/backends/loki.ts
codewithcheese: good point, updated.

src/lib/device-logs/backends/loki.ts
codewithcheese: nice, updated.

src/lib/device-logs/backends/loki.ts
codewithcheese: I have removed this. For Redis backend `avro` would apply a default which was null but for Loki to `serviceId` will not be part of the JSON string if undefined so deleting it is not necessary.

src/lib/device-logs/backends/loki.ts
codewithcheese: Yes Loki labels must be strings

src/lib/device-logs/backends/loki.ts
codewithcheese: Correct. Queries will only return a device's logs relative to its current application. I will bring up this issue in the check in.

src/lib/device-logs/backends/loki.ts
codewithcheese: I've updated the `Error` message with the status code and response body. The response body contains a error message for non 200 status codes.

src/lib/device-logs/backends/loki.ts
codewithcheese: Oops, nice catch :facepalm: updated

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,194 @@
...
+		this.subscriptions.on(key, subscription);
+	}
+
+	public unsubscribe(ctx: LogContext, subscription: Subscription) {
+		const key = this.getKey(ctx);
+		this.subscriptions.removeListener(key, subscription);
+		if (this.webSockets.has(key)) {
+			const ws = this.webSockets.get(key);
+			ws?.close();
+		}
```
Page-, line 118: I think you can simplify this to
```suggestion
		const ws = this.webSockets.get(key);
		ws?.close();
```
and only look up the key once rather than once to check it exists and then again to actually retrieve it

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,194 @@
...
+				query: this.getDeviceQuery(ctx),
+				limit: Number.isFinite(count) ? count : 1000,
+			},
+			json: true,
+		});
+		if (response.statusCode === 200) {
+			return this.fromLokiStreamsToDeviceLogs(body.data.result);
+		} else {
+			throw new BadRequestError(
+				`Query failed ${response.statusCode} "${response.body}"`,
```
Page-, line 61: This message will be returned to the user, I don't know in what ways this could fail/what could be contained in the body but I'd rather `captureException` with the extra info for us to debug and send a message that makes sense for a user - avoiding exposing potentially unknown/sensitive info that might be in `response.body`

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,194 @@
...
+					is_system: String(log.isSystem),
+					is_std_err: String(log.isStdErr),
+				},
+				values: [
+					[
+						`${log.timestamp}000000`, // convert timestamp to nanosecond string
+						JSON.stringify(log), // store log as JSON string to retain any meta not writen as label
+					],
+				],
+			};
```
Page-, line 188: This seems like a lot of JSON to be creating for every single log line and I would expect the amount of object creation/garbage collection and JSON serializing it requires to add up with the amount of logs we have to handle - is it possible to reduce this?

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,194 @@
...
+		} catch (err) {
+			// Rethrow with a type of error that will end up as status 400
+			throw new BadRequestError(err);
```
Page-, line 191: The only thing that can throw is the `this.validateLog(log);` right? In that case I'd prefer they throw `BadRequestError`s directly and avoid this rewrapping

src/lib/device-logs/backends/loki.ts
codewithcheese: Now using protobuf instead, and grouping streams with similar labels

src/lib/device-logs/backends/loki.ts
codewithcheese: updated

src/routes/device-logs.ts
```diff
@@ -216,6 +221,13 @@ export const store: RequestHandler = async (req: Request, res: Response) => {
 		const logs: DeviceLog[] = supervisor.convertLogs(ctx, body);
 		if (logs.length) {
 			await getBackend(ctx).publish(ctx, logs);
+			if (LOKI_HOST && LOKI_WRITE_PCT > randomPct()) {
+				try {
+					await loki.publish(ctx, logs);
+				} catch (err) {
+					captureException(err, 'Failed to publish log to Loki');
+				}
+			}
```
Page-, line 230: It's probably worth starting both publish in parallel rather in serial, otherwise we could introduce a bunch of latency to responding which might have a knock on effect

Page-, line 230: Having seen this logic is duplicated in two places I think it'd be worth creating a helper function that does
```ts
const lokiPublish = async (ctx, logs) =>
	if (LOKI_HOST && LOKI_WRITE_PCT > randomPct()) {
		try {
			await loki.publish(ctx, logs);
		} catch (err) {
			captureException(err, 'Failed to publish log to Loki');
		}
	}
```
which will make the calling sites much simpler and also avoid the need to juggle promises as much, eg this site could be simplified to
```ts
await Promise.all([
	getBackend(ctx).publish(ctx, logs),
	lokiPublish(ctx, logs),
]);

File: src/lib/device-logs/backends/loki.ts [hunk 25]
```diff
@@ -0,0 +1,302 @@
...
+				pushRequest,
+				createOrgIdMetadata(String(appId)),
+				(err: Error, pushResponse: PushResponse) => {
+					if (err) {
+						reject(err);
+					} else {
+						resolve(pushResponse);
+					}
+				},
+			);
```
Page-, line 174: Is it possible to promisify this instead?

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,302 @@
...
+				} catch (error) {
+					// only retry when Loki is down or rate limiting
+					if (
+						tries <= MAX_RETRIES &&
+						[grpc.status.UNAVAILABLE, grpc.status.RESOURCE_EXHAUSTED].includes(
+							error.code,
+						)
+					) {
+						this.ready = false;
+						await sleep(nextBackoff);
```
Page-, line 146:
```suggestion
						await Bluebird.delay(nextBackoff);
```

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,302 @@
...
+	public async publish(
+		ctx: LogWriteContext,
+		logs: Array<DeviceLog & { version?: number }>,
+	): Promise<any> {
+		const streams = this.fromDeviceLogsToStreams(ctx, logs);
+		try {
+			let tries = 0;
+			let nextBackoff = MIN_BACKOFF;
+			let prevBackoff = MIN_BACKOFF;
+			while (nextBackoff <= MAX_BACKOFF) {
```
Page-, line 133: Why do we have both a max retries and max backoff? I would expect only one can ever apply in reality and so doing
```suggestion
			for (let tries = 0; tries < MAX_BACKOFF; tries++) {
```
would be more straightforward, and then you can also use
```ts
await Bluebird.delay(Math.min(MAX_BACKOFF, nextBackoff));
```
to cap the backoff to reasonable levels

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,302 @@
...
+						nextBackoff = nextBackoff + prevBackoff;
+						prevBackoff = nextBackoff - prevBackoff;
+					} else {
+						throw new BadRequestError(`Publish failed: ${error.details}`);
+					}
+				}
+			}
+		} finally {
+			this.ready = true;
+		}
```
Page-, line 157: Since this is in a `finally` it means that if we run out of retries we'll mark as ready even though we're not - that probably needs fixing

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,302 @@
...
+
+			const deviceLogs: DeviceLog[] = [];
+			const call = this.querier.query(
+				queryRequest,
+				createOrgIdMetadata(String(ctx.belongs_to__application)),
+			);
+			call.on('data', (queryResponse: QueryResponse) => {
+				deviceLogs.push(
+					...this.fromStreamsToDeviceLogs(queryResponse.getStreamsList()),
+				);
```
Page-, line 113: I think it would make sense to do
```suggestion
				streamLogs.push(
					...queryResponse.getStreamsList(),
				);
```
and then
```ts
const streamLogs = await new Promise(...);
return _.orderBy(this.fromStreamsToDeviceLogs(deviceLogs), 'timestamp', 'desc');
```
so that we only process the logs if we need to, since if an error is hit partway through then we can skip the work. The counterpoints are:
1. if the processing would potentially lock up the event loop, but if that's possible then I think we'd need to put in dedicated measures regardless of when we process
2. if the success latency is particularly important to minimize at the cost of potentially wasted work

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,302 @@
...
+	 *
+	 * Return $count of logs matching device_id in descending (BACKWARD) order.
+	 *
+	 * The logs are sorted by timestamp since Loki returns a distinct stream for each label combination.
+	 *
+	 * @param ctx
+	 * @param count
+	 */
+	public history(ctx: LogContext, count: number): Promise<DeviceLog[]> {
+		return new Promise((resolve, reject) => {
```
Page-, line 95: As a general point I prefer to put the `new Promise` around the absolute minimum code to show precisely what is being wrapped and can resolve/reject, eg
```ts
new Promise((resolve, reject) => {
	call.on('error', (error: Error & { details: string }) => {
		reject(...);
	});
	call.on('end', () => {
		resolve(...);
	});
});
```

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,302 @@
...
+}
+
+function createTimestampFromMilliseconds(milliseconds: number) {
+	const MS_TO_NANOS = 1000000;
+	const seconds = Math.floor(milliseconds / 1000);
+	const nanos = (milliseconds - seconds * 1000) * MS_TO_NANOS;
+
+	const timestamp = new Timestamp();
+	timestamp.setSeconds(seconds);
+	timestamp.setNanos(nanos);
```
Page-, line 56: Is it possible to pass this in the constructor, eg
```suggestion
	const timestamp = new Timestamp(seconds, nanos);
```
would be far nicer and seems sane enough as an option - I know we might be restricted by the library but if it's possible to do it'd be nice to

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,302 @@
...
+			let prevBackoff = MIN_BACKOFF;
+			while (nextBackoff <= MAX_BACKOFF) {
+				try {
+					tries += 1;
+					return await this.push(ctx.belongs_to__application, streams);
+				} catch (error) {
+					// only retry when Loki is down or rate limiting
+					if (
+						tries <= MAX_RETRIES &&
+						[grpc.status.UNAVAILABLE, grpc.status.RESOURCE_EXHAUSTED].includes(
```
Page-, line 141: These seem like they should be exported via loki-grpc-client since that's what's returning them

src/lib/device-logs/backends/loki.ts
codewithcheese: Updated. Now `BadRequestError` returns a generic message and `captureException` is called with the error object before `BadRequest` error is thrown

test/07-loki-backend.ts
```diff
@@ -0,0 +1,118 @@
...
+				try {
+					expect(incomingLog).to.deep.equal(log);
+					resolve();
+				} catch (err) {
+					reject(err);
+				}
+			});
+			const log = createLog();
+			loki.publish(ctx, [_.clone(log)]);
+			await sleep(1500);
```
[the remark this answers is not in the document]
codewithcheese: Removed `sleep`, using `Bluebird.delay` going forward

test/07-loki-backend.ts
```diff
@@ -0,0 +1,118 @@
...
+			if (!received) {
+				reject(Error('Subscription did not receive log'));
+			}
```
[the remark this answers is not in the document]
codewithcheese: Updated. Indeed its cleaner, timing logic is minimized.

test/07-loki-backend.ts
```diff
@@ -0,0 +1,118 @@
...
+
+	it('should push multiple logs with different labels and return in order', async function () {
+		const loki = new LokiBackend();
+		const ctx = createContext();
+		const logs = [
+			createLog({ timestamp: Date.now() - 4 }),
+			createLog({ timestamp: Date.now() - 3 }),
+			createLog({ timestamp: Date.now() - 2, isStdErr: false }),
+			createLog({ timestamp: Date.now() - 1, isStdErr: false }),
+			createLog({ timestamp: Date.now(), isStdErr: false, isSystem: false }),
```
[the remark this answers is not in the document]
codewithcheese: Yep good point updated.

src/routes/device-logs.ts
```diff
@@ -381,6 +408,7 @@ async function getWriteContext(
 	})) as Array<{
 		id: number;
 		logs_channel?: string;
+		belongs_to__application: number;
```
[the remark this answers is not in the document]
codewithcheese: Updated.

src/lib/device-logs/backends/loki.ts
codewithcheese: Ok makes sense, updated.

src/lib/device-logs/backends/loki.ts
codewithcheese: True, I have updated and its easier to read.

src/lib/device-logs/backends/loki.ts
codewithcheese: Updated.

src/lib/device-logs/backends/loki.ts
codewithcheese: Yeah its awkward, this is auto generated from the proto schema. I could give it a wrapper but that would be one more thing to check/maintain when updating the schema for new versions.

src/lib/device-logs/backends/loki.ts
codewithcheese: The backend will retry publishing and mark the backend as not ready, after the max backoff is reached it throws and the logs are essentially dropped. The backend is marked `ready` again so that the next batch can attempt to publish. If its not marked as ready here, not sure how else it would be.

src/routes/device-logs.ts
```diff
@@ -309,13 +321,28 @@ function handleStreamingWrite(
...
 				// Resume in case it was paused due to buffering
 				if (req.isPaused()) {
 					req.resume();
 				}
 				await promise;
+				try {
+					await promise2;
+				} catch (err) {
+					captureException(err, 'Failed to publish logs to Loki');
+				}
```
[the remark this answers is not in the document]
codewithcheese: updated

src/routes/device-logs.ts
codewithcheese: Good point, updated

codewithcheese: Moved conditional logic to function, and implemented `Promise.all`

src/lib/device-logs/backends/loki.ts
codewithcheese: Yep, done

src/lib/device-logs/backends/loki.ts
codewithcheese: Updated

src/lib/device-logs/backends/loki.ts
codewithcheese: Yep `tries` and `MAX_BACKOFF` was overly redundant. Removed the `tries` logic but kept the back-off logic the same. Let me know if its acceptable.

src/routes/device-logs.ts
```diff
@@ -307,14 +323,22 @@ function handleStreamingWrite(
...
 				// Even if the connection was closed, still flush the buffer
-				const promise = backend.publish(ctx, buffer);
+				const publishingToRedis = backend.publish(ctx, buffer);
+				const publishingToLoki = shouldPublishToLoki()
+					? loki
+							.publish(ctx, buffer)
+							.catch((err) =>
+								captureException(err, 'Failed to publish logs to Loki'),
+							)
+					: Promise.resolve();
```
Page-, line 333:
```suggestion
					: undefined;
```

src/routes/device-logs.ts
```diff
@@ -214,7 +220,17 @@ export const store: RequestHandler = async (req: Request, res: Response) => {
...
+			// start publishing to both backends
+			await Promise.all([
+				getBackend(ctx).publish(ctx, logs),
+				shouldPublishToLoki()
+					? loki
+							.publish(ctx, logs)
+							.catch((err) =>
+								captureException(err, 'Failed to publish logs to Loki'),
+							)
+					: Promise.resolve(),
```
Page-, line 232:
```suggestion
					: undefined,
```

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,296 @@
...
+		return `app:${ctx.belongs_to__application}:device:${ctx.id}:${suffix}`;
+	}
+
+	private getLabels(ctx: LogContext, log: DeviceLog): string {
+		return `{device_id="${ctx.id}", service_id="${
+			log.serviceId ? log.serviceId : 'null'
+		}", is_system="${log.isSystem}", is_std_err="${log.isStdErr}"}`;
+	}
+
+	private validateLog(log: DeviceLog) {
```
Page-, line 228: You can mark this as an assertion function with something like:
```suggestion
	private validateLog(log: DeviceLog): asserts log is $type {
```

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,296 @@
...
+		return `{device_id="${ctx.id}"}`;
+	}
+
+	private getKey(ctx: LogContext, suffix = 'logs') {
+		return `app:${ctx.belongs_to__application}:device:${ctx.id}:${suffix}`;
+	}
+
+	private getLabels(ctx: LogContext, log: DeviceLog): string {
+		return `{device_id="${ctx.id}", service_id="${
+			log.serviceId ? log.serviceId : 'null'
```
Page-, line 224: Is this meant to be checking for nullish or falsey? If it's meant to check nullish which is my assumption then it'd be better as:
```suggestion
			log.serviceId ?? 'null'
```

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	public async publish(
+		ctx: LogWriteContext,
+		logs: Array<DeviceLog & { version?: number }>,
+	): Promise<any> {
+		const streams = this.fromDeviceLogsToStreams(ctx, logs);
+		try {
+			let nextBackoff = MIN_BACKOFF;
+			let prevBackoff = MIN_BACKOFF;
```
Page-, line 139: Do the existing logs include this exponential backoff? In general I think it'd be useful to separate out the exponential backoff part into its own function to make both the exponential backoff and the publishing logic clearer/easier to understand

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,296 @@
...
+				call.on('end', () => {
+					resolve(streams);
+				});
+			},
+		);
+		return _.orderBy(
+			this.fromStreamsToDeviceLogs(responseStreams),
+			'timestamp',
+			'desc',
+		);
```
Page-, line 129: Am I right in understand that when getting history we return the newest entry first and the oldest last? That seems strange to me - is that what the redis backend does?

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,296 @@
...
+			queryRequest,
+			createOrgIdMetadata(String(ctx.belongs_to__application)),
+		);
+		const responseStreams: StreamAdapter[] = await new Promise(
+			(resolve, reject) => {
+				call.on('data', (queryResponse: QueryResponse) => {
+					streams.push(...queryResponse.getStreamsList());
+				});
+				call.on('error', (error: Error & { details: string }) => {
+					captureException(error);
```
Page-, line 115: Can you add some blurb to the captureException so it's a bit easier to understand where it comes from when looking in sentry

src/lib/device-logs/backends/loki.ts
Page-: If we don't have control over it then that's fair enough, I was just being hopeful :P

package.json
```diff
@@ -58,10 +58,12 @@
     "eventemitter3": "^4.0.4",
     "express": "^4.17.1",
     "fast-boot2": "^1.1.0",
+    "grpc": "^1.24.3",
     "ipaddr.js": "^1.9.1",
     "json-schema": "^0.2.5",
     "jsonwebtoken": "^8.5.1",
     "lodash": "^4.17.19",
+    "loki-grpc-client": "1.0.0-loki-client-4437232d59aaf48f511210b9f11cd8dedda7a521",
```
Page-, line 66: This will need to be updated before merging

src/lib/device-logs/backends/loki.ts
```diff
@@ -0,0 +1,296 @@
...
+	timestamp.setNanos(nanos);
+	return timestamp;
+}
+
+export class LokiBackend implements DeviceLogsBackend {
+	private subscriptions: EventEmitter;
+	private querier: QuerierClient;
+	private pusher: IGrpcClientAsync;
+	private ready: boolean;
+	private tailCalls: Map<string, grpc.ClientReadableStream<TailResponse>>;
```
Page-, line 63: `grpc.ClientReadableStream` seems like it'd make sense to export alongside `QuerierClient` since the semantic type we care about is `ReturnType<QuerierClient['tail']>`

src/lib/device-logs/backends/loki.ts
codewithcheese: I've removed the logic that sets ready to false when retrying. Not convinced its a good solution, will test if back-off is enough.

src/routes/device-logs.ts
codewithcheese: Updated, do you prefer `undefined` for speed?

src/routes/device-logs.ts
codewithcheese: Updated

src/lib/device-logs/backends/loki.ts
codewithcheese: Good point

src/lib/device-logs/backends/loki.ts
codewithcheese: Your right the order is incorrect. I have updated it. The logic for fetching `history` will likely change before we switch to reading from Loki, the reason is that with redis it was only required to request all the logs stored, where with Loki we want to display the last N logs but its uncertain at what time period those logs occurred. So multiple requests scanning backwards might be necessary. I have set the `LokiBackend` history to just retrieve all logs from the last hour, but for devices that don't log very often it may return just a few or 0 logs.

src/lib/device-logs/backends/loki.ts
codewithcheese: Updated

src/lib/device-logs/backends/loki.ts
codewithcheese: Updated

src/lib/device-logs/backends/loki.ts
codewithcheese: Yep its cleaner, I have moved the backoff logic to a funciton

src/lib/device-logs/backends/loki.ts
codewithcheese: Interesting did not know about this feature, updated.

test/09-loki-backend.ts
```diff
@@ -0,0 +1,135 @@
...
+				try {
+					expect(incomingLog).to.deep.equal(log);
+					resolve();
+				} catch (err) {
+					reject(err);
+				}
+			});
+			const log = createLog();
+			loki.publish(ctx, [_.clone(log)]);
+		}).timeout(5000, 'Subscription did not receive log');
```
Page-, line 107: This would be clearer imo as
```suggestion
		const log = createLog();
		const incomingLog = await new Bluebird(async (resolve) => {
			loki.subscribe(ctx, resolve);
			loki.publish(ctx, [_.clone(log)]);
		}).timeout(5000, 'Subscription did not receive log');
		expect(incomingLog).to.deep.equal(log);
```

src/features/device-logs/lib/backends/loki.ts
```diff
@@ -0,0 +1,302 @@
...
+function createTimestampFromDate(date = new Date()) {
+	const timestamp = new Timestamp();
+	timestamp.fromDate(date);
+	return timestamp;
+}
+
+function convertToNanoseconds(milliseconds: number, nonce: number) {
+	const MS_TO_NANOS = 1000000;
+	const seconds = Math.floor(milliseconds / 1000);
+	const nanos = (milliseconds - seconds * 1000) * MS_TO_NANOS;
```
Page-, line 50: If you just want the ms portion without the seconds then using a modulo makes more sense, ie
```suggestion
	const nanos = (milliseconds % 1000) * MS_TO_NANOS;
```

src/features/device-logs/lib/backends/loki.ts
```diff
@@ -0,0 +1,302 @@
...
+
+	// use nonce to create unique nanosecond component
+	return [seconds, nanos + nonce];
+}
+
+function backoff(fn: (...args: any) => any, retryIf: (err: Error) => boolean) {
+	return async (...args: any) => {
```
Page-, line 57: You can get typings for this with:
```suggestion
function backoff<T extends (...args: any[]) => any>(fn: T, retryIf: (err: Error) => boolean) {
	return async (...args: Parameters<T>): ReturnType<T> => {
```

automation/common.sh
```diff
@@ -84,6 +94,7 @@ runapi () {
 		-e SUPERUSER_PASSWORD=foobarbaz\
 		$extra_vol_args \
 		$extra_env_args \
+		-p 9229:9229 \
```
Page-, line 97: There's a `--profile` option already in at least `fasttest` for exposing the inspector port, and it also enables the debugger/profiler at the same time

codewithcheese: Yeah I have been using `--profile` but it did not seem to open a port on `9229`. Now taking a closer look I see I should have been attaching to `9228` `api_id=$(runapi $IMAGE_NAME $db_id $redis_id $loki_id " -p 9228:9229 -v $(pwd):/usr/src/app")`. is the port change intentional?

Removed `-p 9229:9229 \`.

src/features/device-logs/lib/backends/loki.ts
codewithcheese: Thanks, my type foo is not that strong yet.

test/09-loki-backend.ts
codewithcheese: Good point. much nicer. updated.

src/features/device-logs/lib/backends/loki.ts
codewithcheese: Good point, updated.

automation/common.sh
Page-: Ah yeah, that was to help avoid clashes with other debugger ports that might be open

test/09-loki-backend.ts
```diff
@@ -0,0 +1,128 @@
...
+			createLog({ serviceId: 2 }),
+			createLog({ serviceId: 3 }),
+		];
+		// @ts-ignore usage of private function
+		const streams = loki.fromDeviceLogsToStreams(ctx, _.cloneDeep(logs));
+		expect(streams.length).to.be.equal(
+			4,
+			'should be 4 streams since 5 logs have 4 distinct services (null, 1, 2, 3)',
+		);
+		// @ts-ignore usage of private function
```
Page-, line 60: Can you make this a
```suggestion
		// @ts-expect-error usage of private function
```
just for the slight extra safety, and same for the other instances of `ts-ignore` where we'd expect an error

package.json
```diff
@@ -65,6 +65,7 @@
     "json-schema": "^0.2.5",
     "jsonwebtoken": "^8.5.1",
     "lodash": "^4.17.20",
+    "loki-grpc-client": "1.0.0-loki-client-f6cc2ad4ce08191ebc06fe7b961e093df8298644",
```
Page-, line 68: This'll need updating before an approve (since approve = merge)

codewithcheese: updated

test/09-loki-backend.ts
codewithcheese: updated

Page- (approved):

codewithcheese:
@balena-ci I self-certify!

PR #344, opened 2020-04-30
Select "email" on findUser query

iamsolankiamit:
Change-type: minor
Signed-off-by: Amit Solanki <amit@balena.io>

### DIFF
--- a/src/platform/auth.ts
+++ b/src/platform/auth.ts
@@ -349,7 +349,7 @@ export const findUser = async (
 	} else {
 		loginField = 'username';
 	}
-	const $select = ['id', 'actor', 'username', 'password'] as const;
+	const $select = ['id', 'actor', 'username', 'password', 'email'] as const;
 	type UserResult = Pick<DbUser, typeof $select[number]>;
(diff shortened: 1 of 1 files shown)


thgreasi:
I would personally prefer to allow passing an array of $selected props, so that we can avoid $electing unnecessary columns.
That's a fun TS exercise though.
Let me open a draft PR for this.

iamsolankiamit:
@thgreasi so just an optional param $select with defaults which we already have?

thgreasi:
@iamsolankiamit let me know what you think about:
See: https://github.com/balena-io/open-balena-api/pull/346

thgreasi:
Well wrt
> so just an optional param $select

it's not a **just** in the TS world while trying to get TS both to compile and emit the correct types...

iamsolankiamit:
😸 Ya I figured, I forget we are using TS sometimes. That PR looks solid to me. Good one 👍

iamsolankiamit:
closing in favour of https://github.com/balena-io/open-balena-api/pull/346

PR #346, opened 2020-05-01
platform/auth/findUser: Add support for an optional $select argument

thgreasi:
This allows to do calls like:
```
const x = findUser('asadf'); // Promise<Pick<DbUser, "id" | "actor" | "username" | "password"> | undefined>
const y = findUser('asadf', undefined, ['email']); // Promise<Pick<DbUser, "email"> | undefined>
```

Change-type: minor
Signed-off-by: Thodoris Greasidis <thodoris@balena.io>

PR #348, opened 2020-05-01
Select only the required fields in findUser calls

thgreasi:
That's on top of #346

Change-type: patch
Depends-on: https://github.com/balena-io/open-balena-api/pull/346


thgreasi:
Well I still have to wait for an extra review on #346 :)

PR #349, opened 2020-05-01
Prepare the heartbeat interval queries only once, on first use

thgreasi:
It seems that so far these were re-calculated on each
request. The queries are now prepared lazily, so the
startup time will not be affected, but had to do this
anyway since at the point that this loaded api.resin
was still undefined :)

Change-type: patch

### DIFF
--- a/src/lib/device-online-state.ts
+++ b/src/lib/device-online-state.ts
@@ -21,8 +21,8 @@ import {
 
 const { root, api } = sbvrUtils;
 
-export const getPollInterval = async (uuid: string) => {
-	const getPollIntervalForDevice = api.resin.prepare<{ uuid: string }>({
+const getPollIntervalForDevice = _.once(() =>
+	api.resin.prepare<{ uuid: string }>({
 		resource: 'device_config_variable',
 		passthrough: { req: root },
 		options: {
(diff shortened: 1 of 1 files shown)


src/lib/device-online-state.ts
```diff
@@ -21,8 +21,8 @@ import {
 
 const { root, api } = sbvrUtils;
 
-export const getPollInterval = async (uuid: string) => {
-	const getPollIntervalForDevice = api.resin.prepare<{ uuid: string }>({
+const getPollIntervalForDevice = _.once(() =>
```
Page-, line 24: There's no need to do it lazily as it's pretty cheap/quick - you might as well do it eagerly to simplify the code

thgreasi: @Page- is mentioned in the description, if I do this inline it will error with `api.resin is undefined`.
I didn't really had in mind to use lazy loading, I just had to, in order to avoid the `undefined` issue, which seems import/setup order related.
Is that what you meant with "eagerly"?

Page-: Oh right, I guess this file is required fairly early on, fair enough

PR #350, opened 2020-05-04
Allow gateway downloads to have a null download progress

Page-:
This matches other download progress entries, and the supervisor attempts to send null progress in at least some cases

Change-type: minor


src/migrations/00027-gateway-download-nullable-download-progress.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "gateway download"
+ALTER COLUMN "download progress" SET NOT NULL;
```
dfunckt, line 2:
```suggestion
ALTER COLUMN "download progress" SET NULL;
```

Page-:
@balena-ci rebase

PR #355, opened 2020-05-11
Handle false values correctly when updating the DT model

sradevski:
Change-type: patch


sradevski:
@balena-ci rebase

Issue #321, opened 2020-03-31
The API should set a supervisor_release in the state endpoint when it's null


CameronDiver:
We are aiming to make it so that all devices have an assigned supervisor_release field. To do this we should use the supervisor version in the state endpoint and find the correct supervisor_release based on that version, and set that in the device resource.

xginn8:
We will now set the `should_be_managed_by__supervisor_release` value upon provision for new devices corresponding to what the device is running already. This is now deployed in production.

PR #356, opened 2020-05-14
supervisor release: set corresponding resource if current state is null

xginn8:
HQ: balena-io/balena#2177
Change-type: patch

### DIFF
--- a/src/hooks/resources/device.ts
+++ b/src/hooks/resources/device.ts
@@ -1,3 +1,4 @@
+import * as _ from 'lodash';
 import * as Bluebird from 'bluebird';
 
 import { sbvrUtils } from '@resin/pinejs';
@@ -207,7 +208,8 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
 			request.values.should_be_running__release !== undefined ||
 			[false, 0].includes(request.values.is_online) ||
 			request.values.belongs_to__application != null ||
-			request.values.device_name != null
+			request.values.device_name != null ||
+			request.values.supervisor_version != null
 		) {
 			// Cache affected ids for later
 			waitPromises.push(getCurrentRequestAffectedIds(args));
@@ -353,6 +355,20 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
 						request.values.should_be_managed_by__supervisor_release,
 					),
 				),
+			);
+		}
+
+		if (request.values.supervisor_version != null) {
+			// When a device checks in with it's initial supervisor version, set the corresponding supervisor_release
+			// resource using its current version
+			waitPromises.push(
+				getCurrentRequestAffectedIds(args).then(async (ids) => {
+					await setSupervisorReleaseResource(
+						api,
+						ids,
+						request.values.supervisor_version,
+					);
+				}),
 			);
 		}
 
@@ -577,4 +593,86 @@ async function checkSupervisorReleaseUpgrades(
 			);
 		}
 	}
+}
+
+async function setSupervisorReleaseResource(
+	api: sbvrUtils.PinejsClient,
+	deviceIds: number[],
+	supervisorVersion: string,
+) {
+	if (deviceIds.length === 0) {
+		return;
+	}
+	const devices = (await api.get({
+		resource: 'device',
+		options: {
+			// if the device already has a supervisor_version, just bail.
+			$filter: {
+				id: { $in: deviceIds },
+				supervisor_version: null,
+			},
+			$select: ['id', 'is_of__device_type'],
+		},
+	})) as AnyObject;
+
+	if (devices.length === 0) {
+		return;
+	}
+
+	const devicesByDeviceType = _.groupBy(devices, (d) => {
+		return d.is_of__device_type.__id;
+	});
+
+	if (Object.keys(devicesByDeviceType).length === 0) {
+		return;
+	}
+
+	const rootApi = api.clone({
+		passthrough: {
+			req: root,
+		},
+	});
+
+	return Promise.all(
+		_.map(devicesByDeviceType, async (affectedDevices, deviceType) => {
+			const affectedDeviceIds = affectedDevices.map((d) => d.id);
+
+			const supervisorRelease = (await rootApi.get({
+				resource: 'supervisor_release',
+				options: {
+					$select: ['id'],
+					$filter: {
+						supervisor_version: `v${supervisorVersion}`,
+						is_public: true,
+						is_for__device_type: {
+							$any: {
+								$alias: 'dt',
+								$expr: {
+									dt: {
+										id: deviceType,
+									},
+								},
+							},
+						},
+					},
+				},
+			})) as AnyObject[];
+
+			if (supervisorRelease.length === 0) {
+				return;
+			}
+
+			await rootApi.patch({
+				resource: 'device',
+				options: {
+					$filter: {
+						id: { $in: affectedDeviceIds },
+					},
+				},
+				body: {
+					should_be_managed_by__supervisor_release: supervisorRelease[0].id,
+				},
+			});
+		}),
+	);
 }
--- a/test/test-lib/fake-device.ts
+++ b/test/test-lib/fake-device.ts
@@ -26,7 +26,8 @@ interface DeviceStateApp {
 
 export interface DeviceState {
 	local: {
-		name: string;
+		name?: string;
+		supervisor_version?: string;
 		config: _.Dictionary<string>;
 		apps: _.Dictionary<DeviceStateApp>;
 	};
@@ -60,8 +61,6 @@ export async function provisionDevice(admin: UserObjectParam, appId: number) {
 			belongs_to__application: appId,
 			uuid: uuid.v4().replace(/\-/g, '').toLowerCase(),
 			device_type: deviceType,
-			os_version: '2.38.0+rev1',
-			supervisor_version: 'v10.0.0',
 		})
 		.expect(201);
 
--- a/test/03-device-state-v2.ts
+++ b/test/03-device-state-v2.ts
@@ -182,7 +182,7 @@ describe('Device State v2', () => {
 				stateMock.getInstance().off('stats', statsEventSpy);
 			});
 
-			it('Should see state initialy as "unknown"', async () => {
+			it('Should see state initially as "unknown"', async () => {
 				const { body } = await supertest(app, admin)
 					.get(`/resin/device(${device.id})`)
 					.expect(200);
--- a/test/05-supervisor-release.ts
+++ b/test/05-supervisor-release.ts
@@ -39,7 +39,7 @@ describe('supervisor release', function () {
 				'/resin/supervisor_release?$select=id,image_name,supervisor_version,is_for__device_type',
 			)
 			.expect(200);
-		expect(res.body).to.have.nested.property('d.length', 1);
+		expect(res.body).to.have.nested.property('d.length', 3);
 		expect(res.body.d[0]).to.have.nested.property('image_name');
 		expect(res.body.d[0]).to.have.nested.property('supervisor_version');
 		expect(res.body.d[0]).to.have.nested.property('is_for__device_type');
@@ -55,6 +55,7 @@ describe('supervisor release', function () {
 	describe('Devices running supervisor releases', () => {
 		let applicationId: number;
 		let device: fakeDevice.Device;
+		let device2: fakeDevice.Device;
 		let fx: fixtures.Fixtures;
 		let supervisorReleases: Dictionary<{ id: number }>;
 
@@ -64,6 +65,7 @@ describe('supervisor release', function () {
 			applicationId = fx.applications.app1.id;
 
 			device = await fakeDevice.provisionDevice(admin, applicationId);
+			device2 = await fakeDevice.provisionDevice(admin, applicationId);
 			supervisorReleases = fx['supervisor-release'];
 		});
 
@@ -71,14 +73,51 @@ describe('supervisor release', function () {
 			await fixtures.clean(fx);
 		});
 
-		it('should allow setting a device to a supervisor release', async () => {
+		it('should be provisioned with a non-null supervisor release after device PATCH', async () => {
 			await supertest(app, admin)
 				.patch(`/resin/device(${device.id})`)
 				.send({
-					should_be_managed_by__supervisor_release:
-						supervisorReleases['5.0.1'].id,
+					os_version: '2.38.0+rev1',
+					supervisor_version: '5.0.1',
 				})
 				.expect(200);
+
+			const res = await supertest(app, admin).get(
+				`/resin/device(${device.id})?$expand=should_be_managed_by__supervisor_release`,
+			);
+			expect(res.body)
+				.to.have.nested.property(
+					'd[0].should_be_managed_by__supervisor_release[0].supervisor_version',
+				)
+				.that.equals(`v${res.body.d[0].supervisor_version}`);
+		});
+
+		it('should be set to a non-null supervisor release after state endpoint PATCH', async () => {
+			await device2.patchStateV2({
+				local: {
+					api_port: 48484,
+					api_secret: 'somesecret',
+					os_version: '2.38.0+rev1',
+					os_variant: 'dev',
+					supervisor_version: '5.0.1',
+					provisioning_progress: null,
+					provisioning_state: '',
+					status: 'Idle',
+					logs_channel: null,
+					update_failed: false,
+					update_pending: false,
+					update_downloaded: false,
+				},
+			});
+
+			const res = await supertest(app, admin).get(
+				`/resin/device(${device2.id})?$expand=should_be_managed_by__supervisor_release`,
+			);
+			expect(res.body)
+				.to.have.nested.property(
+					'd[0].should_be_managed_by__supervisor_release[0].supervisor_version',
+				)
+				.that.equals(`v${res.body.d[0].supervisor_version}`);
 		});
 
 		it('should allow upgrading to a logstream version', async () => {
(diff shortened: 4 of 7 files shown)


src/hooks/resources/device.ts
```diff
@@ -572,3 +585,58 @@ async function checkSupervisorReleaseUpgrades(
...
+						},
+					},
+				},
+			},
+		},
+	})) as AnyObject[];
+
+	if (supervisorRelease == null) {
+		throw new BadRequestError(
+			// TODO: maybe this shouldn't throw (for the oB case with empty supervisor_releases)
```
xginn8, line 630: This error has two disparate cases:

1. In oB, I believe we expect no `supervisor_release`s to exist by design (but this is not necessarily always true).
1. In cloud, we want to not allow a device to set itself to a non-existent release (though the mechanism of "not allow" is perhaps vague and needs to be better elucidated).

Perhaps this means that this code is better suited for the closed-source API?

src/hooks/resources/device.ts
```diff
@@ -353,6 +353,19 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
 				),
 			);
 		}
+		if (request.values.supervisor_version) {
```
Page-, line 356: This would be safer as
```suggestion
		if (request.values.supervisor_version != null) {
```

src/hooks/resources/device.ts
```diff
@@ -572,3 +585,58 @@ async function checkSupervisorReleaseUpgrades(
...
+	deviceIds: number[],
+	supervisorVersion: string,
+) {
+	// This function _should_ get called with one device (since each supervisor reports itself)
+	const device = (await api.get({
+		resource: 'device',
+		id: deviceIds[0],
+		options: {
+			$expand: ['is_of__device_type'],
+		},
```
Page-, line 600: Please add `$select`s

src/hooks/resources/device.ts
Page-: There's also the case of testing releases where they probably have a version that doesn't exist right?

src/hooks/resources/device.ts
```diff
@@ -572,3 +585,58 @@ async function checkSupervisorReleaseUpgrades(
...
+		);
+	}
+
+	return await api.patch({
+		resource: 'device',
+		id: device.id,
+		body: {
+			should_be_managed_by__supervisor_release: supervisorRelease[0].id,
+		},
+	});
```
Page-, line 641: If it needs to be done this way can you add a
```ts
$filter: {
	should_be_managed_by__supervisor_release: { $ne: supervisorRelease[0].id }
}
```
please, to avoid work if it hasn't changed

src/hooks/resources/device.ts
```diff
@@ -572,3 +585,58 @@ async function checkSupervisorReleaseUpgrades(
...
+	})) as AnyObject[];
+
+	if (supervisorRelease == null) {
+		throw new BadRequestError(
+			// TODO: maybe this shouldn't throw (for the oB case with empty supervisor_releases)
+			`Could not find a supervisor release for this device type & version (${device.is_of__device_type}, ${supervisorVersion})`,
+		);
+	}
+
+	return await api.patch({
```
Page-, line 635: I think this can be handled by adding
```ts
request.values.should_be_managed_by__supervisor_release = supervisorRelease[0].id
```
and adding it to the current patch that's running

src/hooks/resources/device.ts
xginn8: When @CameronDiver and I discussed, he mentioned he wanted to handle this at the app level (i.e. as a balenaCloud feature), unless I misunderstood

CameronDiver: Ok so there was a little confusion here, and we've reconciled it. We're not going to throw an error here, and instead just bail on trying to set the `should_be_managed_by__supervisor_release` field.

dfunckt: Yeah, I too think that's the only option really, but we do need to consider soon how openBalena is going to have the Supervisor app and releases populated.

src/hooks/resources/device.ts
xginn8: How can I add this bit to the current patch? Is that something better suited for a `POSTPARSE` hook to reuse the SQL compile step?

xginn8:
This change as-is is insufficient for the same reason that's listed here: https://github.com/balena-io/open-balena-api/pull/359#issuecomment-630822862 (the existence of non-canonical "target versions"). Discussing elsewhere, shall report back

src/hooks/resources/device.ts
```diff
@@ -576,3 +589,59 @@ async function checkSupervisorReleaseUpgrades(
 		}
 	}
 }
+
+async function setSupervisorReleaseResource(
+	api: sbvrUtils.PinejsClient,
+	deviceIds: number[],
```
richbayliss, line 595: Since this Fn only handles the first device ID in this list, I would like it to either;
- take a single ID, or
- set the value for all the IDs

test/test-lib/fake-device.ts
```diff
@@ -70,11 +70,17 @@ export async function provisionDevice(
...
 		.expect(201);
 
+	await supertest(app, admin)
+		.patch(`/resin/device(${device.id})`)
+		.send({
+			os_version: '2.38.0+rev1',
+			supervisor_version: '5.0.1',
+		})
+		.expect(200);
+
```
richbayliss, line 83: I thought the aim of the PR was to allow the supervisor release to be set when the device `PATCH`es the state endpoint, and subsequently patches the device resource? This step here doesn't seem to make sense since this action isn't performed by the supervisor directly... what am I missing?

xginn8: We discussed this internally. tl;dr is that only the new test case will PATCH in this way (but the code will be removed from the initial POST). additionally we will explicitly test a PATCH to the state endpoint.

src/hooks/resources/device.ts
```diff
@@ -576,3 +589,59 @@ async function checkSupervisorReleaseUpgrades(
...
+							},
+						},
+					},
+				},
+			},
+		},
+	})) as AnyObject[];
+
+	if (supervisorRelease != null) {
+		return await api.patch({
```
Page-, line 634: The return + await is unnecessary here:
```suggestion
		return api.patch({
```
or
```suggestion
		await api.patch({
```

src/hooks/resources/device.ts
```diff
@@ -576,3 +589,59 @@ async function checkSupervisorReleaseUpgrades(
...
+	const device = (await api.get({
+		resource: 'device',
+		id: deviceIds[0],
+		options: {
+			$expand: ['is_of__device_type'],
+			$select: ['id', 'is_of__device_type', 'supervisor_version'],
+		},
+	})) as AnyObject;
+
+	if (device.supervisor_version != null) {
```
Page-, line 608: Can you put this into a `$filter` above please

src/hooks/resources/device.ts
```diff
@@ -576,3 +589,59 @@ async function checkSupervisorReleaseUpgrades(
...
+	api: sbvrUtils.PinejsClient,
+	deviceIds: number[],
+	supervisorVersion: string,
+) {
+	// This function _should_ get called with one device (since each supervisor reports itself)
+	const device = (await api.get({
+		resource: 'device',
+		id: deviceIds[0],
+		options: {
+			$expand: ['is_of__device_type'],
```
Page-, line 603: Can you add a $select for `is_of__device_type` please

src/hooks/resources/device.ts
Page-: 👍

src/hooks/resources/device.ts
xginn8: I'm not sure i follow, I do `$select` that in line 604: https://github.com/balena-io/open-balena-api/pull/356/files#diff-2a82ae8626a83fe22f032b4e504317d5R604

src/hooks/resources/device.ts
xginn8: done

src/hooks/resources/device.ts
xginn8: done

src/hooks/resources/device.ts
xginn8: done

src/hooks/resources/device.ts
xginn8: I'm not sure I follow this, I believe it is selected in the following line?

CameronDiver: If you expand here, its going to pull the entire device type object, whereas doing
```
$expand: {
  is_of__device_type: {
    $select: 'id'
  }
}
```
it won't

CameronDiver: I also don't know what that select at the bottom is doing, I don't think you need the `is_of__device_type` there, but @Page- can confirm.

Page-: @CameronDiver it's redundant as anything you expand is naturally also selected, but having it won't cause an issue

xginn8: Thanks for explaining that! I didn't realize i could `$select` inside my `$expand`. Updated.

src/lib/auth.ts
```diff
@@ -66,7 +66,7 @@ export const DEVICE_API_KEY_PERMISSIONS = [
 	'resin.device_type.read?describes__device/canAccess()',
 	`resin.device.read?${matchesActor}`,
 	`resin.device.update?${matchesActor}`,
-	'resin.supervisor_release.read?should_manage__device/canAccess()',
+	'resin.supervisor_release.read?is_for__device_type/any(dt:dt/describes__device/canAccess())',
```
xginn8, line 69: this _might_ not be required (perhaps I misunderstand the type of token created/used as part of the fake-device initialization).

test/test-lib/fake-device.ts
xginn8: This is now implemented.

test/test-lib/fake-device.ts
```diff
@@ -96,5 +96,25 @@ export async function provisionDevice(
...
+				config: {},
+				apps: {},
+			},
+			dependent: {
+				apps: {},
+				devices: {},
+			},
+		};
+		await supertest(app, device)
+			.patch(`/device/v2/${device.uuid}/state?apiKey=${device.token}`)
```
Page-, line 113: I don't think the `apiKey` arg is necessary as `supertest(app, device)` should use the auth header already, and I think that's how the supervisor patches @CameronDiver ?
```suggestion
		await supertest(app, device)
			.patch(`/device/v2/${device.uuid}/state`)
```

src/lib/auth.ts
Page-: Yeah I don't understand the change for this, the device should only need to read the supervisor_release that it's actually meant to be running, in this case a device can read all supervisor releases for its device type including ones that are marked as `is public = false`

src/hooks/resources/device.ts
```diff
@@ -576,3 +590,65 @@ async function checkSupervisorReleaseUpgrades(
...
+							},
+						},
+					},
+				},
+			},
+		},
+	})) as AnyObject[];
+
+	if (supervisorRelease.length !== 0) {
+		await api.patch({
```
Page-, line 641: I'd go with
```suggestion
	if (supervisorRelease.length === 0) {
		return;
	}
	await api.patch({
```
to match the short-circuiting style done previously and to stay consistent - it also makes it clear that if it's 0 then we do nothing and return whereas in the current version there could technically be code after the `if`

src/hooks/resources/device.ts
```diff
@@ -354,6 +354,20 @@ sbvrUtils.addPureHook('PATCH', 'resin', 'device', {
...
+			// When a device checks in with it's initial supervisor version, set the corresponding supervisor_release
+			// resource using its current version
+			waitPromises.push(
+				getCurrentRequestAffectedIds(args).map((id) =>
+					setSupervisorReleaseResource(
+						api,
+						id,
+						request.values.supervisor_version,
+					),
+				),
```
Page-, line 367: This will work but I generally prefer to pass all of the ids in so that the api requests can work on the whole batch at once and have `O(1)` requests rather than `O(n)`

src/lib/auth.ts
xginn8: This PR addresses the case where we're not actually managed by any `supervisor_release`, so i think we'll need to access all public editions of compatible supervisors for our device-type?

src/hooks/resources/device.ts
xginn8: I wish you had mentioned that in https://github.com/balena-io/open-balena-api/pull/356#discussion_r433220642. I'm happy to update it if it's preferable, but I'm unclear how this could ever be more than one device at a time unless we start queuing/batching ingesting updates from supervisors?

src/lib/auth.ts
Page-: I think it makes more sense to do the `.get` itself with higher privileges rather than grant the device key these privileges as imo minimizing the power of a device key is a good approach. It's also possible that if the device at any point does `(GET /supervisor_release)[0]` that this would be a breaking change - something it definitely shouldn't rely on but I would not be shocked if it does

src/hooks/resources/device.ts
Page-: Sorry, when I read only take a single id I saw it as the caller only allowing it to be a single id - eg if it was more it could bail if that should be blocked. Specifically the way I could see a batch being updated would be if it was either a batch of dependent devices being updated or a user doing the update for some reason

xginn8: Just so I understand, dependent devices don't have their own supervisors, correct? This hook will only run once per device, unless the current supervisor has not been properly released and there are no matches (in which case this hook will run at each state endpoint PATCH).

I will rewrite to batch the PATCHes.

xginn8:
Related to https://github.com/balena-io/open-balena-api/issues/321

src/hooks/resources/device.ts
xginn8: tl;dr discussed offline, we decided we'll do the GET in one batch and patch device-type by device-type since there's no way to do an inner join in Pine.js.

src/lib/auth.ts
xginn8: I have raised the privileges for the `get` and `patch` accordingly, rather than elevating the permissions of the key itself.

xginn8:
@Page- PTAL. The only outstanding question I am left with is how to deal with the possibility of multiple device-types for a single device. I suppose since the architectures must be the same it's an inconsequential issue, but we do release a supervisor per device type at this point (that will go away once the supervisor becomes a proper balena app).

src/hooks/resources/device.ts
```diff
@@ -579,3 +595,105 @@ async function checkSupervisorReleaseUpgrades(
...
+		return;
+	}
+
+	await Bluebird.map(deviceTypes, async (deviceType) => {
+		const affectedDeviceIds = _.map(devices, (d) => {
+			// TODO: hardcoded index
+			if (d.is_of__device_type[0].id === deviceType) {
+				return d.id;
+			}
+		});
```
Page-, line 651: Take a look at [_.groupBy](https://lodash.com/docs/4.17.15#groupBy) - you should be able to handle this and the `uniqBy` above in one go with
```ts
const devicesByDeviceType = _.groupBy(device, (d) => d.is_of__device_type[0].id)
```
and get a `{ [deviceTypeId: number]: device[] }` mapping back

src/hooks/resources/device.ts
```diff
@@ -579,3 +595,105 @@ async function checkSupervisorReleaseUpgrades(
...
+		options: {
+			$expand: {
+				is_of__device_type: {
+					$select: 'id',
+				},
+			},
+			// if the device already has a supervisor_version, just bail.
+			$filter: {
+				id: { $in: deviceIds },
+				supervisor_version: { $eq: null },
```
Page-, line 618: `$eq` is the default so you can simplify this to
```suggestion
				supervisor_version: null,
```

src/hooks/resources/device.ts
```diff
@@ -579,3 +595,105 @@ async function checkSupervisorReleaseUpgrades(
...
+
+		await rootApi.patch({
+			resource: 'device',
+			options: {
+				$filter: {
+					id: { $in: affectedDeviceIds },
+				},
+			},
+			body: {
+				// TODO: hardcoded index
```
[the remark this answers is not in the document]
xginn8: Discussed offline, since this is defined in SBVR as `Necessity: each device is of exactly one device type.` (https://github.com/balena-io/open-balena-api/blob/master/src/balena.sbvr#L433), we can comfortably ignore this (or further clean it up by using the `__id` selector).

src/hooks/resources/device.ts
xginn8: this be much cleaner, thanks

src/hooks/resources/device.ts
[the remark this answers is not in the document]
xginn8: included in latest commit.

src/hooks/resources/device.ts
```diff
@@ -578,3 +594,85 @@ async function checkSupervisorReleaseUpgrades(
...
+
+	if (devices.length === 0) {
+		return;
+	}
+
+	const devicesByDeviceType = _.groupBy(devices, (d) => {
+		return d.is_of__device_type.__id;
+	});
+
+	if (Object.keys(devicesByDeviceType).length === 0) {
```
Page-, line 626: If you want you could assign this and reuse it for the map like
```ts
const deviceTypes = Object.keys(devicesByDeviceType)
if (deviceTypes.length === 0) {
...
await Promise.all(
	deviceTypes.map(async (deviceType) => {
		const affectedDeviceIds = devicesByDeviceType[deviceType].map((d) => d.id);
```
it's not necessary though

PR #357, opened 2020-05-14
Update dependencies

Page-:
Typescript 3.9 fixes the `Promise.all` typings so we can convert the `Bluebird.all` uses finally :)


Page-:
@balena-ci retest

Page-:
@balena-ci retest

PR #359, opened 2020-05-19
supervisor release: set target state to current state if already set

xginn8:
HQ: https://github.com/balena-io/balena/pull/2177
Change-type: patch


xginn8:
I turned this back to a draft as it'll need to be `_logstream{,2}` aware.

xginn8:
This actually is not required since https://github.com/balena-io/open-balena-api/pull/356 will take care of this via hook.

PR #360, opened 2020-05-21
Abstract out the throttled iteration of requestDevices

Page-:
Change-type: patch

### DIFF
--- a/src/lib/device-proxy.ts
+++ b/src/lib/device-proxy.ts
@@ -1,4 +1,3 @@
-import * as Bluebird from 'bluebird';
 import type { Request, Response } from 'express';
 import * as _ from 'lodash';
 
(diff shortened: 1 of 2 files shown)


src/lib/device-proxy.ts
```diff
@@ -136,6 +136,25 @@ export const proxy = async (req: Request, res: Response) => {
 	}
 };
 
+export async function throttledForEach<T, U>(
```
dfunckt, line 139: I'd rather this was exported from a `utils` module and received the delay to apply as an argument.

Page-: Urg, I did mean to make the delay configurable - I'll get that updated

Page-: done

src/lib/utils.ts
```diff
@@ -56,3 +57,23 @@ export const varListInsert = (varList: EnvVarList, obj: Dictionary<string>) => {
...
+): Promise<U[]> => {
+	const promises: Array<PromiseLike<U>> = [];
+	for (const item of array) {
+		// We do not wait for each individual fn, we just throttle the calling of them
+		promises.push(fn(item));
+		// We add a delay between each notification so that we do not in essence
+		// trigger a DDOS from balena devices against us, but we do not wait for
+		// completion of individual requests because doing so could cause a
+		// terrible UX if we have a device time out, as that would block all the
+		// subsequent notifications
```
dfunckt, line 74: This comment is better placed IMO at the call site in device-proxy.ts.

PR #361, opened 2020-05-21
permissions: allow device key to read hostapp if it can access DT

sradevski:
Change-type: minor

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -63,11 +63,11 @@ export const ROLES: {
 };
 
 export const DEVICE_API_KEY_PERMISSIONS = [
-	'resin.device_type.read?describes_device/canAccess()',
+	'resin.device_type.read?describes__device/canAccess()',
 	`resin.device.read?${matchesActor}`,
 	`resin.device.update?${matchesActor}`,
 	'resin.supervisor_release.read?should_manage__device/canAccess()',
-	'resin.application.read?owns__device/canAccess() or depends_on__application/any(a:a/owns__device/canAccess())',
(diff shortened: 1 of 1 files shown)


src/lib/auth.ts
```diff
@@ -63,11 +63,11 @@ export const ROLES: {
 };
 
 export const DEVICE_API_KEY_PERMISSIONS = [
-	'resin.device_type.read?describes_device/canAccess()',
+	'resin.device_type.read?describes__device/canAccess()',
```
sradevski, line 66: It took me a while until I realized that this was broken...

src/lib/auth.ts
```diff
@@ -63,11 +63,11 @@ export const ROLES: {
...
 	`resin.device.read?${matchesActor}`,
 	`resin.device.update?${matchesActor}`,
 	'resin.supervisor_release.read?should_manage__device/canAccess()',
-	'resin.application.read?owns__device/canAccess() or depends_on__application/any(a:a/owns__device/canAccess())',
+	'resin.application.read?owns__device/canAccess() or depends_on__application/any(a:a/owns__device/canAccess()) or (is_host eq true and is_for__device_type/canAccess())',
```
sradevski, line 70: The tests fail with this, but pass with `resin.application.read?owns__device/canAccess() or depends_on__application/any(a:a/owns__device/canAccess()) or (is_host eq true and is_for__device_type/any(dt:dt/describes__device/canAccess()))`, why is that the case?

sradevski:
@balena-ci rebase

Issue #366, opened 2020-06-04
Should have a mac_address field for devices


thgreasi:
HQ: https://github.com/balena-io/balena/issues/2195

PR #367, opened 2020-06-04
Add a mac_address field to devices

thgreasi:
Resolves: #366
HQ: https://github.com/balena-io/balena/issues/2195
Change-type: minor

### DIFF
--- a/test/03-device-state-v2.ts
+++ b/test/03-device-state-v2.ts
@@ -292,3 +292,59 @@ describe('Device State v2', () => {
 		});
 	});
 });
+
+describe('Device State v2 patch', function () {
+	let fx: fixtures.Fixtures;
+	let admin: UserObjectParam;
+	let applicationId: number;
+	let device: fakeDevice.Device;
+
+	before(async () => {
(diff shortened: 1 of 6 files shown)


test/03-device-state-v2.ts
```diff
@@ -292,3 +292,62 @@ describe('Device State v2', () => {
...
+				provisioning_progress: null,
+				provisioning_state: null,
+				ip_address: '192.168.1.1',
+				mac_address: '00:11:22:33:44:55',
+				download_progress: null,
+				api_port: 48484,
+			},
+		};
+
+		await supertest(app, device)
```
richbayliss, line 334: I think this would be a good opertunity to mock out a state `PATCH` in the `fake-device` class... https://github.com/balena-io/open-balena-api/blob/master/test/test-lib/fake-device.ts#L87

This would make the behaviour repeatable and consistent 👍

thgreasi:
I remember someone mentioning that this was discussed at some point but I've no extra info about the more abstract approach. Maybe @CameronDiver knows more and we can move a separate "DeviceInfo" kind of resource at a later point.

PR #372, opened 2020-06-09
Update dependencies

Page-:
Update pinejs from 12.3.10 to 12.3.13

Change-type: patch


package.json
```diff
@@ -109,5 +109,13 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
+  },
+  "mocha": {
```
thgreasi, line 113: How about moving this to a `.mocharc.json` to keep the package.json smaller?
See: https://mochajs.org/#configuring-mocha-nodejs

thgreasi: I will PR that :fist_raised:

thgreasi: See: https://github.com/balena-io/open-balena-api/pull/374

PR #373, opened 2020-06-12
Move iterating of devices to create service installs to a lower level

Page-:
This makes it easier to optimize away the additional requests in future

It also avoids issues when `affectedIds` stops being a bluebird promise


Page-:
I'd recommend hiding whitespace changes btw, a lot of the change is just due to indentation

Page-:
@balena-ci rebase

Page-:
@balena-ci rebase

PR #374, opened 2020-06-12
Use a mocharc and a huskyrc to keep the package.json cleaner

thgreasi:
Subjectively I prefer having a separate single purpose
file rather than a larger package.json.
This also avoids unnecessary extra lines on the
published package.json.
See: https://unpkg.com/browse/@balena/open-balena-api@0.78.4/package.json

Change-type: patch


thgreasi:
:(

PR #376, opened 2020-06-15
Add remaining indexes for foreign keys

Page-:
Also add explanation for some indexes that are for specific queries

Change-type: patch

### DIFF
--- a/src/balena-init.sql
+++ b/src/balena-init.sql
@@ -12,86 +12,128 @@ ADD COLUMN IF NOT EXISTS "jwt secret" VARCHAR(255) NULL;
 ALTER TABLE "device"
 ALTER COLUMN "api heartbeat state" SET DEFAULT 'unknown';
 
-CREATE INDEX IF NOT EXISTS "release_application_commit_status_idx"
-ON "release" ("belongs to-application", "commit", "status");
-
+-------------------------------
+-- Start foreign key indexes --
+-------------------------------
 CREATE INDEX IF NOT EXISTS "api_key_actor_idx"
 ON "api key" ("is of-actor");
 
-CREATE INDEX IF NOT EXISTS "user_actor_idx"
-ON "user" ("actor");
+CREATE INDEX IF NOT EXISTS "application_actor_idx"
+ON "application" ("actor");
+CREATE INDEX IF NOT EXISTS "application_depends_on_application_idx"
+ON "application" ("depends on-application");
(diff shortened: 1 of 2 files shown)


src/migrations/00029-add-fk-indexes.sql
```diff
@@ -0,0 +1,15 @@
...
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "device_is_running_release_idx"
+ON "device" ("is running-release");
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "device_should_be_running_release_idx"
+ON "device" ("should be running-release");
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "gateway_download_device_idx"
+ON "gateway download" ("is downloaded by-device");
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "release_tag_release_idx"
+ON "release tag" ("release");
```
thgreasi, line 15: I guess that this provides a measurable speed improvement on top of the composite release-tag_key unique index?

src/balena-init.sql
```diff
@@ -12,86 +12,129 @@ ADD COLUMN IF NOT EXISTS "jwt secret" VARCHAR(255) NULL;
...
+-- "service environment variable"."service" is the first part of an automated unique index
+
+-- "service label"."service" is the first part of an automated unique index
+
+-- "service install"."device" is the first part of an automated unique index
+CREATE INDEX IF NOT EXISTS "service_install_service_idx"
+ON "service install" ("installs-service");
+
+CREATE INDEX IF NOT EXISTS "supervisor_release_device_type_idx"
+ON "supervisor release" ("is for-device type");
```
thgreasi, line 110: Any reason these two are not in the migration?
For sure they are not in my BoB's DB.

src/migrations/00029-add-fk-indexes.sql
Page-: Nope, this is a mistake - fixed it now

src/balena-init.sql
Page-: I must have missed them when comparing what was moved vs new - fixed

thgreasi:
Sad story:
I remember a few months ago while testing my orgs ui branch that I created a git branch with device_is_running_release_idx, device_should_be_running_release_idx which I later at some point deleted thinking that this was already merged...

PR #380, opened 2020-06-17
Update pinejs and use updated imports

Page-:
Update pinejs from 12.3.14 to 12.4.0

Change-type: patch

### DIFF
--- a/src/lib/device-types/index.ts
+++ b/src/lib/device-types/index.ts
@@ -3,8 +3,7 @@ import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
 
 import * as deviceTypesLib from '@resin.io/device-types';
-import { sbvrUtils } from '@resin/pinejs';
-import type { Tx } from '@resin/pinejs/out/database-layer/db';
+import { sbvrUtils, permissions, errors } from '@resin/pinejs';
 import * as semver from 'balena-semver';
 import type { PinejsClientCoreFactory } from 'pinejs-client-core';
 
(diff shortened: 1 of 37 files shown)


src/lib/device-types/index.ts
```diff
@@ -18,8 +17,9 @@ import {
 } from './build-info-facade';
 import { getImageKey, IMAGE_STORAGE_PREFIX, listFolders } from './storage';
 
-const { InternalRequestError, root, api } = sbvrUtils;
-export const { BadRequestError, NotFoundError } = sbvrUtils;
+const { api } = sbvrUtils;
+const { InternalRequestError } = errors;
+export const { BadRequestError, NotFoundError } = errors;
```
dfunckt, line 22: Can you please find where these are imported and used and replace them with Pine imports?

Page-: They're here in order to avoid errors when generating the type declarations:
```
src/lib/device-types/index.ts:26:45 - error TS4020: 'extends' clause of exported class 'InvalidDeviceTypeError' has or is using private name 'BadRequestError'.

26 export class InvalidDeviceTypeError extends BadRequestError {}
                                               ~~~~~~~~~~~~~~~

src/lib/device-types/index.ts:28:45 - error TS4020: 'extends' clause of exported class 'UnknownDeviceTypeError' has or is using private name 'NotFoundError'.

28 export class UnknownDeviceTypeError extends NotFoundError {
                                               ~~~~~~~~~~~~~

src/lib/device-types/index.ts:34:42 - error TS4020: 'extends' clause of exported class 'UnknownVersionError' has or is using private name 'NotFoundError'.

34 export class UnknownVersionError extends NotFoundError {
```
but I've found it's possible to re-export them as purely types so I've switched to that method

dfunckt: Ugh, I see. That's a lot better though, thanks.

PR #381, opened 2020-06-18
Add “organization” resource

dfunckt:
Organizations sit one level higher than applications and users, and scope access to resources. Applications belong to organizations and users are members of organizations, which grants users access to those applications.

Because the core API is single-user, organizations offer nothing new (the admin user has full access to every resource unconditionally) but allow the core and cloud APIs to expose a uniform interface to clients, which is the motivation of this change.

People familiar with the cloud model will observe that I’ve included the bare minimum of resources that would let clients:

- query which organizations exist
- query which users are members of an organization
- query which organizations a user is a member of

You’ll also observe that I’ve not updated user authorization to go through org membership. We’d still have to grant all permissions unconditionally and it would make things explicit, but the permissions would become much more complicated adding both maintenance and performance ovehead and we’d still not be able to reuse the permissions on the cloud API. I don’t think the tradeoff is worth it and we can easily change that in the future if there’s a solid need.

I’ve also omitted DELETE hooks — the single org and org mebership are not expected to (and should not) be deleted on instances of the core API.

Change-type: minor

### DIFF
--- a/src/balena-init.sql
+++ b/src/balena-init.sql
@@ -24,6 +24,8 @@ CREATE INDEX IF NOT EXISTS "application_depends_on_application_idx"
 ON "application" ("depends on-application");
 CREATE INDEX IF NOT EXISTS "application_device_type_idx"
 ON "application" ("is for-device type");
+CREATE INDEX IF NOT EXISTS "application_organization_idx"
+ON "application" ("organization");
 CREATE INDEX IF NOT EXISTS "application_release_idx"
 ON "application" ("should be running-release");
 CREATE INDEX IF NOT EXISTS "application_application_type_idx"
@@ -88,6 +90,10 @@ CREATE INDEX IF NOT EXISTS "ipr_ipr_image_idx"
 ON "image-is part of-release" ("is part of-release", "image");
 
 -- "image label"."release image" is the first part of an automated unique index
+
+-- "organization membership"."user" is the first part of an automated unique index
+CREATE INDEX IF NOT EXISTS "organization_membership_is_member_of_organization_idx"
+ON "organization membership" ("is member of-organization");
 
 -- TODO: Check what the extra columns are optimizing for
 CREATE INDEX IF NOT EXISTS "release_application_commit_status_idx"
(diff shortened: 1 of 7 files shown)


src/balena-init.sql
```diff
@@ -18,6 +18,8 @@ ALTER COLUMN "api heartbeat state" SET DEFAULT 'unknown';
 CREATE INDEX IF NOT EXISTS "api_key_actor_idx"
 ON "api key" ("is of-actor");
 
+CREATE INDEX IF NOT EXISTS "application_organization_idx"
+ON "application" ("organization");
```
Page-, line 22: NIT: Could you put this below the actor index please, it feels weird not to have actor be first in the list due to its specialness

test/test-lib/init-tests.ts
```diff
@@ -37,19 +37,24 @@ const getAdminUser = async () => {
...
-	adminUser.token = token;
-
-	adminUser.actor = await supertest(app, adminUser)
-		.get(`/resin/user(${adminUser.id})`)
+	const user = (await balenaToken.parse(token)) as UserObjectParam;
+	user.token = token;
+	user.actor = await supertest(app, user)
+		.get(`/resin/user(${user.id})`)
 		.expect(200)
 		.then((res) => res.body.d[0].actor as number);
```
Page-, line 45:
```suggestion
	user.actor = (
			await supertest(app, user)
			.get(`/resin/user(${user.id})`)
			.expect(200)
		).body.d[0].actor as number;
```

test/test-lib/init-tests.ts
```diff
@@ -37,19 +37,24 @@ const getAdminUser = async () => {
...
 
-	return adminUser;
+	const org = await supertest(app, user)
+		.get(`/resin/organization?$select=id,name,handle&$filter=handle eq 'admin'`)
+		.expect(200)
+		.then((res) => res.body.d[0]);
```
Page-, line 50:
```suggestion
	const org = (
			await supertest(app, user)
			.get(`/resin/organization?$select=id,name,handle&$filter=handle eq 'admin'`)
			.expect(200)
		).body.d[0] as number;
```

PR #384, opened 2020-06-19
Avoid promisifyAll for redis due to performance

Page-:
This also ends up with slightly more accurate automatic typings, the only problem is that for `set` we use an overloaded form which cannot be fully inferred


tsconfig.build.json
```diff
@@ -12,6 +12,7 @@
 		"removeComments": true,
 		"sourceMap": true,
 		"strictNullChecks": true,
+		"strictBindCallApply": true,
```
Page-, line 15: This means the `.bind(redis)` returns an actual type instead of just `any` which is what you get with this option off

Page-:
@balena-ci retest

PR #391, opened 2020-06-30
allow service instances to specify IP on registration

wrboyce:
Change-type: minor

### DIFF
--- a/src/hooks/resources/service_instance.ts
+++ b/src/hooks/resources/service_instance.ts
@@ -8,7 +8,7 @@ import { getIP } from '../../lib/utils';
 sbvrUtils.addPureHook('POST', 'resin', 'service_instance', {
 	POSTPARSE: ({ request, req }) => {
 		request.values.service_type = getServiceFromRequest(req);
-		request.values.ip_address = getIP(req as Request);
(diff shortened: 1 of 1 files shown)


src/hooks/resources/service_instance.ts
```diff
@@ -8,7 +8,7 @@ import { getIP } from '../../lib/utils';
 sbvrUtils.addPureHook('POST', 'resin', 'service_instance', {
 	POSTPARSE: ({ request, req }) => {
 		request.values.service_type = getServiceFromRequest(req);
-		request.values.ip_address = getIP(req as Request);
+		request.values.ip_address = req?.body?.ip_address ?? getIP(req as Request);
```
Page-, line 11: I'm pretty sure most of these `?` are unnecessary, at a minimum
```suggestion
		request.values.ip_address = req.body?.ip_address ?? getIP(req as Request);
```
but I think
```suggestion
		request.values.ip_address = req.body.ip_address ?? getIP(req as Request);
```
would also work

wrboyce: Yeah, you're right… I think I got a bit carried away ha.

wrboyce:
@Page- yeah, we discussed it (albeit briefly) around the time we switched to k8s.

PR #392, opened 2020-07-02
supervisor-release: drop entire resource & references

xginn8:
we will move away from a dedicated resource in favor of native releases

HQ: https://github.com/balena-io/balena-io/pull/2177
Change-type: minor


src/migrations/00031-drop-supervisor-release.sql
```diff
@@ -0,0 +1,7 @@
+DROP TABLE "supervisor release";
```
Page-, line 1: This needs to be the last bit, after all the things that depend on it have been dropped

xginn8: updated.

xginn8:
this PR is pending an arch call

Page-:
@balena-ci republish

thgreasi:
@xginn8 was this good to be merged?
The UI still has references to that resource & the respective fields of the device resource.
See: https://github.com/balena-io/balena-ui/search?q=should_be_managed_by__supervisor_release+OR+supervisor_release&unscoped_q=should_be_managed_by__supervisor_release+OR+supervisor_release

xginn8:
@thgreasi thanks for checking -- we are moving the resource back to the closed api for the time being, see https://github.com/balena-io/balena-api/pull/2662.

thgreasi:
Oh I missed that. Thanks for clarifying :+1:

PR #397, opened 2020-07-16
Add api key tests

thgreasi:
Extracted from the cloud API, converted to async await.

Change-type: patch


thgreasi:
Prepares the ground for https://github.com/balena-io/balena-io/pull/2226

Issue #393, opened 2020-07-07
Should provide a generic endpoint to generate ApiKeys scoped on specific roles


thgreasi:
Moreover, when the user looses a role that allowed them to create an ApiKey with a specific scope (role), then such ApiKeys should get invalidated.
HQ: https://github.com/balena-io/balena-io/pull/2226

PR #398, opened 2020-07-16
Add a generic endpoint to generate ApiKeys scoped on specific roles

thgreasi:
On top of #397 so let's get that in first and ignore that commit while checking this.

Connects-to: #393
Change-type: minor
HQ: https://github.com/balena-io/balena-io/pull/2226
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/bFduRFrd1wEUZmDxFc5FHlGNcxl

### DIFF
--- a/src/lib/api-keys.ts
+++ b/src/lib/api-keys.ts
@@ -1,6 +1,78 @@
+import type { Request } from 'express';
 import * as _ from 'lodash';
+import { errors } from '@balena/pinejs';
+import {
+	createApiKey,
+	ApiKeyOptions,
+	PartialCreateKey,
+} from '../platform/api-keys';
 
-import { createApiKey, PartialCreateKey } from '../platform/api-keys';
+const { BadRequestError } = errors;
+
+const supportedActorTypes = ['application', 'user', 'device'] as const;
+
+export type ApiKeyParameters = {
+	actorType: typeof supportedActorTypes[number];
+	actorTypeId: number;
+	roles: string[];
+} & Pick<ApiKeyOptions, 'name' | 'description' | 'apiKey'>;
+
+export const createGenericApiKey = async (
+	req: Request,
+	{
+		actorType,
+		actorTypeId,
+		roles,
+		name,
+		description,
+		apiKey,
+	}: ApiKeyParameters,
+) => {
+	if (!supportedActorTypes.includes(actorType)) {
+		throw new BadRequestError('Unsupported actor type');
+	}
+
+	if (!Number.isFinite(actorTypeId)) {
+		throw new BadRequestError('Actor type id must be a number');
+	}
+
+	if (
+		!Array.isArray(roles) ||
+		roles.length === 0 ||
+		roles.some((r) => typeof r !== 'string' || r.length === 0)
+	) {
+		throw new BadRequestError('Roles should be an array of role names');
+	}
+
+	if (roles.length !== 1) {
+		throw new BadRequestError('API Keys currently only support a single role');
+	}
+
+	if (!name) {
+		const namedRole = roles.find((r) => r.startsWith('named-'));
+
+		if (namedRole != null) {
+			throw new BadRequestError(
+				`API keys with the '${namedRole}' role require a name`,
+			);
+		}
+	}
+
+	const roleName = roles[0];
+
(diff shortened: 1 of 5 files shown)


src/lib/api-keys.ts
```diff
@@ -1,6 +1,69 @@
...
+} from '../platform/api-keys';
+
+const { BadRequestError } = errors;
+
+const supportedActorTypes = ['application', 'user', 'device'] as const;
+
+export type ApiKeyParameters = {
+	actorType: typeof supportedActorTypes[number];
+	actor: number;
+	roles: string | [string];
```
Page-, line 17: Any reason we allow both forms?

thgreasi: Not really, will update it :+1:

src/lib/api-keys.ts
```diff
@@ -1,6 +1,74 @@
...
+	if (!supportedActorTypes.includes(actorType)) {
+		throw new BadRequestError('Unsupported actor type');
+	}
+
+	if (!Number.isFinite(actorTypeId)) {
+		throw new BadRequestError('Actor must be a number');
+	}
+
+	if (!Array.isArray(roles) || roles.some((r) => typeof r !== 'string' || !r)) {
+		throw new BadRequestError('Roles should be an array of non-empty strings');
```
Page-, line 40: This is probably more user friendly if written as
```suggestion
		throw new BadRequestError('Roles should be an array of role names');
```
It provides the same info since a role name is a string and is also non-empty whilst also providing the context of what the string represents

src/lib/api-keys.ts
```diff
@@ -1,6 +1,74 @@
...
+		description,
+		apiKey,
+	}: ApiKeyParameters,
+) => {
+	if (!supportedActorTypes.includes(actorType)) {
+		throw new BadRequestError('Unsupported actor type');
+	}
+
+	if (!Number.isFinite(actorTypeId)) {
+		throw new BadRequestError('Actor must be a number');
```
Page-, line 36: This is not an actor but rather an "actor type id", eg a user id or device id, which makes the error message fairly unclear. I think it'd be better as:
```suggestion
		throw new BadRequestError('Actor type id must be a number');
```
although maybe it could be better still?

src/lib/api-keys.ts
```diff
@@ -1,6 +1,74 @@
...
+	if (roles.length !== 1) {
+		throw new BadRequestError('API Keys currently only support a single role');
+	}
+
+	if (!name) {
+		const namedRole = roles.find((r) => r.startsWith('named-'));
+
+		if (namedRole != null) {
+			throw new BadRequestError(
+				`API keys with the ${namedRole} role require a name`,
```
Page-, line 52: I like to quote the variable bit in case it includes spaces or something, ie
```suggestion
				`API keys with the '${namedRole}' role require a name`,
```

src/lib/api-keys.ts
```diff
@@ -1,6 +1,74 @@
...
+) => {
+	if (!supportedActorTypes.includes(actorType)) {
+		throw new BadRequestError('Unsupported actor type');
+	}
+
+	if (!Number.isFinite(actorTypeId)) {
+		throw new BadRequestError('Actor must be a number');
+	}
+
+	if (!Array.isArray(roles) || roles.some((r) => typeof r !== 'string' || !r)) {
```
Page-, line 39: If you change this to `every` then it will also fail for an empty array, eg
```suggestion
	if (!Array.isArray(roles) || roles.every((r) => typeof r === 'string' && r.length > 0)) {
```
see:
```
[].every(() => false) // true

src/lib/api-keys.ts
thgreasi: Loved it :heart:

src/lib/api-keys.ts
thgreasi: I guess you missed the `!` in front of the `every`.

Thought of that but it felt that `some` is cleaner than `!every` (the `!` makes then equivalent, despite their 0 length difference in the general case).
PS: In my head `some` sounds like the early return pattern, and that's probably why I prefer it.

Updated a bit, but let me know if you still prefer `!every`.

PR #400, opened 2020-07-21
Update pinejs, adding support for named keys

Page-:
Update pinejs from 13.2.1 to 13.3.0

Change-type: patch


src/hooks/resources/user.ts
```diff
@@ -40,9 +40,12 @@ sbvrUtils.addPureHook('DELETE', 'resin', 'user', {
 		if (userIdBind == null) {
 			throw new BadRequestError('You must provide user ID');
 		}
+		if (!('bind' in userIdBind)) {
```
Page-, line 43: These checks can be changed to use `isBindReference` from https://github.com/balena-io-modules/odata-to-abstract-sql/pull/66 to make it nicer but it's not a blocker

test/test-lib/fake-device.ts
```diff
@@ -71,9 +71,7 @@ export async function provisionDevice(
 		.expect(201);
 
 	const { body: provisionedDevice } = await supertest(app, admin)
-		.get(
-			`/resin/device?$filter=uuid eq '${deviceUuid}'&$select=supervisor_version`,
-		)
+		.get(`/resin/device(uuid='${deviceUuid}')?$select=supervisor_version`)
 		.expect(200);
 	expect(provisionedDevice.d[0].supervisor_version).to.equal(supervisorVersion);
```
dfunckt, line 76: This type of query still returns an array of results, right? It is slightly confusing if so, but I didn't actually check the OData spec.

Page-: At least with pinejs it does, and it'd be a very invasive breaking change to change that. As for the odata spec I'm not actually sure what it says, and it probably changes based on version (eg I'm fairly sure the `d` property is no longer a thing with the latest version). What you're probably thinking of is `pinejsClient.get({ ..., id: x })` which automatically does `return result.d[0]` instead of `return result.d`

dfunckt: Yes exactly -- my thinking is that `/device(uuid=abcde1234)` is equivalent to `/device(1234)`, right? What's the point of having support for the new form if it is to return an array, which is identical to a filter?

Page-: Yes, and both `/device(uuid='abcde1234') and `/device(1234)` return exactly the same thing

dfunckt: I'm now confused :) Does the OData response *always* in the form `{ "d": [ <result1>, <result2>, ... ] }` and these two forms just happen to return zero or one results?

Page-: Yes, and pinejs-client-core as a sugar when specifying `id: x` will unwrap to just the single result and return `result | undefined`: https://github.com/balena-io-modules/pinejs-client-js/blob/v6.1.0/src/index.ts#L891-L898

dfunckt: Yeah makes complete sense -- I was confused thinking that sugar was on the API response itself, but it's a client-side thing.

PR #401, opened 2020-07-21
Update pinejs and make use of named keys

Page-:


src/routes/devices.ts
```diff
@@ -662,11 +662,13 @@ const upsertImageInstall = async (
 		}
 		await resinApi.patch({
 			resource: 'image_install',
+			id: {
+				installs__image: imageId,
+				device: deviceId,
```
thgreasi, line 667:
```suggestion
				device: deviceId,
				installs__image: imageId,
```
Reordered to match the sbvr :)

src/routes/devices.ts
```diff
@@ -705,16 +707,14 @@ const upsertGatewayDownload = async (
 	} else {
 		await resinApi.patch({
 			resource: 'gateway_download',
+			id: {
+				is_downloaded_by__device: deviceId,
+				image: imageId,
```
thgreasi, line 712:
```suggestion
				image: imageId,
				is_downloaded_by__device: deviceId,
```
Reorder to match the sbvr

src/routes/registry.ts
```diff
@@ -341,7 +343,7 @@ const $getSubject = memoize(
 							},
 						},
 					},
-				})) as AnyObject[];
+				})) as AnyObject;
 				if (device != null && device.id != null) {
```
thgreasi, line 347: Aside: Seems we don't actually need to check the `device.id` and can drop it.

src/routes/registry.ts
```diff
@@ -316,14 +316,16 @@ const $getSubject = memoize(
 		if (subject) {
 			try {
 				// Try to resolve as a device api key first, using the passed in subject
-				const [device] = (await api.resin.get({
+				const device = (await api.resin.get({
 					resource: 'device',
 					passthrough: { req: permissions.root },
+					id: {
+						// uuids are passed as `d_${uuid}`
+						uuid: subject.replace(/^d_/, ''),
```
thgreasi, line 324: Not against it, but wanted to say that having the natural key in the `id` as well as a `$filter` a bit lower felt a bit disconnected. 
Maybe it's just the part of me which thinks/translates $filter as a WHERE...

src/routes/devices.ts
```diff
@@ -626,16 +626,16 @@ const upsertImageInstall = async (
 	releaseId: number,
 	dlProg?: number,
 ): Promise<void> => {
-	const [imgInstall] = (await resinApi.get({
+	const imgInstall = (await resinApi.get({
 		resource: 'image_install',
+		id: {
+			installs__image: imageId,
```
thgreasi, line 632: BTW, will pine error if we specify an incomplete natural key or props that non-unique props in the `id` field?
Eg: just the `id: {device}`, or `id: {device, created_at}`

src/routes/registry.ts
Page-: It's not in the $filter? The only places it should still be in the $filter if it's something like `uuid: { $in: ... }` or `uuid: { $startswith: '...' }`

src/routes/devices.ts
Page-: Yes

thgreasi: Awesome?

PR #404, opened 2020-07-27
Avoid error in device state patching when local state is null

Page-:
Change-type: patch


Page-:
@balena-ci rebase

Page-:
@balena-ci I self-certify!

PR #405, opened 2020-07-29
Improve compatibility with async/await stack traces

Page-:
Update pinejs from 13.4.0 to 13.4.1

Change-type: patch


src/routes/devices.ts
```diff
@@ -185,19 +184,24 @@ export const receiveOnlineDependentDevices: RequestHandler = async (
...
-						belongs_to__application: dependent_app,
-						device_type: dependent_device_type,
-						local_id: localId,
-						logs_channel: randomstring.generate({ length: 62, charset: 'hex' }),
-					},
-					options: { returnResource: false },
+			await Promise.all(
+				toBeProvisioned.map(async (localId) => {
+					// Provision new dependent devices
+					await resinApiTx.post({
```
thgreasi, line 190: Do we really benefit that much by having the stacktrace point to the map in here?
I mean, it sounds like an extra thing to remember to do while coding new things.

Page-: It's probably not that big of a deal to miss it but the better the stack trace the easier a lot of things are to debug, so I leant towards too much rather than too little

PR #408, opened 2020-08-03
Export useful functions so that they are directly accessible on import

Page-:
Change-type: minor


Page-:
@balena-ci rebase

Page-:
@balena-ci retest

PR #411, opened 2020-08-04
Remove unused return value when resolving device type

sradevski:
Change-type: patch

### DIFF
--- a/src/hooks/common.ts
+++ b/src/hooks/common.ts
@@ -6,7 +6,7 @@ export const resolveDeviceType = async (
 	api: sbvrUtils.PinejsClient,
 	request: sbvrUtils.HookRequest,
 	fkValue: string,
-): Promise<deviceTypes.DeviceType> => {
+): Promise<void> => {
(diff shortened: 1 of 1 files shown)


src/hooks/common.ts
```diff
@@ -6,7 +6,7 @@ export const resolveDeviceType = async (
 	api: sbvrUtils.PinejsClient,
 	request: sbvrUtils.HookRequest,
 	fkValue: string,
-): Promise<deviceTypes.DeviceType> => {
+) => {
```
dfunckt, line 9: It's good to have explicit types on function definitions.

```suggestion
): Promise<void> => {
```

thgreasi:
@sradevski there is a `throw new UnknownDeviceTypeError(slug);` in the `findBySlug` call.
Don't we need to still `await findBySlug()` so that we also check against the dt cache?
If we did that change on purpose, since we are already checking in the db, then I'm fine with it.

sradevski:
@thgreasi that is what I thought that request is for, but it is enough to check against the DB, which should become the source of truth (and already is for existence)

PR #413, opened 2020-08-05
Avoid fetching unnecessary device info in /dependent/v1/scan

Page-:
Change-type: patch


Page-:
@balena-ci rebase

Page-:
@balena-ci I self-certify!

PR #417, opened 2020-08-07
Skip trying to provision online dependent devices if there are none

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #418, opened 2020-08-10
Sync contracts

sradevski:
Doing this makes it easier to switch to syncing with the device type contracts instead of S3, and is cleaner overall.


src/lib/device-types/sync/index.ts
```diff
@@ -0,0 +1,102 @@
...
+	const existingSlugs = new Set(
+		existingDeviceTypes.map((deviceType) => deviceType.slug),
+	);
+
+	return Bluebird.map(
+		mappedDeviceTypes,
+		async (deviceType) => {
+			try {
+				if (existingSlugs.has(deviceType.slug)) {
+					return await rootApi.patch({
```
sradevski, line 66: This used to happen in a transaction, but I don't think it's necessary

src/lib/device-types/sync/index.ts
```diff
@@ -0,0 +1,104 @@
...
+						options: { returnResource: false },
+					});
+				} catch (err) {
+					captureException(
+						err,
+						`Failed to synchronize device type: ${deviceType.slug}, skipping...`,
+					);
+				}
+			},
+			{ concurrency: 20 },
```
Page-, line 89: Is there a reason to add this now?

sradevski: @Page- just made sense to me so we don't just dump 100s of requests at once (once we have that many DTs), for example, and it seemed fine since DT sync is not time-sensitive anyway. Would you prefer if we avoid it?

Page-: I think it's better to avoid it, using the concurrency flag just extends the duration of the transaction which can cause its own problems

src/lib/device-types/sync/index.ts
```diff
@@ -0,0 +1,104 @@
...
+					return await rootApi.post({
+						resource: 'device_type',
+						body: deviceType,
+						options: { returnResource: false },
+					});
+				} catch (err) {
+					captureException(
+						err,
+						`Failed to synchronize device type: ${deviceType.slug}, skipping...`,
+					);
```
Page-, line 86: If a request within a transaction fails then you can't really ignore the error because it'll almost certainly roll back the transaction and cause any other db calls to fail

sradevski: Yeah makes sense, I just didn't want to fail all DT sync because of one having issues in the device-type.json for example, but I guess it's better to fail fast.

src/index.ts
```diff
@@ -29,6 +29,7 @@ import {
 } from './lib/config';
 
 import * as _applicationRoutes from './routes/applications';
+import { startDeviceTypeSynchronization } from './lib/device-types/sync';
```
dfunckt, line 32: Can you please coalesce this import with the other one below?

src/lib/device-types/index.ts
```diff
@@ -16,27 +17,14 @@ import {
...
 import { getImageKey, IMAGE_STORAGE_PREFIX, listFolders } from './storage';
-
-const { api } = sbvrUtils;
-const { BadRequestError, InternalRequestError, NotFoundError } = errors;
-export type { BadRequestError, NotFoundError };
+import {
+	UnknownDeviceTypeError,
+	InvalidDeviceTypeError,
+	UnknownVersionError,
+} from '../errors';
```
dfunckt, line 24: Why did you move these errors out of here? They're quite specific to device types and here (or a in new file `lib/device-types/`) is their place IMO.

src/lib/device-types/sync/index.ts
```diff
@@ -0,0 +1,95 @@
+import * as _ from 'lodash';
```
dfunckt, line 1: What's the point of making `sync/` a folder if it only has a single file?

src/lib/device-types/index.ts
sradevski: @dfunckt the same can be said about all errors that are in `errors.ts`, since they are also quite specific (`InaccessibleAppError`, `NoDevicesFoundError`, etc.). The reason I think it's better if they are in errors is that even though the error is for a device type, it can be used in different contexts (when creating an app, when provisioning a device, etc.), and we should either have all error definitions in one place, or have them all defined in the resources they act on (app, device, etc.). Currently, it is a mix of the two, and this change just makes everything consistent. Does that make sense?

src/lib/device-types/sync/index.ts
sradevski: I plan on adding more (I have done so locally) as we switch the syncing to contracts, but I can also move it out of the folder for the time being if you prefer that

src/lib/device-types/index.ts
dfunckt: Gotcha, that makes sense to me -- agreed.

src/lib/device-types/sync/index.ts
dfunckt: Yes please, don't commit to future stuff. A rename from `device-types/sync.ts` to `device-type/sync/index.ts` is a trivial diff too.

src/lib/device-types/sync.ts
```diff
@@ -0,0 +1,95 @@
...
+						resource: 'device_type',
+						body: deviceType,
+						options: {
+							$filter: { slug: deviceType.slug },
+							returnResource: false,
+						},
+					});
+				}
+
+				return rootApi.post({
```
Page-, line 75: for better stack traces:
```suggestion
				return await rootApi.post({
```

src/lib/device-types/sync.ts
```diff
@@ -0,0 +1,95 @@
...
+		})) as Array<{ slug: string }>;
+
+		const existingSlugs = new Set(
+			existingDeviceTypes.map((deviceType) => deviceType.slug),
+		);
+
+		return Promise.all(
+			mappedDeviceTypes.map((deviceType) => {
+				if (existingSlugs.has(deviceType.slug)) {
+					return rootApi.patch({
```
Page-, line 65: for better stack traces:
```suggestion
					return await rootApi.patch({
```

src/lib/device-types/sync.ts
```diff
@@ -0,0 +1,95 @@
...
+	);
+};
+
+const syncDeviceTypes = async () => {
+	const deviceTypes = await getAllDeviceTypes();
+	const mappedDeviceTypes = deviceTypes.map((deviceType) =>
+		mapModel(deviceType, syncSettings.map),
+	);
+
+	return sbvrUtils.db.transaction(async (tx) => {
```
Page-, line 48: for better stack traces:
```suggestion
	return await sbvrUtils.db.transaction(async (tx) => {
```

Page-, line 48: Do we care about the return value?

src/lib/device-types/sync.ts
```diff
@@ -0,0 +1,95 @@
...
+	return Object.keys(map).reduce(
+		(
+			mappedModel: { [k in DeviceTypeResourceKeys]?: any },
+			key: Partial<DeviceTypeResourceKeys>,
+		) => {
+			mappedModel[key] = deviceType[map[key]!.name] ?? map[key]!.default;
+			return mappedModel;
+		},
+		{},
+	);
```
Page-, line 39: I think it's clearer and more concise for a straightforward for loop:
```suggestion
	const mappedModel: { [k in DeviceTypeResourceKeys]?: any } = {};
	for (const key of Object.keys(map)) {
		mappedModel[key] = deviceType[map[key]!.name] ?? map[key]!.default;
	}
	return mappedModel;
```

src/lib/device-types/sync.ts
sradevski: we don't no 👍

sradevski:
@balena-ci rebase

src/lib/device-types/sync.ts
```diff
@@ -0,0 +1,91 @@
...
+	map: { slug: { name: 'slug' } },
+};
+
+export function setSyncMap(map: typeof syncSettings['map']) {
+	syncSettings.map = _.merge(map, syncSettings.map);
+}
+
+const mapModel = (deviceType: DeviceType, map: typeof syncSettings['map']) => {
+	const mappedModel: { [k in DeviceTypeResourceKeys]?: any } = {};
+	for (const key of Object.keys(map) as Array<keyof typeof map>) {
```
Page-, line 31: Is the cast actually necessary? I would assume that's the default type?

src/lib/device-types/sync.ts
```diff
@@ -0,0 +1,91 @@
...
+		const existingDeviceTypes = (await rootApi.get({
+			resource: 'device_type',
+			options: { $select: ['slug'] },
+		})) as Array<{ slug: string }>;
+
+		const existingSlugs = new Set(
+			existingDeviceTypes.map((deviceType) => deviceType.slug),
+		);
+
+		return Promise.all(
```
Page-, line 58: Since we don't care about the return value:
```suggestion
		await Promise.all(
```

src/lib/device-types/sync.ts
```diff
@@ -0,0 +1,91 @@
...
+						resource: 'device_type',
+						body: deviceType,
+						options: {
+							$filter: { slug: deviceType.slug },
+							returnResource: false,
+						},
+					});
+				}
+
+				return await rootApi.post({
```
Page-, line 71: Since we don't care about the return value:
```suggestion
				await rootApi.post({
```

src/lib/device-types/sync.ts
sradevski: I also thought it should, but it only recognizes it as a `string`

sradevski:
@balena-ci rebase

PR #419, opened 2020-08-11
Update pinejs and use $count sugar

Page-:
Update pinejs from 13.4.2 to 13.5.0
Update typed-error from 3.2.0 to 3.2.1

Change-type: patch


src/platform/middleware.ts
```diff
@@ -108,7 +110,7 @@ export const gracefullyDenyDeletedDevices: RequestHandler = async (
 
 	const returnCode = req.method === 'GET' ? 304 : 200;
 
-	const devices = (await checkDeviceExistsQuery()({ uuid })) as number;
```
thgreasi, line 111: :sparkles: :mage:

PR #420, opened 2020-08-11
Update pinejs and make use of improved typings

Page-:
Update pinejs from 13.5.0 to 13.6.0

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #421, opened 2020-08-12
Update pinejs to 13.6.1

Page-:
Update pinejs from 13.6.0 to 13.6.1

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #423, opened 2020-08-17
Update dependencies

Page-:
Update pinejs from 13.7.0 to 13.8.1

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #425, opened 2020-08-19
Vastly increase permitted app name length

dfunckt:
Application name was used as part of the image name deployed to devices, which meant it had to be compatible with Docker’s name requirements. We’ve long moved away from that naming scheme so the limit is now superfluous and this patch effectively removes it.

Change-type: patch


dfunckt:
@balena-ci I self-certify!

PR #426, opened 2020-08-20
Update pinejs to 13.13.2

Page-:
Update pinejs from 13.8.1 to 13.13.2

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #436, opened 2020-08-24
Fix the device type synchronization interval to 5 minutes

sradevski:
Change-type: patch


sradevski:
@balena-ci I self-certify!

PR #441, opened 2020-08-24
Extract device-proxy to a feature

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #452, opened 2020-08-26
Add a "cpu_architecture" resource

sradevski:
As discussed in the architecture call on 25th of August 2020, the device type resource will have a foreign key to an
architecture, which for now only has a slug.

Change-type: minor

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -7,7 +7,7 @@ export = {
 		{
 			username: 'guest',
 			password: ' ',
-			permissions: ['resin.device_type.read'],
+			permissions: ['resin.device_type.read', 'resin.cpu_architecture.read'],
 		},
 	],
 } as ConfigLoader.Config;
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -274,6 +274,7 @@ Term: vpn address
 -- Complex terms
 
 Term: application type
+Term: cpu architecture
 Term: config
 Term: device type
 Term: image
@@ -553,6 +554,13 @@ Fact type: device type has slug
 
 Fact type: device type has name (Auth)
 	Necessity: each device type has exactly one name (Auth)
+
+-- cpu architecture
+
+Fact type: cpu architecture has slug
+	Necessity: each cpu architecture has exactly one slug
+	Necessity: each slug is of exactly one cpu architecture
+
 
 -- release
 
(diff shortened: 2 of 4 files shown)


src/migrations/00031-drop-supervisor-release.sql
```diff
@@ -1,7 +1,7 @@
-DROP TRIGGER "supervisor release_trigger_update_modified_at"
-DROP INDEX "device_supervisor_release_device_type_idx";
+DROP TRIGGER IF EXISTS "supervisor release_trigger_update_modified_at" ON "supervisor release";
```
sradevski, line 1: Due to the syntax error here, the API wouldn't even start up

src/balena.sbvr
```diff
@@ -250,6 +250,7 @@ Term: vpn address
 -- Complex terms
 
 Term: application type
+Term: architecture
```
Page-, line 253: I think it'd make sense to name this as `cpu architecture` since `architecture` is potentially generic and could clash

sradevski: @Page- IMO architecture is fine, since the same can be said for `application`, `device` etc. I don't have a strong opinion though, just trying to think through whether the added verbosity helps in any way, and just `architecture` feels more consistent with the rest of the resource naming. What do you think? How else would we use the term `architecture` in the DB, for example?

sradevski:
@balena-ci rebase

src/balena.sbvr
Page-: I would say architecture is more generic, it could be used to refer to the network architecture of the device, the architecture of the device's PCB and how buses/etc are laid out, even "CPU architecture" can mean either the CPU architecture in terms of NUMA nodes/etc or the instruction set architecture, which is what we care about for this resource. Generally starting off with a more specific name is going to reduce the number of support questions that might come up around what "architecture" we're referring to and also saves the need to rename it later

dfunckt: FWIW I agree with Page that "cpu architecture" is the proper term here -- it is a term of art.

sradevski: Ok, you convinced me :)

sradevski: @Page- can I get a re-review please?

config.ts
```diff
@@ -7,7 +7,7 @@ export = {
 		{
 			username: 'guest',
 			password: ' ',
-			permissions: ['resin.device_type.get'],
+			permissions: ['resin.device_type.get', 'resin.cpu_architecture.get'],
```
Page-, line 10: This needs to change to
```suggestion
			permissions: ['resin.device_type.read', 'resin.cpu_architecture.read'],
```
in order to continue working after the rebase

PR #455, opened 2020-08-26
Convert `device belongs to application` into a link table

Page-:
Change-type: patch


src/balena.sbvr
```diff
@@ -716,7 +717,7 @@ Fact type: config has description (Auth)
 
 -- Rules
 
-Rule: It is necessary that each device1 that is managed by a device2, belongs to an application1 that depends on an application2 that owns the device2.
+Rule: It is necessary that each device1 that is managed by a device2, belongs to no applications or belongs to an application1 that depends on an application2 that owns the device2.
```
dfunckt, line 720: > no applications

Is plural a thing?

src/features/device-provisioning/register.ts
```diff
@@ -69,6 +74,30 @@ export const register: RequestHandler = async (req, res) => {
 			if (device == null) {
 				throw new Error('Failed to create device');
 			}
+			if (req.apiKey?.permissions != null) {
+				// Add the permission to assign the created device to the current application for application/provisioning keys, will have no effect for device/user keys
+				req.apiKey.permissions.push(
+					`resin.device__belongs_to__application.create?device eq ${device.id} and belongs_to__application/any(a:a/actor eq @__ACTOR_ID)`,
+					`resin.application.read?actor eq @__ACTOR_ID`,
```
dfunckt, line 81: These seem identical to above -- can you please move them someplace where both hooks can use them?

src/lib/auth.ts
```diff
@@ -39,6 +40,9 @@ export const ROLES: {
 		'resin.application_tag.all',
 		'resin.application_type.all',
 		'resin.device.all',
+		'resin.device__belongs_to__application.read',
+		'resin.device__belongs_to__application.create',
+		'resin.device__belongs_to__application.delete',
 		'resin.device.tunnel-22222',
```
dfunckt, line 46:
```suggestion
		'resin.device.tunnel-22222',
		'resin.device__belongs_to__application.read',
		'resin.device__belongs_to__application.create',
		'resin.device__belongs_to__application.delete',
```

src/features/devices/hooks/defaults-validation.ts
```diff
@@ -58,15 +58,6 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
 			}
 		}
 
-		// When moving application make sure to set the build to null, unless a specific new
-		// build has been targeted, instead of pointing to a build of the wrong application
-		if (
-			request.values.belongs_to__application != null &&
-			request.values.should_be_running__release === undefined
-		) {
-			request.values.should_be_running__release = null;
```
dfunckt, line 67: Is this not needed anymore? I don't see an equivalent hook on `device_application`.

src/balena.sbvr
Page-: Yup, you can pluralize things, I think it even handles some edge case plurals as well as the easy case of `s` suffix. It'll work with/without the plural here but reads more naturally with

src/features/devices/hooks/defaults-validation.ts
Page-: It won't be needed when it's on device_application itself but for now it is, fixed

src/features/device-provisioning/register.ts
Page-: Technically I think it's only needed in the other places with changes I made to it, but it will have to move here later - I've removed it now and with true multi-app it'll have to be reinstated

src/features/cascade-delete/hooks.ts
```diff
@@ -16,13 +15,97 @@ setupDeleteCascade('application', [
 	['application', 'depends_on__application'],
 ]);
 
+hooks.addPureHook('DELETE', 'resin', 'application', {
+	/**
+	 * Cascade from app -> device when deleting the app to match pre-single-multi-app behavior
```
dfunckt, line 20: Will proper multi-app permit devices with no apps? If not, you could delete the device if it was only associated with the application the request is deleting. You may want to add a feature-specific TODO here either way as a reminder.

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -73,20 +73,21 @@ const createAppServiceInstalls = async (
 	api: sbvrUtils.PinejsClient,
 	appId: number,
 	deviceIds: number[],
-): Promise<void> =>
-	createReleaseServiceInstalls(api, deviceIds, {
+): Promise<void> => {
+	await createReleaseServiceInstalls(api, deviceIds, {
```
dfunckt, line 77: This is possibly fixing issues with things being inaccessible for a bit due to permissions on the cloud -- did you observe any yourself?

src/features/devices/hooks/moving.ts
```diff
@@ -1,90 +1,77 @@
...
-					deviceIds,
-				),
-			]);
-
-			// get the devices being affected and store them for the POSTRUN...
-			request.custom.movedDevices = devices.map((device) => device.id);
+				},
+			});
 		}
+		// Store the devices being affected the POSTRUN
```
dfunckt, line 45:
```suggestion
		// Store the devices being affected for POSTRUN
```

src/features/cascade-delete/hooks.ts
Page-: Yes it definitely allows it, a device will be part of a fleet and a fleet can have as many or as few apps as it wants - and a fleet with no apps is essentially a "waiting room"/"staging area"

src/features/ci-cd/hooks/service-installs.ts
Page-: This shouldn't have any functional change, it just improves potential stack traces

src/features/ci-cd/hooks/device-update-trigger.ts
```diff
@@ -8,11 +8,7 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
 		// Update devices if they have had their:
 		// * target release changed
```
dfunckt, line 9:
```suggestion
```

src/features/ci-cd/hooks/device-update-trigger.ts
```diff
@@ -25,3 +17,62 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
+		});
+	},
+});
+
+hooks.addPureHook('PATCH', 'resin', 'device_application', {
+	POSTRUN: async (args) => {
+		const affectedIds = args.request.affectedIds!;
+		if (
+			affectedIds.length === 0 ||
+			args.request.values.should_be_running__release === undefined
```
thgreasi, line 46: Do we also need to handle `device_application` PATCHes of the app property?

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -95,78 +96,82 @@ hooks.addPureHook('POST', 'resin', 'device', {
...
+				id: request.values.should_be_running__release,
+			});
+		} else {
+			const devices = (await api.get({
+				resource: 'device',
+				options: {
+					$select: 'id',
+					$expand: {
+						device_application: {
+							$select: ['belongs_to__application'],
```
thgreasi, line 145: How about:
```suggestion
			const devices = (await api.get({
				resource: 'device_application',
				options: {
					$select: ['device', 'belongs_to__application'],
```
This removes the need for a JOIN and should also allow us to avoid doing the `[0]` in the `devicesByApp[groupId][0]` and also drop the `?` from `?.belongs_to__application.__id`?

src/features/device-provisioning/register.ts
```diff
@@ -49,21 +49,24 @@ export const register: RequestHandler = async (req, res) => {
...
+					req,
+					tx,
+				},
 				body: {
 					belongs_to__user: userId,
-					belongs_to__application: applicationId,
 					device_type: deviceType,
 					supervisor_version: supervisorVersion,
 					uuid,
+					belongs_to__application: applicationId,
```
thgreasi, line 69: Does this need the famous `// TODO-MULTI-APP: This relies on the hook creating the device_application` comment?

src/lib/auth.ts
```diff
@@ -40,6 +41,10 @@ export const ROLES: {
 		'resin.application_type.all',
 		'resin.device.all',
 		'resin.device.tunnel-22222',
+		'resin.device__belongs_to__application.read',
+		'resin.device__belongs_to__application.create',
+		'resin.device__belongs_to__application.update',
+		'resin.device__belongs_to__application.delete',
```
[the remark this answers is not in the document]
Page-: Yep, the current version is an artifact of how the multiple commits came together

src/features/device-state/routes/state.ts
```diff
@@ -94,19 +93,11 @@ const stateQuery = _.once(() =>
 						},
 					},
 				},
-				belongs_to__application: {
-					$select: ['id', 'app_name'],
+				device_application: {
+					$select: 'id',
```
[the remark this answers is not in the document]
Page-: I'd rather leave it as `id` to avoid confusion around `$select`ing a `$expand`, especially if we're just selecting one out of multiple expands

src/features/ci-cd/hooks/device-update-trigger.ts
Page-: Yes and no, they aren't mean to be allowed to be updated so doing so is a mistake - technically open-balena-api grants much more access so it is possible but doing so is not intended and would ideally be blocked

Page-:
@balena-ci retest

PR #457, opened 2020-08-26
Remove unnecessary passing of app to every `supertest` usage

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #463, opened 2020-08-28
Export tag hooks registration helper

dfunckt:
- Rename `tags/tags.ts` to `tags/validation.ts`
- Move, rename and export `registerTagHooks` helper

Change-type: patch


dfunckt:
@balena-ci I self-certify!

PR #465, opened 2020-08-31
Export the heartbeat related DeviceOnlineStates enum

thgreasi:
I will need that for https://github.com/balena-io/balena-api/pull/2353

Change-type: patch

### DIFF
--- a/src/features/device-heartbeat/index.ts
+++ b/src/features/device-heartbeat/index.ts
@@ -122,7 +122,7 @@ export const getPollInterval = async (uuid: string) => {
 export const POLL_JITTER_FACTOR = 1.5;
 
 // these align to the text enums coming from the SBVR definition of available values...
-export const enum DeviceOnlineStates {
+export enum DeviceOnlineStates {
 	Unknown = 'unknown',
 	Timeout = 'timeout',
 	Offline = 'offline',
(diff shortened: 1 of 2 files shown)


src/features/device-heartbeat/index.ts
```diff
@@ -122,7 +122,7 @@ export const getPollInterval = async (uuid: string) => {
 export const POLL_JITTER_FACTOR = 1.5;
 
 // these align to the text enums coming from the SBVR definition of available values...
-export const enum DeviceOnlineStates {
```
thgreasi, line 125: We can't re-export const enums as part of the `device` object, since they are supposed to be replaced with string values on build time.
See: https://app.circleci.com/pipelines/github/balena-io/open-balena-api/5280/workflows/946eb8f6-a24b-453e-8661-955cbd9a3a77/jobs/4976

I would expect that this wouldn't be an issue, given that we use `preserveConstEnums` in our tsconfig, but it ends up that TS still complains.

dfunckt (approved):

Page- (approved):

thgreasi:
@balena-ci rebase

thgreasi:
@balena-ci retest

thgreasi:
@balena-ci retest

PR #466, opened 2020-08-31
Enforce minimum os versions when creating a device

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #470, opened 2020-09-03
Add device metric fields

richbayliss:
metadata: Add device metadata

Allow the supervisor to send device metadata to the API in the
state patch. We can accept CPU, memory, and disk health.

Change-type: patch
Signed-off-by: Page <page@balena.io>


thgreasi:
HQ: https://github.com/balena-io/balena-io/issues/2264

CameronDiver (approved):

src/migrations/00032-add-device-metric-fields.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE "device"
+ADD COLUMN IF NOT EXISTS "memory usage" INTEGER NULL,
+ADD COLUMN IF NOT EXISTS "memory total" INTEGER NULL,
+ADD COLUMN IF NOT EXISTS "storage block device" VARCHAR(255) NULL,
+ADD COLUMN IF NOT EXISTS "storage usage" INTEGER NULL,
+ADD COLUMN IF NOT EXISTS "storage total" INTEGER NULL,
+ADD COLUMN IF NOT EXISTS "cpu usage" INTEGER NULL,
+ADD COLUMN IF NOT EXISTS "cpu temp" INTEGER NULL,
+ADD COLUMN IF NOT EXISTS "cpu id" VARCHAR(255) NULL,
+ADD COLUMN IF NOT EXISTS "is undervolted" INTEGER DEFAULT 0 NOT NULL;
```
LucianBuzzo, line 10: Why is this the only one that can't be null?

thgreasi: Seems so, since it makes sense to have NULLs for older supervisors that have never reported any metadata, rather than a default value like 0 or empty string.

Page-: It's a boolean field @LucianBuzzo and they can only be true/false in pinejs, it doesn't allow for three-valued logic there. Although you could choose to make a three-valued by doing:
```
Fact type: x has integer value
Necessity: each x has at most one integer value
Definition: 0 or 1

PR #473, opened 2020-09-10
Start synchronizing resources with contracts

sradevski:
Change-type: minor


test/fixtures/s3/listObjectsV2.json
```diff
@@ -67,7 +65,7 @@
 			{ "Prefix": "images/dt-with-no-valid-releases/" },
 			{ "Prefix": "images/dt-with-failing-listing/" }
 		],
-		"KeyCount": 55
+		"KeyCount": 57
```
sradevski, line 68: This was incorrect to begin with, but all of S3 things should go away once this PR is merged and we enrich the device type model with everything needed.

test/02-device-types.ts
```diff
@@ -14,7 +14,8 @@ describe('device type endpoints', () => {
 				expect(deviceType).to.have.property('name').that.is.a('string');
 			});
 
-			expect(res.body.d).to.have.property('length', 57);
+			// This also includes the fake test device types that we are adding.
+			expect(res.body.d).to.have.property('length', 97);
```
sradevski, line 18: There are many more simply because they now come from the contracts tarball

test/02-device-types.ts
```diff
@@ -1,7 +1,64 @@
...
+				slug: 'dt-with-no-valid-releases',
+				name: 'dt-with-no-valid-releases',
+			},
+			{
+				id: 998,
+				slug: 'dt-with-failing-listing',
+				name: 'dt-with-failing-listing',
+			},
+		].map(async (dt) => {
+			return await api.resin.post({
```
Page-, line 52:
```suggestion
			await api.resin.post({
```

src/features/contracts/index.ts
```diff
@@ -0,0 +1,196 @@
...
+	await upsertEntries(
+		mappedModel,
+		existingKeys,
+		typeMap.resource,
+		typeMap.uniqueKey,
+		rootApi,
+	);
+};
+
+export const synchronizeContracts = memoizee(
```
Page-, line 159: Is there any reason to memoize this? It should only be called in a single place when in runtime code so I'm not sure what the intention of memoizing is

src/features/contracts/index.ts
```diff
@@ -0,0 +1,196 @@
...
+	const mappedModel = contracts.map((contract) =>
+		mapModel(contract, typeMap.map),
+	);
+
+	const rootApi = sbvrUtils.api.resin.clone({
+		passthrough: { req: permissions.root },
+	});
+
+	const existingEntries = (await rootApi.get({
+		resource: typeMap.resource,
```
Page-, line 142: Isn't the db resource a fixed thing since we have to declare it in the sbvr, I'm not sure on the use case for specifying it?

test/02-device-types.ts
```diff
@@ -1,7 +1,64 @@
...
+				resource: 'device_type',
+				passthrough: {
+					req: permissions.root,
+				},
+				body: dt,
```
Page-, line 57: Minor performance optimization since we don't care about the result:
```suggestion
				passthrough: {
					req: permissions.root,
				},
				options: { returnResource: false },
				body: dt,
```

test/test-lib/contracts-mock.ts
```diff
@@ -0,0 +1,42 @@
...
+		Location: `https://codeload.github.com/${owner}/${name}/legacy.tar.gz/master`,
+	});
+
+	const fileInterceptor = nock('https://codeload.github.com/')
+		.persist(persist)
+		.get(`/${owner}/${name}/legacy.tar.gz/master`);
+
+	fileInterceptor.replyWithFile(
+		200,
+		path.join(__dirname, `../fixtures/contracts/${filename ?? name}.tar.gz`),
```
Page-, line 24: I think you can use
```ts
filename?: = name
```
in the parameters and then use filename directly here (and below)
```suggestion
		path.join(__dirname, `../fixtures/contracts/${filename}.tar.gz`),
```

Page-, line 24: How big are these contracts? In general I'd appreciate it if you could make them a text file that we compress on the fly here. That'll make viewing/editing/diff'ing them much easier

src/features/contracts/index.ts
```diff
@@ -0,0 +1,196 @@
...
+	await Bluebird.map(
+		newData,
+		async (entry: any) => {
+			try {
+				if (existingData.has(entry[uniqueField])) {
+					return await rootApi.patch({
+						resource,
+						body: entry,
+						options: {
+							$filter: { slug: entry[uniqueField] },
```
Page-, line 101: Here you use `slug` for the db-side unique field but elsewhere you use `uniqueField` as the db-side field. I'm assuming that the unique field in the contract happens to be "slug" - could you do a test with the unique field being "foobar" and setting that in the mapping as I expect it'll catch some issues where you assume they have to match

src/features/contracts/contracts-directory.ts
```diff
@@ -0,0 +1,108 @@
...
+
+// Keeps the contract repos locally and in sync with upstream, if accessible.
+export const fetchContractsLocally = async (repos: RepositoryInfo[]) => {
+	await Promise.all(
+		repos.map(async (repo) => {
+			const get = request.get(
+				getArchiveLinkForRepo(repo),
+				getRequestOptions(repo),
+				handleResponse,
+			) as any;
```
Page-, line 72: Why is this cast necessary?

src/features/contracts/index.ts
sradevski: @Page- It was simply done so we can await the same promise in the tests, but I guess it's fine to just call synchronizeContracts twice, it shouldn't matter. let me remove it.

src/features/contracts/index.ts
sradevski: @Page- this is not just for device types, it is also for `cpu_architecture` currently, but it might be for other resources. This makes it possible to create a mapping between a contract and a db resource, and if that resource doesn't exist in sbvr, it will fail. You can see how `setSyncSettings` is used

src/features/contracts/index.ts
sradevski: Yeah I missed that one, nice catch.

src/features/contracts/contracts-directory.ts
sradevski: This is `request.Request` type, which inherits from `Stream`, but `pipeline` expects a `NodeJS.ReadableStream` instead, so there is a mismatch (even though `request.Request` can be readable or writable and is handled by a boolean internally it seems). I changed it to `as unknown as NodeJS.ReadableStream` to make it a bit clearer.

src/features/contracts/contracts-directory.ts
```diff
@@ -0,0 +1,108 @@
...
+
+// Keeps the contract repos locally and in sync with upstream, if accessible.
+export const fetchContractsLocally = async (repos: RepositoryInfo[]) => {
+	await Promise.all(
+		repos.map(async (repo) => {
+			const get = (request.get(
+				getArchiveLinkForRepo(repo),
+				getRequestOptions(repo),
+				handleResponse,
+			) as unknown) as NodeJS.ReadableStream;
```
Page-, line 72: Can you add a comment with the explanation for the cast please

src/features/contracts/contracts-directory.ts
```diff
@@ -0,0 +1,108 @@
...
+
+			const untar = tar.x({
+				C: await prepareContractDirectory(repo),
```
Page-, line 75: I'd prefer to use "long flags" for readability:
```suggestion
			const untar = tar.extract({
				cwd: await prepareContractDirectory(repo),
```

test/test-lib/contracts-mock.ts
```diff
@@ -0,0 +1,51 @@
...
+	});
+
+	const fileInterceptor = nock('https://codeload.github.com/')
+		.persist(persist)
+		.get(`/${owner}/${name}/legacy.tar.gz/master`);
+
+	fileInterceptor.reply(
+		200,
+		() => {
+			return tar.c(
```
Page-, line 26:
```suggestion
			return tar.create(
```

test/09_contracts.ts
```diff
@@ -0,0 +1,145 @@
...
+
+			const dbDeviceTypes = (
+				await supertest().get('/resin/device_type').expect(200)
+			).body.d;
+
+			const newDt = dbDeviceTypes.find(
+				(dbDeviceType: any) => dbDeviceType.slug === 'new-dt',
+			);
+			const finDt = dbDeviceTypes.find(
+				(dbDeviceType: any) => dbDeviceType.slug === 'fincm3',
```
Page-, line 137: Is this cast necessary?

test/09_contracts.ts
```diff
@@ -0,0 +1,145 @@
...
+
+			mockRepo('balena-io', 'contracts', 'updated-base-contracts');
+			await synchronizeContracts([{ owner: 'balena-io', name: 'contracts' }]);
+
+			const dbDeviceTypes = (
+				await supertest().get('/resin/device_type').expect(200)
+			).body.d;
+
+			const newDt = dbDeviceTypes.find(
+				(dbDeviceType: any) => dbDeviceType.slug === 'new-dt',
```
Page-, line 134: Is this cast necessary?

test/09_contracts.ts
```diff
@@ -0,0 +1,145 @@
...
+			mockRepo('balena-io', 'contracts', 'updated-base-contracts');
+			await fetchContractsLocally([{ owner: 'balena-io', name: 'contracts' }]);
+
+			const contracts = await getContracts('hw.device-type');
+
+			const newDt = contracts.find(
+				(dbDeviceType: any) => dbDeviceType.slug === 'new-dt',
+			);
+			const finDt = contracts.find(
+				(dbDeviceType: any) => dbDeviceType.slug === 'fincm3',
```
Page-, line 81: Is this cast necessary?

test/09_contracts.ts
```diff
@@ -0,0 +1,145 @@
...
+			await fetchContractsLocally([{ owner: 'balena-io', name: 'contracts' }]);
+
+			// A new device type was added, and the fin contract was modified in the updated contracts tarball.
+			mockRepo('balena-io', 'contracts', 'updated-base-contracts');
+			await fetchContractsLocally([{ owner: 'balena-io', name: 'contracts' }]);
+
+			const contracts = await getContracts('hw.device-type');
+
+			const newDt = contracts.find(
+				(dbDeviceType: any) => dbDeviceType.slug === 'new-dt',
```
Page-, line 78: Is this cast necessary?

PR #476, opened 2020-09-17
Update dependencies

Page-:
Update pinejs from 14.0.0 to 14.1.0

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #481, opened 2020-10-05
Fix fetching contracts randomly failing due to a bug in request

sradevski:
Based on my tests this should fix it (I added `Bluebird.delay(2000)` before calling `pipeline` to reproduce it). The reason this works is, `PassThrough` doesn't switch to `flowing` mode until it is piped somewhere, unlike `request.get`

Change-type: patch

### DIFF
--- a/src/features/contracts/contracts-directory.ts
+++ b/src/features/contracts/contracts-directory.ts
@@ -65,17 +65,17 @@ const handleResponse: request.RequestCallback = (err, response) => {
 export const fetchContractsLocally = async (repos: RepositoryInfo[]) => {
 	await Promise.all(
 		repos.map(async (repo) => {
+			const untar = tar.extract({
+				C: await prepareContractDirectory(repo),
+				strip: 1,
+			});
+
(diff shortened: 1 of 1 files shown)


src/features/contracts/contracts-directory.ts
```diff
@@ -65,12 +65,15 @@ const handleResponse: request.RequestCallback = (err, response) => {
...
-			) as unknown) as NodeJS.ReadableStream;
+			// Streaming to a passthrough stream fixes an issue when request.get starts draining before it is passsed to `pipeline`, see https://github.com/request/request/issues/2696, https://github.com/request/request/issues/887
+			const pass = new stream.PassThrough();
+			const get = request
+				.get(
+					getArchiveLinkForRepo(repo),
+					getRequestOptions(repo),
+					handleResponse,
+				)
+				.pipe(pass);
```
Page-, line 76: Can you move this after the `await prepareContractDirectory` and avoid the passthrough buffering the response in memory/keep backpressure support

sradevski: Yeah that would work as well, I just found using the passthrough to be more resilient to timing issues, but any solution would do really.

Page- (approved):

Issue #267, opened 2020-01-30
Heartbeat is updated regardless of request source


xginn8:
Any direct calls to a device's state endpoint will update the heartbeat state, even if the request came from a non-device user-agent or API key.

Simple reproduction:

```bash
❯ curl -qs -X GET  "https://api.balena-cloud.com/v5/device?\$filter=uuid%20eq%20'$UUID'" \
-H "Content-Type: application/json" \
-H "Authorization: Bearer $token" | jq -r '.d[0] | "\(.api_heartbeat_state) \(.last_connectivity_event)"'
offline 2019-12-03T23:05:41.039Z

~
❯ curl -qs -H "authorization: bearer $token" "https://api.balena-cloud.com/device/v2/$UUID/state" > /dev/null

~
❯ curl -qs -X GET  "https://api.balena-cloud.com/v5/device?\$filter=uuid%20eq%20'$UUID'" \                   
-H "Content-Type: application/json" \
-H "Authorization: Bearer $token" | jq -r '.d[0] | "\(.api_heartbeat_state) \(.last_connectivity_event)"'
online 2019-12-03T23:05:41.039Z
```

PR #482, opened 2020-10-05
device-heartbeat: Only update for requests using the device API Key

thgreasi:
I suggest reviewing the commits individually, since the first is only adding tests for the current state and then the second will just be the actual changes.

Resolves: #267
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/5zTYkavDVQisCX1p9iI2ZQsmUcf

### DIFF
--- a/src/features/api-keys/lib.ts
+++ b/src/features/api-keys/lib.ts
@@ -221,3 +221,42 @@ export const createGenericApiKey = async (
 		},
 	);
 };
+
+export const isApiKeyWithRole = async (
+	key: string,
+	roleName: string,
+	tx?: Tx,
+) => {
+	const role = await api.Auth.get({
+		resource: 'role',
+		passthrough: { tx, req: permissions.root },
+		id: {
+			name: roleName,
+		},
+		options: {
+			$select: 'id',
+			$filter: {
+				is_of__api_key: {
+					$any: {
+						$alias: 'khr',
+						$expr: {
+							khr: {
+								api_key: {
+									$any: {
+										$alias: 'k',
+										$expr: {
+											k: {
+												key,
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	});
+	return role != null;
+};
--- a/src/features/device-heartbeat/index.ts
+++ b/src/features/device-heartbeat/index.ts
@@ -7,6 +7,7 @@ import { RedisClient } from 'redis';
 import { sbvrUtils, permissions } from '@balena/pinejs';
 
 import { captureException } from '../../infra/error-handling';
+import { isApiKeyWithRole } from '../api-keys/lib';
 
 import { events as deviceStateEvents } from '../device-state';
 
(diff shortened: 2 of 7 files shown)


src/infra/auth/api-keys.ts
```diff
@@ -27,3 +29,42 @@ export const retrieveAPIKey = async (
 		await permissions.authorizationMiddleware(req);
 	}
 };
+
+export const isApiKeyOfActorType = async (
```
thgreasi, line 33: This will replace a pre-existing part of the cloud API.

Page-: `infra` is for bits that would in theory move into pinejs or a pinejs module - it shouldn't have anything balena-specific

src/features/device-heartbeat/index.ts
```diff
@@ -429,9 +430,21 @@ export class DeviceOnlineStateManager extends events.EventEmitter {
...
+		deviceStateEvents.on('get-state', async (uuid, { apiKey }) => {
 			try {
-				const pollInterval = await getPollInterval(uuid);
+				const key = apiKey?.key;
+				if (typeof key !== 'string') {
+					return;
+				}
+
+				const [isDeviceApiKey, pollInterval] = await Promise.all([
+					isApiKeyOfActorType('device', key),
```
Page-, line 441: I would rather this checks for the api key role via `device-api-key` because in theory we could have other device keys that are scoped/etc and should not update the status

src/features/device-state/index.ts
```diff
@@ -24,7 +24,10 @@ export const setup = (app: Application) => {
 };
 
 export interface Events {
-	'get-state': (uuid: string) => void;
+	'get-state': (
+		uuid: string,
+		req: Pick<Request, 'apiKey' | 'prefetchApiKey' | 'user'>,
```
Page-, line 29: Is there any reason for exporting all of these rather than just `apiKey`?

thgreasi: Picking just the apiKey instead of all the possible ways that the request might be authenticated, felt a bit like hardcoding the pub-sub implementation for the heartbeat needs. Will change to just pick the `apiKey` and we can extend later if needed :+1:

src/features/device-heartbeat/index.ts
thgreasi: @Page- does `isApiKeyWithRole` sounds closed to what you mean?
If so, does it make sense to have it in `infra`?

src/features/api-keys/lib.ts
```diff
@@ -221,3 +221,42 @@ export const createGenericApiKey = async (
...
+	tx?: Tx,
+) => {
+	const role = await api.Auth.get({
+		resource: 'role',
+		passthrough: { tx, req: permissions.root },
+		id: {
+			name: roleName,
+		},
+		options: {
+			$select: '*',
```
Page-, line 237: It'd be better to just select a single property, eg
```suggestion
			$select: 'id',
```

thgreasi: Looks like I forgot a debug * in there :(

Page- (approved):

PR #483, opened 2020-10-06
Bump OData version to v6

dfunckt:
Change-type: minor
Refs: https://github.com/balena-io/open-balena/pull/88

### DIFF
--- a/src/features/devices/models/device-additions.ts
+++ b/src/features/devices/models/device-additions.ts
@@ -1,0 +1,360 @@
+import {
+	AbstractSqlModel,
+	AbstractSqlQuery,
+	AndNode,
+	EqualsNode,
+} from '@balena/abstract-sql-compiler';
+
+export const addToModel = (
+	abstractSql: AbstractSqlModel,
+	addShims: boolean = true,
+) => {
+	if (addShims) {
+		// FIXME: this is a shim for the `is web accessible` attribute on the device
+		// resource. The core API has no support for "Device Public URLs".
+		abstractSql.tables['device'].fields.push({
+			fieldName: 'is web accessible',
+			dataType: 'Boolean',
+			// The cast is needed because AbstractSqlQuery cannot express a constant value.
+			computed: ['Boolean', false] as AbstractSqlQuery,
+		});
+	}
+
+	// FIXME: cloud has the notion of active/inactive devices via a field
+	// on the device resource. Computing the overall status below depends
+	// on whether the device is active but also on a few other statuses
+	// (Ordered, Preparing, etc.) that make no sense for the core API.
+	//
+	// We must somehow expose the overall status/progress attributes on the
+	// model and we could contrive a way for the cloud to inject the extra
+	// cases, but that would quickly become overly complex because the order
+	// of cases matters. We'd never be able to cleanly make the core API
+	// agnostic to the cloud.
+	//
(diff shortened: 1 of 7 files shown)


dfunckt:
@Page- I need your eyes on this one.

src/features/devices/models/device-additions.ts
```diff
@@ -0,0 +1,367 @@
...
+	// of cases matters. We'd never be able to cleanly make the core API
+	// agnostic to the cloud.
+	//
+	// Therefore, I opted for the most straightforward way, which is to switch
+	// the computation based on a flag. When `shimActiveField` is true (the
+	// default), it'll generate a dummy case (the equivalent of 1 == 2, which is
+	// always false), otherwise it'll generate a case that looks into the actual
+	// model field.
+	const isInactive: EqualsNode = shimActiveField
+		? ['Equals', ['Number', 1], ['Number', 2]]
```
Page-, line 37: Can this be shimmed in the same way as `is web accessible`?

src/migrations/00012-rename-root-user.sql
```diff
@@ -1 +1,11 @@
+CREATE TABLE IF NOT EXISTS "user-has-public key" (
+	"created at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL
+,	"user" INTEGER NOT NULL
+,	"public key" TEXT NOT NULL
+,	"id" SERIAL NOT NULL PRIMARY KEY
+,	"title" VARCHAR(255) NOT NULL
+,	FOREIGN KEY ("user") REFERENCES "user" ("id")
+,	UNIQUE("user", "public key")
+);
```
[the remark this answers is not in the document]
dfunckt: @Page- this is actually interesting -- this is needed because this table is referenced in the `modified-at` migration. We haven't released an open-balena version that would cause this migration to run, and have the table automatically created, so Pine will execute this migration and the `modified-at` one in the same "transaction" and fail. This is interesting because there's a gotcha here we haven't been taking into account: we may introduce a new table and in later migrations we assume this table exists because there's been a deployment.

src/features/devices/models/device-additions.ts
dfunckt: @Page- do you mean the specific case where `shimActiveField` is `true`? It would be nice, but the types are incompatible -- `isInactive` is an `EqualsNode` and `is web accessible` is cast to `AbstractSqlQuery`.

dfunckt: I guess I could type `isInactive` as `AbstractSqlQuery`?

Page-: I meant something like
```ts
	abstractSql.tables['device'].fields.push({
		fieldName: 'is active',
		dataType: 'Boolean',
		computed: ['Boolean', true],
	});
```
and then you don't need the shim here at all

dfunckt: @Page- I tried that before but doesn't work -- the field isn't being replaced in the query, so it fails with:

```
DatabaseError: column device.is active does not exist
```

Which is very unfortunate because this is the most elegant way to provide model-level default behaviour in core API.

Page- (approved):

dfunckt:
@balena-ci I self-certify!

PR #484, opened 2020-10-06
Improve `aliasTable` helper

Page-:
Change-type: patch

### DIFF
--- a/src/abstract-sql-utils.ts
+++ b/src/abstract-sql-utils.ts
@@ -22,6 +22,7 @@ export const aliasTable = (
 	abstractSqlModel.tables[toResourceName] = _.cloneDeep(
 		abstractSqlModel.tables[resourceName],
 	);
+	abstractSqlModel.tables[toResourceName].name = toResourceName;
 	if (definition) {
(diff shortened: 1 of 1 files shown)


src/abstract-sql-utils.ts
```diff
@@ -22,6 +22,7 @@ export const aliasTable = (
 	abstractSqlModel.tables[toResourceName] = _.cloneDeep(
 		abstractSqlModel.tables[resourceName],
 	);
+	abstractSqlModel.tables[toResourceName].name = toResourceName;
```
dfunckt, line 25: Is `name` used anywhere or you plan to use it? Why add it if not? How did it work previously otherwise?

Page-: It gets used by abstract-sql-compiler to generate the db name, but it's also planned to be used by https://github.com/balena-io-modules/abstract-sql-to-typescript/pull/1 - the reason the tests are failing is that setting it to a sensible value causes abstract-sql-compiler to actually generate a table for it.. so I have to handle that before this can be merged

Page-:
This was done as part of #535

PR #491, opened 2020-10-13
model: Add UUID to Application model

richbayliss:
Change-type: patch


thgreasi (commented):
LGTM but that's definitely a `minor` though.

src/features/applications/hooks/defaults.ts
```diff
@@ -14,6 +15,15 @@ hooks.addPureHook('POST', 'resin', 'application', {
 			);
 		}
 
+		request.values.uuid =
+			request.values.uuid || (await pseudoRandomBytesAsync(16)).toString('hex');
```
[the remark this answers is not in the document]
richbayliss: indeed @thgreasi I copied the existing code :grin:

Page-, line 19: That's not strictly a valid uuid btw, since uuid v4 has some header info around version stuff I think

richbayliss: true... we could be strict and enforce a fully valid UUIDv4.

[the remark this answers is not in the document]
richbayliss: updated this code to `??`

src/features/applications/hooks/defaults.ts
```diff
@@ -14,6 +15,30 @@ hooks.addPureHook('POST', 'resin', 'application', {
...
+		request.values.uuid = request.values.uuid ?? uuid.v4().replace(/-/g, '');
+
+		// reconstitute the value into a properly formatted UUID...
+		const toUuid = (i: string): string => {
+			if (i.length !== 32) {
+				return '';
+			}
+
+			const parts: string[] = [];
+			parts.push(i.substr(0, 8));
```
thgreasi, line 27: You could probably use `_.chunk(i, 4)`.
PS: And probably use a better param name than `i` ;)

thgreasi (approved):
Other than the NIT, LGTM :+1:

src/features/applications/hooks/defaults.ts
```diff
@@ -14,6 +15,30 @@ hooks.addPureHook('POST', 'resin', 'application', {
 			);
 		}
 
+		request.values.uuid = request.values.uuid ?? uuid.v4().replace(/-/g, '');
+
+		// reconstitute the value into a properly formatted UUID...
+		const toUuid = (i: string): string => {
```
thgreasi, line 21: We probably also don't need to redefine this function each time that the hook runs.

src/features/applications/hooks/defaults.ts
richbayliss: I didn't want to bring in lodash when a simple `substr` is ample.

Page- (approved):

PR #492, opened 2020-10-14
Add is of cpu architecture to device type

sradevski:
I need to test the migration first

Change-type: minor

### DIFF
--- a/src/features/contracts/index.ts
+++ b/src/features/contracts/index.ts
@@ -54,6 +54,10 @@ type FieldsMap = {
 	[dbField: string]: {
 		contractField: string;
 		default?: any;
+		refersTo?: {
+			resource: string;
+			uniqueKey: string;
+		};
 	};
 };
 // This map will hold information on which contract fields imported from the contract type will be synced to which db model and fields.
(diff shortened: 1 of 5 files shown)


src/migrations/00033-add-cpu-arch-to-device-type.sql
```diff
@@ -0,0 +1,9 @@
+-- In case the cpu architecture is empty, we just add a dummy value to satisfy the SBVR constraints.
+INSERT INTO "cpu architecture" ("slug") VALUES ('unknown');
```
sradevski, line 2: @Page- if sbvr is ran first the table should exist, but I see some migrations where we create a table as part of it. Is that necessary here? I just want to add a dummy cpu architecture if there is none, so we can satisfy the foreign key requirement. If you have a better suggestion on how to handle such case, I'm all ears :)

src/features/contracts/index.ts
```diff
@@ -164,8 +192,10 @@ export const synchronizeContracts = async (contractRepos: RepositoryInfo[]) => {
 		console.error(`Failed to fetch contracts, skipping...`, err.message);
 	}
 
-	await Promise.all(
-		['hw.device-type', 'arch.sw'].map(async (contractType) => {
+	// We don't have automatic dependency resolution, so the order matters here.
+	await Bluebird.map(
+		['arch.sw', 'hw.device-type'],
+		async (contractType) => {
```
Page-, line 198: If you want concurrency of 1 you can just use a for loop:
```suggestion
	for (const contractType of ['arch.sw', 'hw.device-type']) {
```

src/migrations/00033-add-cpu-arch-to-device-type.sql
Page-: We create the table if we need to insert data/modify it as part of the migration, in this case it should not be necessary as long as the various "keyframes" in the open-balena-api chain are deployed in order. Having the create table explicitly won't cause any issues though as long as you include the `IF NOT EXISTS`.

Also if you want this cpu arch to always be populated you should consider adding it to `balena-init.sql` as well

sradevski:
@balena-ci rebase

Page- (approved):

PR #488, opened 2020-10-12
tests: add basic migration tests

xginn8:
Change-type: patch
Signed-off-by: Matthew McGinn <matthew@balena.io>

PR #497, opened 2020-10-26
Add tests that validate migrations

Page-:
Change-type: patch


dfunckt:
There’s also #488

xginn8 (approved):

dfunckt:
@Page- if this works properly, tests should be failing. #483 fixes the issues and you may want to take those commits and bundle them all together.

Page-:
@dfunckt I just saw this comment but I fixed this in another way, simply reordering the migrations won't work though because eg `00033-add-application-uuid` may already have been run and renaming it to `00034` means it'll be run again, and if `00031` was already run then just renaming `00030-add-release-version` to `00031` won't make it run. As it is the order of `00030-add-release-version` thankfully doesn't matter in this case so I was able to just bump it to `00034` without issue

dfunckt:
Out of multiple migrations with the same number only one of them will run @Page-, so you can be sure it's going to be their first time after renaming them. The catch is determining which one didn't run and can be safely renamed. In practice though, this matters only a little, assuming openBalena is the only user/client of the core API because it hasn't updated to a new API version for a while.

PR #501, opened 2020-10-28
cpu architecture: add synonymous form to device type

xginn8:
Change-type: patch

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -559,7 +559,9 @@ Fact type: device type has name (Auth)
 	Necessity: each device type has exactly one name (Auth)
 
 Fact type: device type is of cpu architecture
-	Necessity: each device type is of exactly one cpu architecture
+	Synonymous form: device type supports cpu architecture
+	Synonymous form: cpu architecture is supported by device type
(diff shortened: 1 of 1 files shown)


src/balena.sbvr
```diff
@@ -558,8 +558,10 @@ Fact type: device type has slug
 Fact type: device type has name (Auth)
 	Necessity: each device type has exactly one name (Auth)
 
-Fact type: device type is of cpu architecture
-	Necessity: each device type is of exactly one cpu architecture
+Fact type: device type supports cpu architecture
+	Synonymous form: device type is of cpu architecture
+	Synonymous form: cpu architecture is supported by device type
+	Necessity: each device type supports exactly one cpu architecture
```
Page-, line 564: Go with
```suggestion
Fact type: device type is of cpu architecture
	Synonymous form: device type supports cpu architecture
	Synonymous form: cpu architecture is supported by device type
	Necessity: each device type supports exactly one cpu architecture
```
for now, changing the `Fact type` version will affect the db column name so it's better to leave for later

xginn8: done, thanks

Page- (approved):

PR #504, opened 2020-11-03
Add supervisor to v3 state model

richbayliss:
As part of the work to allow devices to run multiple applications, this PR adds systemApps to the device state for v2.  Currently on the supervisor will be returned.


dfunckt (commented):
Code-wise looks solid. I'd test with current OS releases to ensure the new attribute in the response doesn't trip the Supervisor (I don't expect it will though).

I'm not sure about the choice of specifying system apps via an env var -- fells fragile, and also means initialisation of the API must happen in two steps: create the apps, set the env vars and relaunch the server. I would prefer it if whether an app is a system one to be a property of the model.

src/lib/auth.ts
```diff
@@ -68,7 +68,7 @@ export const DEVICE_API_KEY_PERMISSIONS = [
 	'resin.device_type.read?describes__device/canAccess()',
 	`resin.device.read?${matchesActor}`,
 	`resin.device.update?${matchesActor}`,
-	'resin.application.read?owns__device/canAccess() or depends_on__application/any(a:a/owns__device/canAccess()) or (is_host eq true and is_for__device_type/any(dt:dt/describes__device/canAccess()))',
+	'resin.application.read?owns__device/canAccess() or depends_on__application/any(a:a/owns__device/canAccess()) or (((is_host eq true) or (is_public eq true)) and is_for__device_type/any(dt:dt/describes__device/canAccess()))',
```
Page-, line 71:
```suggestion
	'resin.application.read?owns__device/canAccess() or depends_on__application/any(a:a/owns__device/canAccess()) or ((is_host eq true or is_public eq true) and is_for__device_type/any(dt:dt/describes__device/canAccess()))',
```

src/features/releases/hooks/delete-checks.ts
```diff
@@ -0,0 +1,31 @@
+import { sbvrUtils, hooks, permissions } from '@balena/pinejs';
```
Page-, line 1: This hook should really be in a feature that would create the `should_be_managed_by__release` relationship, presumably something supervisor related

src/features/devices/hooks/should-be-managed-by--release.ts
```diff
@@ -0,0 +1,34 @@
...
+					},
+					should_be_managed_by__release: null,
+				},
+			},
+		});
+
+		/**
+		 * These device(s) should have a valid release assigned based on their current
+		 * running supervisor version, if we have one...
+		 */
```
Page-, line 32: It looks like you're missing code to actually use the GET result?

src/features/devices/hooks/should-be-managed-by--release.ts
```diff
@@ -0,0 +1,34 @@
+import { hooks } from '@balena/pinejs';
```
Page-, line 1: This hook also looks like it should be in the feature that would create the should_be_managed_by__release relationship

src/features/device-state/utils.ts
```diff
@@ -33,10 +33,14 @@ export const serviceInstallFromImage = (
...
-	if (typeof image.is_a_build_of__service === 'object') {
-		id = image.is_a_build_of__service.__id;
+	if (Array.isArray(image.is_a_build_of__service)) {
+		id = image.is_a_build_of__service[0].id;
 	} else {
-		id = image.is_a_build_of__service;
+		id = image.is_a_build_of__service.__id;
+	}
+
+	if (isNaN(id)) {
```
Page-, line 42: Do you want `isNaN` or `Number.isNaN` here?

src/balena.sbvr
```diff
@@ -443,6 +445,10 @@ Fact type: application has uuid
 	Necessity: each application has exactly one uuid.
 	Necessity: each uuid is of exactly one application.
 	Necessity: each application has a uuid that has a Length (Type) that is equal to 32.
+Fact type: application is public
+Fact type: application has install type
+	Necessity: each application has at most one install type
```
Page-, line 450: Should this be
```suggestion
	Necessity: each application has exactly one install type
```
since there's a default?

src/migrations/00039-default-application-install-type.sql
```diff
@@ -0,0 +1,3 @@
+-- migrate to add a new install type value to the application model...
```
Page-, line 1: Any reason for 3 separate migration files instead of just one?

src/features/device-state/routes/state.ts
```diff
@@ -14,6 +14,8 @@ import {
 } from '../utils';
 import { sbvrUtils, errors } from '@balena/pinejs';
 import { events } from '..';
+import * as conf from '../../../lib/config';
```
Page-, line 17:
```suggestion
import { EXTRA_CONTAINERS } from '../../../lib/config';
```

Page- (changes requested):
I still have more to look through, the request changes is about what appears to be missing code from one of the hooks

src/balena.sbvr
```diff
@@ -733,3 +742,6 @@ Rule: It is necessary that each release1 that has a status that is equal to "suc
 Rule: It is necessary that each release that should be running on a device, has a status that is equal to "success" and belongs to an application1 that the device belongs to.
 Rule: It is necessary that each release that should be running on an application, has a status that is equal to "success" and belongs to the application.
 Rule: It is necessary that each release that contains at least 2 images, belongs to an application that has an application type that supports multicontainer.
+
+-- TODO: This rule should be updated to use the device type arch once that is fully modeled.
+Rule: It is necessary that each device that should be managed by a release, should be managed by a release that has a status that is equal to "success" and belongs to an application that is public and is not host and is for a device type that describes the device.
```
xginn8, line 747: this has been updated in the closed API to reflect recent `cpu architecture` changes, be sure to backfill that!

src/lib/auth.ts
```diff
@@ -77,7 +77,7 @@ export const DEVICE_API_KEY_PERMISSIONS = [
 	`resin.device_tag.create?device/any(d:d/${matchesActor})`,
 	`resin.device_tag.update?device/any(d:d/${matchesActor})`,
 	'resin.application_config_variable.read?application/canAccess()',
-	'resin.release.read?should_be_running_on__device/canAccess() or belongs_to__application/canAccess()',
+	'resin.release.read?should_be_running_on__device/canAccess() or should_manage__device/canAccess() or belongs_to__application/canAccess()',
```
richbayliss, line 80: do we _need_ it? I think the original permissions should allow this by the fact the supervisor application should be public.

src/features/device-state/routes/state.ts
```diff
@@ -174,121 +347,133 @@ export const state: RequestHandler = async (req, res) => {
...
+											name: 'asc',
+										},
+									},
+									should_be_running__release: releaseExpand,
+								},
+								$filter: {
+									uuid: {
+										$in: conf.EXTRA_CONTAINERS,
+									},
+									should_be_running__release: {
```
xginn8, line 420: this is likely insufficient, as it'd limit supervisors in the state endpoint to only those that are the current latest release in their respective apps.

xginn8: immediately we can work around this by changing this line as follows i believe:
```suggestion
									owns__release: {
```

xginn8: @Page- specifically here i don't think there's a clean way in the same query to fetch both:

1. the release linked to a device in the case of the supervisor (i.e., whatever we find at `should_be_managed_by__release`), AND
1. the latest release in the case of a hostapp extension/hostapp

am i missing something, or does this in fact need to be split into two queries and merged back into a single object?

src/balena.sbvr
xginn8: updated

src/features/device-state/routes/state.ts
xginn8: updated

toochevere:
> Code-wise looks solid. I'd test with current OS releases to ensure the new attribute in the response doesn't trip the Supervisor (I don't expect it will though).
> 
> I'm not sure about the choice of specifying system apps via an env var -- fells fragile, and also means initialisation of the API must happen in two steps: create the apps, set the env vars and relaunch the server. I would prefer it if whether an app is a system one to be a property of the model.

Was this actually resolved yet?  It seems not

src/features/device-state/utils.ts
toochevere: So, given that `isNaN()` will trap a wider range of conditions than `Number.isNaN()` it seems that we want `isNaN()`.  I'll give it a day or two and then mark this one as resolved unless I hear otherwise.

src/features/devices/hooks/should-be-managed-by--release.ts
toochevere: So, it seems to me that the intention of this code is, at the moment of PATCH, if `should_be_managed_by__release` has not been provided and the device needs `should_be_managed_by__release`, then patch it with the current supervisor version.  

If I am correct, then something like this is what we need to add?
```
const devices = await args.api.get(...
if (devices.length > 0) { 
  request.values.should_be_managed_by__release = devices[0].supervisor_version;
}
```

src/lib/auth.ts
toochevere: Is @richbayliss correct?  Can we revert this or should we leave it?

src/features/devices/hooks/should-be-managed-by--release.ts
xginn8: > On a side note, I'm not sure why this isn't just handled by a migration of all existing data that lacks the desired attribute, rather than doing it at the point of PATCHing?

@toochevere this value can't be known a priori due to how flasher images work (first creating a device stub, then populating it with appropriate data during the first boot from non-flasher media).

`supervisor_version` is what the supervisor itself reports back as a string (current state). We then need to look up the corresponding target state and set the value accordingly, so it requires a second lookup for which release is architecture-compatible and matches the version the device has just checked in with. See https://github.com/balena-io/balena-api/blob/ed7ced9ed0b2f4115aab72da6022bb49b2804da8/src/models/hooks/resources/device.ts#L351-L367 for more.

src/features/devices/hooks/should-be-managed-by--release.ts
```diff
@@ -26,9 +40,72 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
+			args.api,
+			device.supervisor_version,
+			device.is_for__device_type.is_of__cpu_architecture.id,
+		);
+
+		if (!supervisorRelease) {
+			return;
+		}
+
+		request.values.should_be_managed_by__release = supervisorRelease.id;
```
toochevere, line 57: @xginn8 Taking a stab at it... How does this look?

src/features/devices/hooks/should-be-managed-by--release.ts
toochevere: Need some guidance about what this means (in terms of a practical resolution) ...fyi ... @xginn8 (thanks)

src/features/releases/hooks/delete-checks.ts
toochevere: Need some guidance about what this means (in terms of a practical resolution) ...fyi ... @xginn8 (thanks)

src/features/devices/hooks/should-be-managed-by--release.ts
```diff
@@ -0,0 +1,111 @@
...
+		) {
+			return;
+		}
+
+		/**
+		 * These device(s) should have a valid release assigned based on their current
+		 * running supervisor version, if we have one...
+		 */
+
+		const [device] = await args.api.get({
```
xginn8, line 21: the comment above seems to indicate this type of PATCH could include multiple devices, but here is handled as a single device. is that possible here?

src/balena.sbvr
xginn8: sorted.

src/lib/auth.ts
xginn8: i believe this is unnecessary, so i'm dropping it.

src/features/devices/hooks/should-be-managed-by--release.ts
toochevere: yep, good point.  So, since there are multiple devices, are we guaranteed that they will all need the same supervisor release id for `should_be_managed_by__release`?  In other words, each device might have a different `device.supervisor_version` and a different `device.is_for__device_type.is_of__cpu_architecture.id`, right?  So therefore, I can't just set `request.values.should_be_managed_by__release = X` and patch them all to the same release id.

src/features/releases/hooks/delete-checks.ts
toochevere: @xginn8 just some clarification... so instead of PRERUNning when DELETing a `release`, this code will now PRERUN when DELETing a `device`?  :point_left: proceeding with that understanding til I hear otherwise.

src/features/devices/hooks/should-be-managed-by--release.ts
toochevere: @xginn8 hey Matt, just pinging on this :point_up_2: 

Since there are multiple devices, the requirements around `should_be_managed_by__release` might be different for each one, right?  We can't just `request.values.should_be_managed_by__release = X` but rather have to call `api.patch(...)` for each device and patch it according to the correct release id.  Am I on the right track?

xginn8: Here's the current implementation in our closed source API:

https://github.com/balena-io/balena-api/blob/2efc36726e9f9ae93f40e8af5eec193dcac5d7c3/src/models/hooks/resources/device.ts#L347-L363
https://github.com/balena-io/balena-api/blob/2efc36726e9f9ae93f40e8af5eec193dcac5d7c3/src/models/hooks/resources/device.ts#L665-L733

Basically we map over devices grouped by CPU architecture and PATCH the values N architecture times instead of M devices times.

I think that should help clear up the ambiguity, sorry for my tardiness here!!

src/migrations/00046-host-multipapp.sql
```diff
@@ -0,0 +1,11 @@
+-- migrate to add a new public flag value to the application model...
+ALTER TABLE "application"
+ADD COLUMN IF NOT EXISTS "is public" INT NOT NULL DEFAULT 0;
+
+-- migrate to add a new install type value to the application model...
+ALTER TABLE "image"
+ADD COLUMN IF NOT EXISTS "install type" VARCHAR(255) NULL DEFAULT 'supervised';
```
xginn8, line 7: this needs to be dropped

src/balena.sbvr
```diff
@@ -276,6 +276,8 @@ Term: version
 Term: vpn address
 	Concept Type: Short Text (Type)
 
+Term: install type
```
xginn8, line 279: this needs to be dropped

test/fixtures/15-apps/release_tags.json
```diff
@@ -0,0 +1,26 @@
+{
```
xginn8, line 1: these can be removed now

toochevere:
- [ ] Need to move breaking changes to v3.

automation/fasttest.sh
```diff
@@ -71,7 +71,7 @@ if [ -z "$db_id" ] || [ -z "$redis_id" ] || [ -z "$loki_id" ] || [ -z "$api_id"
 	loki_id=$(runloki '-p 3100:3100')
 	echo $loki_id >>"$CONFIG_FILE"
 
-	api_id=$(runapi $IMAGE_NAME $db_id $redis_id $loki_id " -p 9228:9229 -v $(pwd):/usr/src/app")
+	api_id=$(runapi $IMAGE_NAME $db_id $redis_id $loki_id " -p 9229:9229 -v $(pwd):/usr/src/app")
```
xginn8, line 74: is this necessary for this PR?

package.json
```diff
@@ -14,6 +14,7 @@
   "scripts": {
     "fasttest": "IMAGE_NAME=test-open-balena-api-fast ./automation/fasttest.sh",
     "fasttest-cleanup": "IMAGE_NAME=test-open-balena-api-fast ./automation/fasttest.sh --teardown",
+    "fasttest-debug": "IMAGE_NAME=test-open-balena-api-fast ./automation/fasttest.sh --profile",
```
xginn8, line 17: is this necessary for this PR?

src/migrations/00047-host-multipapp.sql
```diff
@@ -0,0 +1,7 @@
+-- migrate to add a new public flag value to the application model...
+ALTER TABLE "application"
```
xginn8, line 2: can be dropped

src/features/device-state/routes/state.ts
```diff
@@ -25,6 +26,163 @@ export const varListInsert = (varList: EnvVarList, obj: Dictionary<string>) => {
...
+	apps: Dictionary<Partial<App>>;
+};
+
+async function buildAppFromRelease(
+	device: AnyObject,
+	application: AnyObject,
+	release: AnyObject,
+	config: Dictionary<string>,
+): Promise<App> {
+	let composition: any = {};
```
Page-, line 65: Please type this better than `any`, at the barest minimum it should be `AnyObject`

src/features/device-state/routes/state.ts
```diff
@@ -25,6 +26,163 @@ export const varListInsert = (varList: EnvVarList, obj: Dictionary<string>) => {
...
+	const services: App['services'] = {};
+
+	// Parse the composition to forward values to the device
+	if (_.isObject(release.composition)) {
+		composition = release.composition;
+	} else {
+		try {
+			composition = JSON.parse(release.composition);
+		} catch (e) {
```
Page-, line 74:
```suggestion
		} catch {
```

src/features/device-state/routes/state.ts
```diff
@@ -25,6 +26,163 @@ export const varListInsert = (varList: EnvVarList, obj: Dictionary<string>) => {
...
+): Promise<App> {
+	let composition: any = {};
+	const services: App['services'] = {};
+
+	// Parse the composition to forward values to the device
+	if (_.isObject(release.composition)) {
+		composition = release.composition;
+	} else {
+		try {
+			composition = JSON.parse(release.composition);
```
Page-, line 73: When does this ever need to be parsed?

src/features/device-state/routes/state.ts
```diff
@@ -25,6 +26,163 @@ export const varListInsert = (varList: EnvVarList, obj: Dictionary<string>) => {
...
+		const image = ipr.image[0];
+
+		const serviceInstall = (() => {
+			const si = serviceInstallFromImage(device, image);
+
+			if (si != null) {
+				return si;
+			}
+
+			if (image.is_a_build_of__service[0] == null) {
```
Page-, line 90: What is this check and how is it different to `si == null`?

src/features/device-state/routes/state.ts
```diff
@@ -25,6 +26,163 @@ export const varListInsert = (varList: EnvVarList, obj: Dictionary<string>) => {
...
+					`Could not find service install for device: '${
+						device.uuid
+					}', image: '${image?.id}', service: '${JSON.stringify(
+						image?.is_a_build_of__service,
+					)}', service installs: '${JSON.stringify(device.service_install)}'`,
+				);
+			}
+
+			/**
+			 * Return a fake service install as this is what we will use for MultiApp
```
Page-, line 101: Why do we have fake service installs rather than real ones?

src/features/device-state/routes/state.ts
```diff
@@ -25,6 +26,163 @@ export const varListInsert = (varList: EnvVarList, obj: Dictionary<string>) => {
...
+		// of bandwidth (a null contract is the same as
+		// the lack of a contract field)
+		if (image.contract != null) {
+			services[service.id].contract = image.contract;
+		}
+
+		if (
+			composition != null &&
+			composition.services != null &&
+			composition.services[service.service_name] != null
```
Page-, line 160:
```suggestion
			composition?.services?.[service.service_name] != null
```

src/features/device-state/routes/state.ts
```diff
@@ -25,6 +26,163 @@ export const varListInsert = (varList: EnvVarList, obj: Dictionary<string>) => {
...
+		};
+		// Don't send a null contract as this is a waste
+		// of bandwidth (a null contract is the same as
+		// the lack of a contract field)
+		if (image.contract != null) {
+			services[service.id].contract = image.contract;
+		}
+
+		if (
+			composition != null &&
```
Page-, line 158: Can composition ever be null?

src/features/device-state/routes/state.ts
```diff
@@ -25,6 +26,163 @@ export const varListInsert = (varList: EnvVarList, obj: Dictionary<string>) => {
...
+	});
+
+	return {
+		releaseId: release.id,
+		releaseVersion: release.release_version,
+		commit: release.commit,
+		name: application.app_name,
+		services,
+		networks: composition?.networks || {},
+		volumes: composition?.volumes || {},
```
Page-, line 180: Should these be
```suggestion
		networks: composition?.networks ?? {},
		volumes: composition?.volumes ?? {},
```
?

src/features/device-state/routes/state.ts
```diff
@@ -395,3 +419,195 @@ export const state: RequestHandler = async (req, res) => {
...
+					},
+					application_environment_variable: {
+						$select: ['name', 'value'],
+						$orderby: {
+							name: 'asc',
+						},
+					},
+				},
+			},
+		});
```
Page-, line 500: Can you make this part of the main query that gets the rest of the state info please rather than running two separate queries in two separate transactions

src/features/device-state/routes/state.ts
```diff
@@ -395,3 +419,195 @@ export const state: RequestHandler = async (req, res) => {
...
+		: await buildAppFromRelease(device, userAppFromApi, release, config);
+};
+
+const getSupervisorAppForState = async (
+	req: Request,
+	device: any,
+	config: Dictionary<string>,
+) => {
+	if (device.should_be_managed_by__release.length === 0) {
+		return undefined;
```
Page-, line 475:
```suggestion
		return;
```

src/features/device-state/routes/state.ts
```diff
@@ -395,3 +419,195 @@ export const state: RequestHandler = async (req, res) => {
...
+							name: 'asc',
+						},
+					},
+				},
+			},
+		});
+	});
+
+	if (!fromApi) {
+		return undefined;
```
Page-, line 504:
```suggestion
		return;
```

src/features/device-state/routes/state.ts
```diff
@@ -395,3 +419,195 @@ export const state: RequestHandler = async (req, res) => {
...
+
+	return await buildAppFromRelease(
+		device,
+		fromApi,
+		device.should_be_managed_by__release[0],
+		config,
+	);
+};
+
+const getDependent = async (device: any) => {
```
Page-, line 517: Can you add a better typing than `any` please

src/features/device-state/routes/state.ts
```diff
@@ -395,3 +419,195 @@ export const state: RequestHandler = async (req, res) => {
...
+		apps: {} as AnyObject,
+		devices: {} as AnyObject,
+	};
+
+	const depAppCache: Dictionary<{
+		release?: AnyObject;
+		application_environment_variable: Array<{
+			name: string;
+			value: string;
+		}>;
```
Page-, line 530: I think this should be
```suggestion
		application_environment_variable: Array<Pick<ApplicationEnvironmentVariable, 'name' | 'value'>>;
```

src/features/device-state/routes/state.ts
```diff
@@ -395,3 +419,195 @@ export const state: RequestHandler = async (req, res) => {
...
+		const environment: Dictionary<string> = {};
+		if (ipr != null) {
+			varListInsert(ipr.image_environment_variable, environment);
+		}
+
+		varListInsert(application_environment_variable, environment);
+		if (
+			svcInstall != null &&
+			svcInstall.service != null &&
+			svcInstall.service[0] != null
```
Page-, line 587:
```suggestion
			svcInstall?.service?.[0] != null
```

src/features/devices/hooks/should-be-managed-by--release.ts
```diff
@@ -0,0 +1,137 @@
...
+	 */
+	async PRERUN(args) {
+		if (args.request.values.supervisor_version != null) {
+			await sbvrUtils.getAffectedIds(args).then(async (ids) => {
+				await setSupervisorReleaseResource(
+					args.api,
+					ids,
+					args.request.values.supervisor_version,
+				);
+			});
```
Page-, line 17:
```suggestion
			const ids = await sbvrUtils.getAffectedIds(args);
			await setSupervisorReleaseResource(
				args.api,
				ids,
				args.request.values.supervisor_version,
			);
```

Page- (commented):
This PR is made very hard to review by the fact that it has merged refactoring and functional changes in a single commit so it's very hard to see what is just code that has been moved around and code that has had it's actual functionality altered - please avoid doing that in future

src/features/device-state/routes/state.ts
toochevere: resolved in other PR

src/features/device-state/routes/state.ts
toochevere: Resolved in other PR

src/features/device-state/routes/state.ts
xginn8: I think this is a vestigial piece of the original implementation for system apps like ingrid -- before true multiapp, we weren't able to properly associate a service_install with a device (maybe?)

src/features/device-state/routes/state.ts
toochevere: we use it further down
```
	return {
		releaseId: release.id,
		releaseVersion: release.release_version,
		commit: release.commit,
		name: application.app_name,
		services,
		networks: composition?.networks || {},
		volumes: composition?.volumes || {},
		appId: application.id,
		uuid: application.uuid,
	};
```

src/features/device-state/routes/state.ts
toochevere: fixed in prev PR

src/features/device-state/routes/state.ts
toochevere: I need some guidance on this bit of code... it seems that it's purpose is to guarantee that `serviceInstall` always defined so that we can proceed with the rest of the code.  The "fake service" bit only runs if a real service install was not retrieved.   Do I remove it?  

This is especially confusing given that `utils.ts/serviceInstallFromImage` has changed significantly.  

We went from this:
```
	if (typeof image.is_a_build_of__service === 'object') {
		id = image.is_a_build_of__service.__id;
	} else {
		id = image.is_a_build_of__service;
	}
```
to this
```
	if (Array.isArray(image.is_a_build_of__service)) {
		id = image.is_a_build_of__service[0].id;
	} else {
		id = image.is_a_build_of__service.__id;
	}
```
implying that the nature of `is_a_build_of__service` has changed.   Also, it now looks in the current code like an exception is possible if `is_a_build_of__service` is undefined.  
@xginn8 gonna need  your advice on how to proceed.

src/features/device-state/routes/state.ts
toochevere: See my comment below.  I need some advice regarding this whole block.

src/features/device-state/routes/state.ts
xginn8: i think service installs are created typically for all required services x devices, but only in the case of system apps could they not exist since it was kind of shoehorned into the model. i think it's best to remove, but i'm not 100% confident on that either.

toochevere: After more investigation this is necessary for the supervisor to appear to be an app.  The only alternative would be to create a service install when we PATCH to link the supervisor.

src/features/device-state/routes/state.ts
toochevere: > What is this check and how is it different to si == null?

The difference is that if there were _no services installs at all_ then we would hit this error.  It is possible for `si` to be null in the case of the supervisor.

src/migrations/00047-host-multipapp.sql
```diff
@@ -0,0 +1,3 @@
+-- migrate to add a should be managed by-release relationship to the devices...
+ALTER TABLE "device"
+ADD COLUMN IF NOT EXISTS "should be managed by-release" INT NULL;
```
xginn8, line 3: might be worth adding an index here as well, since it'll be a hot path..

src/features/devices/hooks/should-be-managed-by--release.ts
```diff
@@ -0,0 +1,136 @@
+import { hooks, sbvrUtils, permissions } from '@balena/pinejs';
```
Page-, line 1: Can you move this to a supervisorapp feature equivalent to the hostapp feature, cc @xginn8

src/features/device-state/routes/state.ts
Page-: I think the correct answer is indeed to add those service installs @toochevere since it moves towards the future where it's a real app rather than hacking around the fact that it's not quite a true app yet

src/features/device-state/routes/state.ts
Page-: I think the real issue is that it should not be null for the supervisor

src/features/device-state/routes/state.ts
Page-: I meant more specifically when is it ever not already parsed - I'd expect it to always be a parsed object at this point and not a string

Page-: Although with the refactoring being split out it's clear this was just an existing wart so it can be left as-is, but I do expect it to be unnecessary and we can clean it up in future

src/features/device-state/routes/state.ts
```diff
@@ -26,16 +27,22 @@ export const varListInsert = (varList: EnvVarList, obj: Dictionary<string>) => {
 	});
 };
 
-export type App = {
+export interface AppV2 {
 	name: string;
 	commit: string;
 	releaseId: number;
+	releaseVersion: string;
```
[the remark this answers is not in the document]
toochevere: Getting rid of that, and adding in some typing to make sure we don't accidentally change V2 response.

src/features/devices/hooks/should-be-managed-by--release.ts
toochevere: Done by Matt

src/features/device-state/routes/state.ts
toochevere: Looking at it.... no resolved yet.

src/features/device-state/routes/state.ts
```diff
@@ -287,6 +330,29 @@ const stateQuery = _.once(() =>
 						},
 					},
 				},
+				should_be_managed_by__release: {
+					...releaseExpand,
+					$expand: {
+						...releaseExpand.$expand,
```
dfunckt, line 336: I understand why you're doing this but there's a lot of overlap and overriding of attributes here and IMO this is prone to bugs. You primarily want to avoid writing the same thing each time but at the use site you also want to ensure that you can extend it safely without clobbering over attributes.

I believe converting `releaseExpand` to a function that returns a PineOptions (not sure about the exact type) and allows for parameterising `$select` and top-level `$expand` is much more readable and less error prone. You'd then use it like so:

```js
should_be_managed_by__release: {
  ...releaseExpand(['any', 'extra', '$select', 'attributes'], {
    belongs_to__application: { ... }
  })
}
```

src/features/device-state/routes/state.ts
```diff
@@ -301,28 +367,83 @@ export const stateV2: RequestHandler = async (req, res) => {
...
+		const userApp = getUserAppForState(device, config);
+		if (userApp) {
+			const userAppV2 = {
+				commit: userApp.commit,
+				name: userApp.name,
+				releaseId: userApp.releaseId,
+				networks: userApp.networks,
+				services: userApp.services,
+				volumes: userApp.volumes,
+			} as AppV2;
```
dfunckt, line 385:
```suggestion 
			const userAppV2: AppV2 = {
  				commit: userApp.commit,
  				name: userApp.name,
  				releaseId: userApp.releaseId,
  				networks: userApp.networks,
  				services: userApp.services,
  				volumes: userApp.volumes,
  			};
```
Typing it like this ensures the value you assign *is* the type you expect, whereas the `as` form coerces the compiler to think that whatever value you assigned is that type, which is not useful here.

src/features/device-state/routes/state.ts
```diff
@@ -301,28 +367,83 @@ export const stateV2: RequestHandler = async (req, res) => {
...
+	} catch (err) {
+		if (handleHttpErrors(req, res, err)) {
+			return;
+		}
+		captureException(err, 'Error getting device state', { req });
+		res.sendStatus(500);
+	}
+};
+
+export const stateV3: RequestHandler = async (req, res) => {
```
dfunckt, line 410: There's a lot of repetition between the two handlers. I think it's best to extract the surrounding common bits into a single function and pass most of the `try` block as a callback. Something like this should work:

```ts
const makeStateHandler = (cb: (uuid, apiKey) => Promise<AnyObject>) => async (req, res) => {
	...

	try {
		const result = await cb(uuid, apikey);
		res.json(...)
	} catch {
		...
	}
}

const stateV2: RequestHandler = makeStateHandler(() => {
	...
});
```

This will still preserve quite a bit of repetition but it's the first obvious step towards DRY'ing this up a bit.

All that said, I don't feel strongly about doing this now -- feel free to merge but please make sure to return and improve this soon after.

src/features/device-state/routes/state.ts
```diff
@@ -362,24 +483,60 @@ const getConfig = (device: AnyObject) => {
...
 				networks: {},
 				volumes: {},
-		  }
+		  } as AppV3)
 		: buildAppFromRelease(device, userAppFromApi, release, config);
 };
 
-const getDependent = (device: AnyObject): Dependent => {
+const getSupervisorAppForState = async (
+	device: any,
```
dfunckt, line 506: This should at a minimum specify the attributes it requires:

```suggestion
	device: AnyObject & { should_be_managed_by__release: Array<{ belongs_to__application: ... }> }
```

src/features/device-state/routes/state.ts
```diff
@@ -362,24 +483,60 @@ const getConfig = (device: AnyObject) => {
...
+	fromApi.id =
+		device.should_be_managed_by__release[0].belongs_to__application[0].id;
+	fromApi.app_name =
+		device.should_be_managed_by__release[0].belongs_to__application[0].app_name;
+	fromApi.uuid =
+		device.should_be_managed_by__release[0].belongs_to__application[0].uuid;
+	fromApi.application_config_variable =
+		device.should_be_managed_by__release[0].belongs_to__application[0].application_config_variable;
+	fromApi.application_environment_variable =
+		device.should_be_managed_by__release[0].belongs_to__application[0].application_environment_variable;
```
dfunckt, line 526: Can you please make local references for the release and application? It would really help with readability.

dfunckt (commented):
Didn't get to review it all -- will try to resume later today.

src/features/device-state/routes/state.ts
toochevere: The previous refactoring (previous to the current PR's) went a long way toward avoiding WET code.  They are somewhat similar, but I felt that trying to coalesce them any further would lead to code that is hard to read.  We can look at this again once this PR is done.

src/features/device-state/routes/state.ts
toochevere: I refactored it into separate objects that can be composed as necessary.

src/features/supervisor-app/hooks/supervisor-app.ts
```diff
@@ -180,7 +208,17 @@ async function getSupervisorReleaseResource(
 	});
 }
 
-async function setSupervisorReleaseResource(
+function createSupervisorServiceInstall(
+	rootApi: sbvrUtils.PinejsClient,
+	deviceIds: number[],
+	superisorReleaseServiceId: number,
```
Page-, line 214:
```suggestion
	supervisorReleaseServiceId: number,
```

src/features/supervisor-app/hooks/supervisor-app.ts
```diff
@@ -1,12 +1,13 @@
 import * as semver from 'balena-semver';
 import * as _ from 'lodash';
-
+import { createReleaseServiceInstalls } from '../../ci-cd/hooks/service-installs';
```
Page-, line 3: This is a big red flag to me that you're importing a function from the hooks of another feature and makes me think that it should be done there instead. And checking the code I see no reason why the behavior can't be handled by adding a hook there that follows the existing
```ts
			if (request.values.should_be_running__release != null) {
				await createReleaseServiceInstalls(api, affectedIds, {
					id: request.values.should_be_running__release,
				});
			}
```
style but instead does
```ts
			if (request.values.should_be_managed_by__release!= null) {
				await createReleaseServiceInstalls(api, affectedIds, {
					id: request.values.should_be_managed_by__release,
				});
			}
```
ie same behavior with the different key

src/features/device-state/routes/state.ts
toochevere: I changed to this AnyObject, but going beyond that did not feel like it made sense for this simple method which only touches a couple of properties and breaks with the existing pattern in this file.  If we want to start typing for use of the properties, I think it would be better in a refactoring PR to do so consistently.

src/features/supervisor-app/hooks/supervisor-app.ts
toochevere: Resolving because these changes were rolled back.

src/features/supervisor-app/hooks/supervisor-app.ts
toochevere: Resolving because these changes were rolled back.

src/features/device-state/routes/state.ts
toochevere: Changes made.. resolving since Akis is out.

src/features/device-state/routes/state.ts
toochevere: resolving since Akis is out.

src/features/device-state/routes/state.ts
toochevere: Changes made.. resolving since Akis is out.

src/features/device-state/routes/state.ts
toochevere: resolving since Akis is out.

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -224,3 +224,138 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
+		}
+
+		// Create supervisor serice installs when device checks in with it's initial version
+		if (request.values.supervisor_version) {
+			await createSupervisorServiceInstallsFromVersion(
+				api,
+				affectedIds,
+				request.values.supervisor_version,
+			);
+			return;
```
Page-, line 242: This `return` would break things if setting `supervisor_version` and `should_be_managed_by__release` in the same request

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -224,3 +224,138 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
+		if (affectedIds.length === 0) {
+			return;
+		}
+
+		// Create supervisor serice installs when device checks in with it's initial version
+		if (request.values.supervisor_version) {
+			await createSupervisorServiceInstallsFromVersion(
+				api,
+				affectedIds,
+				request.values.supervisor_version,
```
Page-, line 240: I think you can skip this as we should never need to create service installs for the currently running release? I think it should only create service installs for the target state of `should_be_managed_by__release` for supervisor releases in the same way it only creates for the target state of `should_be_running__release` for app releases (and ignores `is_running__release`)

src/features/ci-cd/hooks/service-installs.ts
toochevere: Never mind.. it works as you said.

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -224,3 +224,19 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
 		}
 	},
 });
+
+hooks.addPureHook('PATCH', 'resin', 'device', {
```
thgreasi, line 228: We might also need to change the device move implementation as well (earlier in this file), since moving a device atm deletes all service installs & creates the ones of the new app.
See: https://github.com/balena-io/open-balena-api/blob/c40079c5fa6cef1b5744493b96e7f2fcf3ccc594/src/features/ci-cd/hooks/service-installs.ts#L164-L174

In order to not lose the services installs created by this hook, we probably need to avoid deleting supervisor release related service installs, right?

toochevere: Yeah, I had thought about that and was going to do it in another PR.  @Page- What is the best way to do this?

src/features/device-state/routes/state.ts
```diff
@@ -362,24 +488,56 @@ const getConfig = (device: AnyObject) => {
...
+
+	const fromApi: AnyObject = {};
+	fromApi.id = deviceManagedByApplication.id;
+	fromApi.app_name = deviceManagedByApplication.app_name;
+	fromApi.uuid = deviceManagedByApplication.uuid;
+	fromApi.application_config_variable =
+		deviceManagedByApplication.application_config_variable;
+	fromApi.application_environment_variable =
+		deviceManagedByApplication.application_environment_variable;
+	fromApi.should_be_running__release = deviceManagedByRelease;
```
Page-, line 529: Can you create this as an object literal in one go rather than creating an empty object and then assigning props, creating it as an object literal will allow dropping the `: AnyObject` and allow inferring a better type automatically, it also makes it much easier for humans to see the shape of the object we're creating

src/features/device-state/routes/state.ts
```diff
@@ -362,24 +488,56 @@ const getConfig = (device: AnyObject) => {
...
+
+	const supervisorApp = await buildAppFromRelease(
```
Page-, line 531: This `await` is unnecessary right? Can you remove it please
```suggestion
	const supervisorApp = buildAppFromRelease(
```

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -155,7 +197,7 @@ hooks.addPureHook('POST', 'resin', 'device', {
 });
 
 hooks.addPureHook('PATCH', 'resin', 'device', {
-	POSTRUN: async (args) => {
+	PRERUN: async (args) => {
```
Page-, line 200: Why did you change this to a `PRERUN`?

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -164,14 +206,38 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
-				},
-			});
+				});
+
+			await actOnAppServiceInstalls(
+				args.api,
+				args.request.values.belongs_to__application,
+				affectedIds,
+				deleteHandler,
+			);
```
Page-, line 229: I'd rather see this done in a single request as it's acting on existing service installs, and avoid the extra work that is only required in the case we need to find missing service installs
```ts
				api.delete({
					resource: 'service_install',
					options: {
						$filter: {
							device: { $in: deviceIds },
							installs__service: {
								$any: {
									$alias: 's',
									$expr: {
										s: { 
											application: {
												$any: {
													$alias: 'a',
													$expr: { a: { id: appId } },
												},
											},
										},
									},
								},
							},
						},
					},
				});

File: src/features/ci-cd/hooks/service-installs.ts (hunk 50 above)
Comment (toochevere) [reply src/features/ci-cd/hooks/service-installs.ts:200]: We need to get the existing appId _before_ we make any changes in order to delete service installs with from that app only.  My understanding is that once we get to POSTRUN, it's too late for that.

File: src/features/ci-cd/hooks/service-installs.ts (hunk 51 above)
Comment (toochevere) [reply src/features/ci-cd/hooks/service-installs.ts]: I re-wrote the code to create them directly without any comparison with existing service installs, since in this case there won't be any.

File: src/features/device-state/routes/state.ts [hunk 52]
```diff
@@ -301,28 +372,83 @@ export const stateV2: RequestHandler = async (req, res) => {
...
+			appsForState[userApp.appId] = userAppV2;
+		}
+
+		const local: LocalStateV2 = {
+			name: device.device_name ?? '',
+			config,
+			apps: appsForState,
+		};
+
+		const dependent = await getDependent(device);
```
Page-, line 400:
```suggestion
		const dependent = getDependent(device);
```

src/features/device-state/routes/state.ts
```diff
@@ -301,28 +372,83 @@ export const stateV2: RequestHandler = async (req, res) => {
...
 
-		const local: LocalState = {
+		const local: LocalStateV3 = {
 			name: device.device_name,
 			config,
 			apps: appsForState,
 		};
 
-		const dependent = getDependent(device);
+		const dependent = await getDependent(device);
```
Page-, line 446:
```suggestion
		const dependent = getDependent(device);
```

src/features/device-state/routes/state.ts
```diff
@@ -362,24 +488,56 @@ const getConfig = (device: AnyObject) => {
...
+
+	return buildAppFromRelease(
+		device,
+		fromApi,
+		device.should_be_managed_by__release[0],
+		config,
+	);
+};
+
+const getDependent = async (device: AnyObject): Promise<Dependent> => {
```
Page-, line 540:
```suggestion
const getDependent = (device: AnyObject): Dependent => {
```

src/features/device-state/routes/state.ts
```diff
@@ -287,6 +334,30 @@ const stateQuery = _.once(() =>
 						},
 					},
 				},
+				should_be_managed_by__release: {
+					$select: releaseSelect,
+					$expand: {
+						has__tag_key: releaseExpand.has__tag_key,
+						contains__image: releaseExpand.contains__image,
+						belongs_to__application: {
```
Page-, line 342: This is just extending the default expand right? In that case it'd be nicer to do:
```suggestion
					$expand: {
						...releaseExpand,
						belongs_to__application: {
```

src/features/ci-cd/hooks/service-installs.ts
Page-: @toochevere I was talking about deleting the service installs in a single request, I'm not sure where creating service installs came from?

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -155,27 +197,87 @@ hooks.addPureHook('POST', 'resin', 'device', {
...
+				devices,
+				(d) => d.belongs_to__application.__id,
+			);
+
+			await Promise.all(
+				Object.keys(devicesByApp).map((appIdKey) => {
+					const appId = devicesByApp[appIdKey][0].id;
+
+					// delete service installs for devices currently with this app
+					actOnAppServiceInstalls(
```
Page-, line 271: This is missing an `await`
```suggestion
					await actOnAppServiceInstalls(
```

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -155,27 +197,87 @@ hooks.addPureHook('POST', 'resin', 'device', {
...
+						args.api,
+						appId,
+						devicesByApp[appIdKey].map((d) => d.id),
+						deleteHandler,
+					);
+
+					// create service installs for this app on all devices
+					actOnAppServiceInstalls(args.api, appId, affectedIds, createHandler);
```
Page-, line 279: This is missing an `await`
```suggestion
					await actOnAppServiceInstalls(args.api, appId, affectedIds, createHandler);
```

src/features/ci-cd/hooks/service-installs.ts
Page-: Ok, it makes sense that you now need to know the before state in order to only delete those

src/features/ci-cd/hooks/service-installs.ts
toochevere: https://github.com/balena-io/open-balena-api/blob/master/src/features/ci-cd/hooks/service-installs.ts#L167

The original code, on moving apps, deleted _all_ the previous installs and created the new ones for the target app.  We are now deleting _only_ the service installs for the previous app, and then as before creating the new ones for the target app.

Page-: I understand that behavior but the response to my initial message is what has confused me - I proposed deleting the existing service installs for the current app in a single api query instead of many, and then continue with the creation as normal

toochevere: oh, sorry.  I misunderstood. 🤦🏻   In any case, the creation is now also more optimized.  I'll make the deletion change

toochevere: done

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -54,6 +59,38 @@ const createReleaseServiceInstalls = async (
...
+	) => Promise<void>,
+): Promise<void> =>
+	actOnReleaseServiceInstalls(
+		api,
+		deviceIds,
+		{
+			should_be_running_on__application: {
+				$any: {
+					$alias: 'a',
+					$expr: { a: { id: { $in: appIds } } },
```
toochevere, line 82: @Page- just wanted to verify that this is the correct way to do this.

Page-: Yes this is valid/fine

src/features/ci-cd/hooks/service-installs.ts
toochevere: replaced the deletion code with the snippet you gave me.

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -155,27 +231,68 @@ hooks.addPureHook('POST', 'resin', 'device', {
...
+										device: deviceId,
+										installs__service: serviceId,
+									},
+									options: { returnResource: false },
+								});
+							}),
+						);
+					}),
+				);
+			};
```
Page-, line 263: This can now go back to how it was before right? Other than keeping the optimization of course

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -155,27 +231,68 @@ hooks.addPureHook('POST', 'resin', 'device', {
...
-						device: { $in: affectedIds },
+						id: { $in: affectedDeviceIds },
 					},
 				},
-			});
-			await createAppServiceInstalls(
+			})) as Array<{
+				id: number;
+				belongs_to__application: { __id: number };
+			}>;
```
Page-, line 277: It'd be better to fetch the app ids directly and remove the need to fetch the extra data and then process it locally:
```ts
			const apps = (await args.api.get({
				resource: 'application',
				options: {
					$select: ['id'],
					$filter: {
						owns__device: {
							$any: {
								$alias: 'd',
								$expr: { d: { id: { $in: affectedDeviceIds }, }
						}
					},
				},
			})) as Array<Pick<Application, 'id'>>;

src/features/ci-cd/hooks/service-installs.ts
toochevere: The optimization comes from using a separate handler in "moving to a different app" case.  If I put everything back, the optimization goes away.  

Specifically the optimization is that when we delete all service installs, we don't need to get through the process of getting existing ones and creating the difference.  The difference in behavior is because we are using a different handler (since the rest of the process is the same).   The special case handler creates _all_ the service installs for the app instead of doing the comparison.

PR #505, opened 2020-11-04
Add limited retries to device type fetching from S3

sradevski:
Currently if the API fails to fetch the device types,
it would retry infinitely, making the dashboard fail
to load in BoB if there are no device types in S3, and
this should resolve that.

Change-type: minor

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -65,3 +65,20 @@ export const throttledForEach = async <T, U>(
 	// We return the results of the iterator so the caller can await them as necessary
 	return await Promise.all(promises);
 };
+
+export const withRetries = async <T>(
+	func: () => Promise<T>,
+	delayDuration = 2000,
+	retries = 2,
+): Promise<T> => {
+	try {
+		return await func();
+	} catch (err) {
+		if (retries <= 0) {
+			throw err;
+		}
+
+		await delay(delayDuration);
+		return await withRetries(func, delayDuration, retries - 1);
+	}
+};
--- a/test/11-utils.ts
+++ b/test/11-utils.ts
@@ -1,0 +1,33 @@
+import * as _ from 'lodash';
+import { withRetries } from '../src/lib/utils';
+import { expect } from './test-lib/chai';
+
+describe('Utils', () => {
+	it('withRetries should retry before throwing', async () => {
+		let calls = 0;
+		const func = () => {
+			calls += 1;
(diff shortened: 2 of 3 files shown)


test/11-utils.ts
```diff
@@ -0,0 +1,33 @@
+import * as _ from 'lodash';
+import { withRetries } from '../src/lib/utils';
+import { expect } from './test-lib/chai';
+
+describe.only('Utils', () => {
```
thgreasi, line 5:
```suggestion
describe('Utils', () => {
```

sradevski: 🤦  we need a lint rule :)

src/lib/utils.ts
```diff
@@ -65,3 +65,21 @@ export const throttledForEach = async <T, U>(
...
 	return await Promise.all(promises);
 };
+
+export const withRetries = async <T>(
+	func: () => Promise<T>,
+	delayDuration = 2000,
+	retries = 3,
+	err?: Error,
+): Promise<T> => {
+	if (retries <= 0) {
```
thgreasi, line 75: How about moving this inside the `catch` so that we don't need to pass the errors as an argument?

thgreasi (commented):
Happy seeing this change :)
Which reminds me my old PR that I never actually merged...
See: https://github.com/balena-io/open-balena-api/pull/23

src/lib/utils.ts
sradevski: Good call, I should have seen that 😄

thgreasi (approved):

thgreasi:
:+1:

src/lib/utils.ts
```diff
@@ -65,3 +65,20 @@ export const throttledForEach = async <T, U>(
...
+
+export const withRetries = async <T>(
+	func: () => Promise<T>,
+	delayDuration = 2000,
+	retries = 3,
+): Promise<T> => {
+	try {
+		return await func();
+	} catch (err) {
+		if (retries <= 1) {
```
thgreasi, line 77: I thought that the `<= 0` was on purpose, like run this fn and retry N times.
But it seems that you meant the total attempts.
No strong opinion though. LGTM anyway.

sradevski: Hmm I thought of it as total retries (thinking of the first request as a try), but reading it now I think it makes sense what you're saying. I initially used `times` as a variable and didn't notice the subtle difference.

PR #506, opened 2020-11-09
Upgrade Loki test image to 2.0.0

codewithcheese:
Change-type: patch


codewithcheese:
@balena-ci I self-certify!

PR #507, opened 2020-11-10
Fix device types migration

dfunckt:
Fixes commit 81b1879 (#167).

Change-type: patch


sradevski (dismissed):

dfunckt:
@balena-ci I self-certify!

PR #508, opened 2020-11-10
Set Loki timeout to 1 second with no retries so log writes are not delayed

codewithcheese:
Change-type: patch


codewithcheese:
@balena-ci rebase

codewithcheese:
@balena-ci I self-certify!

PR #509, opened 2020-11-11
Add Loki host and port to sentry messages

codewithcheese:
Change-type: patch


codewithcheese:
@balena-ci I self-certify!

PR #510, opened 2020-11-13
Add `user has direct access to application` virtual resource

thgreasi:
Change-type: minor
See: https://github.com/balena-io/balena-api/issues/2808
See: https://github.com/balena-io/balena-api/pull/2849
See: https://www.flowdock.com/app/rulemotion/p-organizations/threads/9dIPQ9751qYPy0G6PEs26riZCOK
See: https://app.frontapp.com/open/cnv_8tqn74d

### DIFF
--- a/src/features/applications/models/user__has_direct_access_to__application.ts
+++ b/src/features/applications/models/user__has_direct_access_to__application.ts
@@ -1,0 +1,124 @@
+import type {
+	AbstractSqlModel,
+	SelectQueryNode,
+} from '@balena/abstract-sql-compiler';
+import * as _ from 'lodash';
+
+export const addToModel = (
+	abstractSql: AbstractSqlModel,
+	selectQueryNodeTransformer: (
+		selectQueryNode: SelectQueryNode,
+	) => SelectQueryNode = _.identity,
+) => {
+	abstractSql.tables['user-has direct access to-application'] = {
+		fields: [
+			{
+				dataType: 'Big Integer',
+				fieldName: 'id',
+				required: true,
+				index: 'PRIMARY KEY',
+			},
+			{
+				dataType: 'ForeignKey',
+				fieldName: 'user',
+				required: true,
+				references: {
+					resourceName: 'user',
+					fieldName: 'id',
+				},
+			},
+			{
+				dataType: 'ForeignKey',
+				fieldName: 'has direct access to-application',
+				required: true,
+				references: {
+					resourceName: 'application',
+					fieldName: 'id',
+				},
+			},
+		],
+		primitive: false,
+		name: 'user-has direct access to-application',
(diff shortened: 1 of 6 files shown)


src/features/applications/models/user__has_direct_access_to__application.ts
```diff
@@ -0,0 +1,126 @@
...
+} from '@balena/abstract-sql-compiler';
+import * as _ from 'lodash';
+
+export const addToModel = (
+	abstractSql: AbstractSqlModel,
+	selectQueryNodeTransformer: (
+		selectQueryNode: SelectQueryNode,
+	) => SelectQueryNode = _.identity,
+) => {
+	abstractSql.tables['user-has direct access to-application'] = {
```
thgreasi, line 14: I see though that someone could argue against the `direct` part of the name, since in oB users do not have a direct link to any app :thinking:

src/features/applications/models/user__has_direct_access_to__application.ts
```diff
@@ -0,0 +1,126 @@
...
+			application: {
+				$: ['has direct access to-application', ['application', 'id']],
+			},
+		},
+		'is directly accessible by': {
+			user: {
+				$: ['user', ['user', 'id']],
+			},
+		},
+	} as any) as Relationship;
```
Page-, line 65: Why is the `as any` necessary?

thgreasi: @Page-  It complains about the `Relationship`s defined not having a `$` property:
```
Type '{ application: { $: [string, [string, string]]; }; }' is not assignable to type 'Relationship | RelationshipMapping'.
  Property '$' is missing in type '{ application: { $: [string, [string, string]]; }; }' but required in type 'Relationship'.

(property) 'has direct access to': {
    application: {
        $: [string, [string, string]];
    };
}
```

The abstract model generated by equivalent sbvr though emitted exactly like this and I also see that we are doing the exact same `as any` case in the `application__can_use__application_as_host` computed term.

After further checking this, I could drop the `as any` with a slight modification of the `Relationship` typings, but will need you to confirm whether that's actually correct or not:
```typescript
export type Relationship = 
    | { $: RelationshipMapping; }
    | { [resourceName: string]: Relationship | RelationshipMapping; }
```
See: https://github.com/balena-io-modules/abstract-sql-compiler/blob/b504516a09ddcd5f4217969536ef823a00c004f9/src/AbstractSQLCompiler.ts#L297-L301

thgreasi: @Page- here is a PR that should fix this:
See: https://github.com/balena-io-modules/abstract-sql-compiler/pull/102

Page- (approved):

PR #511, opened 2020-11-16
Add DeviceLogsUsageMeter interface to be updated by DeviceLogsBackend

codewithcheese:
Customers subscribed to the logging add-on will be billed according to their usage. This PR allows for a `DeviceLogsUsageMeter` to be "attached" to the `DeviceLogsBackend`. This usage meter is updated by the backend when usage is incurred.

Part 1 of 2. Part 2 will be implemented in `balena-api`

Change-type: patch


codewithcheese:
Incrementing like this would be resource intensive. Closing in favor of pulling stats from Loki for a period.

PR #514, opened 2020-11-23
Remove `service_id` from LokiBackend labels to reduce cardinality

codewithcheese:
Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/backends/loki.ts
+++ b/src/features/device-logs/lib/backends/loki.ts
@@ -62,7 +62,7 @@ const RETRIES_ENABLED = false;
 const PUSH_TIMEOUT = 1000;
 const MIN_BACKOFF = 100;
 const MAX_BACKOFF = 10 * 1000;
-const VERSION = 1;
+const VERSION = 2;
 
 function createTimestampFromDate(date = new Date()) {
 	const timestamp = new Timestamp();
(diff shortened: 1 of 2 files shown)


Page- (commented):
Code looks fine, but it'd be nice to know the reasoning/where it was discussed to see if/why the change makes sense/is necessary

codewithcheese:
> Code looks fine, but it'd be nice to know the reasoning/where it was discussed to see if/why the change makes sense/is necessary

@Page- I'm glad you asked, its good to have someone to bounce off, this is a decision I made independently and have not discussed yet. Label cardinality is a bottleneck with Loki. Every unique combination of labels corresponds to an index and each index results in additional requests to object storage during queries. Basically the less label cardinality the better Loki performs. Per tenant the default cardinality limit is 100,000, this can be increased but is a very rough guideline of what the upper limit should be. 

Currently tenancy is based on app id and the largest app has 250k devices. We want a future proof design and just `device_id` alone pushes per tentant cardinality towards the limit. If the labels included both device_id and service_id its possible we could be already exceed 100,000 cardinality for our largest app. 

The device_id label is critical since the default query is to display the latest logs for all service on a device. However, we want to filter by service_id and perform more advanced queries to provide added value beyond our current implementation.  Fortunately Loki v2 can post-process queries by parsing the log line and using the parsed results as labels. Each log line is a JSON string containing the service_id, meta data and the device stdout log line. This JSON string can be post processed to filter by service id, it will be a bit slower but minimizes cardinality issues long term.

codewithcheese:
@balena-ci rebase

codewithcheese:
Looks good on BoB

codewithcheese:
@balena-ci I self-certify!

PR #515, opened 2020-11-23
Improve the cache of the S3-fetched device types

sradevski:
Change-type: patch

### DIFF
--- a/src/features/device-types/device-types.ts
+++ b/src/features/device-types/device-types.ts
@@ -101,50 +101,48 @@ async function fetchDeviceTypes(): Promise<Dictionary<DeviceTypeInfo>> {
 	const result: Dictionary<DeviceTypeInfo> = {};
 	getIsIgnored.clear();
 	getDeviceTypeJson.clear();
-	try {
-		const slugs = await listFolders(IMAGE_STORAGE_PREFIX);
-		await Promise.all(
-			slugs.map(async (slug) => {
-				try {
-					const builds = await listFolders(getImageKey(slug));
-					if (_.isEmpty(builds)) {
-						return;
-					}
-
-					const sortedBuilds = sortBuildIds(builds);
-					const latestBuildInfo = await getFirstValidBuild(slug, sortedBuilds);
-					if (!latestBuildInfo) {
-						return;
-					}
-
-					result[slug] = {
-						versions: builds,
-						latest: latestBuildInfo,
-					};
-
-					_.forEach(latestBuildInfo.deviceType.aliases, (alias) => {
-						result[alias] = result[slug];
-					});
-				} catch (err) {
-					captureException(
-						err,
-						`Failed to find a valid build for device type ${slug}`,
-					);
+	const slugs = await listFolders(IMAGE_STORAGE_PREFIX);
+	await Promise.all(
+		slugs.map(async (slug) => {
+			try {
+				const builds = await listFolders(getImageKey(slug));
+				if (_.isEmpty(builds)) {
+					return;
 				}
-			}),
-		);
 
-		if (_.isEmpty(result) && !_.isEmpty(slugs)) {
-			throw new InternalRequestError('Could not retrieve any device type');
-		}
-		return result;
-	} catch (err) {
-		captureException(err, 'Failed to get device types');
-		throw err;
+				const sortedBuilds = sortBuildIds(builds);
+				const latestBuildInfo = await getFirstValidBuild(slug, sortedBuilds);
+				if (!latestBuildInfo) {
+					return;
+				}
+
+				result[slug] = {
+					versions: builds,
+					latest: latestBuildInfo,
+				};
+
+				_.forEach(latestBuildInfo.deviceType.aliases, (alias) => {
+					result[alias] = result[slug];
+				});
+			} catch (err) {
+				captureException(
+					err,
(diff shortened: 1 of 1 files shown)


src/features/device-types/device-types.ts
```diff
@@ -172,7 +171,12 @@ async function fetchDeviceTypesAndReschedule(): Promise<
 		// in case another api request comes before the first completes
 		if (!deviceTypesCache) {
 			deviceTypesCache = promise;
+			// We want to clear out the cache if the cached promise fails
+			promise.catch(() => {
```
thgreasi, line 175: How about moving this to a proper `catch (err) {}` below so that we don't mix await & promises and in there do something like:
```
if (deviceTypesCache === promise) {
    deviceTypesCache = undefined;
}
```
and then rethrow?
Regardless of that, let's also add a `captureException` in here as well.

src/features/device-types/device-types.ts
```diff
@@ -159,21 +154,27 @@ async function fetchDeviceTypesAndReschedule(): Promise<
...
 	try {
-		const promise = withRetries(fetchDeviceTypes).then(
-			async (deviceTypeInfo) => {
-				// when the promise gets resolved, cache it
-				deviceTypesCache = promise;
-				return deviceTypeInfo;
-			},
-		);
+		const promise = withRetries(fetchDeviceTypes).then((deviceTypeInfo) => {
+			deviceTypesCache = Promise.resolve(deviceTypeInfo);
```
Page-, line 158: Strictly speaking there's no need to wrap this since it's only ever used in an `await` context anyway

src/features/device-types/device-types.ts
```diff
@@ -159,21 +154,27 @@ async function fetchDeviceTypesAndReschedule(): Promise<
...
 		// if the cache is still empty, use this promise so that
 		// we do not start a second set of requests to s3
 		// in case another api request comes before the first completes
 		if (!deviceTypesCache) {
 			deviceTypesCache = promise;
 		}
-		return await promise;
+
+		try {
+			return await promise;
```
Page-, line 170: Based on `deviceTypesCache` not needing to be an actual promise you could remove the `.then` above and replace it with
```suggestion
			deviceTypesCache = await promise;
			return deviceTypesCache;
```
or
```suggestion
			return deviceTypesCache = await promise;
```
here

thgreasi: @Page-  The reason we are assigning the promise to the `deviceTypesCache`, is so that we do not start more than one concurrent `fetchDeviceTypes()` calls in case that multiple requests requiring this come in before we populate the cache.

Page-: @thgreasi that's the reason for
```
		if (!deviceTypesCache) {
			deviceTypesCache = promise;
		}
```
above, to note I am *not* suggesting to remove that but instead to simplify how we replace the cached version on subsequent calls.

thgreasi: Oh and you mean that we would also change:
```
let deviceTypesCache: Promise<Dictionary<DeviceTypeInfo>> | Dictionary<DeviceTypeInfo> | undefined;

// and then later just ...

		const promise = withRetries(fetchDeviceTypes);
		if (!deviceTypesCache) {
			deviceTypesCache = promise;
		}
		return (deviceTypesCache = await promise);
```
?
In that case I love your suggestion :heart: :heart: :heart:

sradevski: @Page- not sure that would work, if `deviceTypesCache` is not a promise, its value will only be set after `promise` resolves, which is too late to handle requests that come in the meantime.

thgreasi: Let me add more :heart:s for this.
:heart: :heart: :heart: 
:heart: :heart: :heart:

sradevski: ohh I just read Thodoris' comment, if that is what you mean then yes, that would work :)

Page-: Yeah, to expand further for clarity I'm thinking
```ts
async function fetchDeviceTypesAndReschedule(): Promise<
	Dictionary<DeviceTypeInfo>
> {
	try {
		const promise = withRetries(fetchDeviceTypes);

		// if the cache is still empty, use this promise so that
		// we do not start a second set of requests to s3
		if (!deviceTypesCache) {
			deviceTypesCache = promise;
		}

		try {
			return deviceTypesCache = await promise;
		} catch (err) {
			if (deviceTypesCache === promise) {
				deviceTypesCache = undefined;
			}
			captureException(err, 'Failed to get device types');
			throw err;
		}
	} finally {
		// schedule a re-run to update the local cache - do not wait for it
		scheduleFetchDeviceTypes();
	}
}
```
and if you want the
```ts
		if (!deviceTypesCache) {
			deviceTypesCache = promise;
		}
```
could be replaced by
```ts
		deviceTypesCache ??= promise;

sradevski: Yeah ok that's how I got it after Thodoris' comment, I thought you are suggesting to not keep the deviceTypesCache as a promise.

Page-: Yeah, more accurately I was suggesting it doesn't *always* need to be a promise

Page- (approved):

PR #516, opened 2020-11-23
Add contract field to release resource

sradevski:
Change-type: minor

### DIFF
--- a/src/migrations/00037-add-contract-release.sql
+++ b/src/migrations/00037-add-contract-release.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "release" 
+ADD COLUMN IF NOT EXISTS "contract" TEXT NULL;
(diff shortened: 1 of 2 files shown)


src/migrations/00037-add-contract-release.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "release" 
+ADD COLUMN IF NOT EXISTS "contract" TEXT NULL;
```
sradevski, line 2: I saw we are using `TEXT` in the migrations instead of JSON, despite the sbvr type being JSON. Any reason for that?

Page-: The SBVR JSON type predates the binary JSON types (and possibly the JSON type altogether?) so it uses TEXT - it'd be nice to move over to more closely match what would be expected but in practical terms we don't support any of the special things the JSON/JSONB types enable so it doesn't really affect anything

Page- (approved):

PR #520, opened 2020-11-27
Add option to Loki backend to print verbose error message

codewithcheese:
Option to print verbose error message to further diagnose out of order message on staging. Will set to false once complete.

Change-type: patch


codewithcheese:
@balena-ci I self-certify!

PR #522, opened 2020-11-30
Fix device log read serialization to ignore `nanoTimestamp`

codewithcheese:
Fix for "TypeError: Do not know how to serialize a BigInt". Fails to stringify BigInt `nanoTimestamp` when reading via `onLog`, added replacer to exclude `nanoTimestamp`. Will consider how to remove the need for this is in subsequent PRs.

Change-type: patch


codewithcheese:
@balena-ci I self-certify!

PR #523, opened 2020-12-02
Add a base64-encoded logo to device types

sradevski:
Change-type: minor


src/balena.sbvr
```diff
@@ -156,6 +156,9 @@ Term: last vpn event
 Term: local id
 	Concept Type: Short Text (Type)
 
+Term: logo
```
dfunckt, line 159:
```suggestion
Term: logotype
```

I prefer `logotype` but don't feel strongly about it as `logo` is also widely used.

Regardless, this seems to be just blob that we happen to store image data into, so either is not fitting. Maybe the term should be `logotype data`? I don't know, just thinking out loud. @Page- do you have an opinion?

src/features/contracts/contracts-directory.ts
```diff
@@ -8,13 +8,82 @@ import * as stream from 'stream';
...
 const exists = util.promisify(fs.exists);
 const mkdir = util.promisify(fs.mkdir);
 
 const CONTRACTS_BASE_DIR = path.join(os.tmpdir(), 'contracts');
 
+const getDataUri = async (
+	filePath: string,
+	mimeType: 'image/png' | 'image/svg+xml',
+) => {
+	const base64Content = (await fs.promises.readFile(filePath)).toString(
```
dfunckt, line 23: Are logotypes stored on disk?

sradevski: @dfunckt They are included as part of the contracts repo, and when we sync the contracts we basically pull the entire repo tarball, so yes, they will be on disk together with the contract json

src/balena.sbvr
sradevski: @dfunckt Since it's in `datauri` format, the idea is to have it interchangeable with a remote URL of the logo. What we decided is to store the logos as base64 encoded data in a datauri format for now as it is simpler to apply permissions against it (we don't want to expose private DT logos to everyone), and if this proves to not scale well, we can replace it with an S3 URL, for example.

src/features/contracts/contracts-directory.ts
dfunckt: Hmm, I didn't realise that we were doing that for contracts -- we should either move away from this sort of writing into the container or discuss exposing a mount point where storage can be bound.

src/balena.sbvr
dfunckt: Maybe name it `logotype url` instead?

src/features/contracts/contracts-directory.ts
sradevski: @dfunckt why do you think we shouldn't be writing in the container? We only write to `tmp` storage and we don't really care if we lose that data as we will refetch it on the next sync.

src/balena.sbvr
sradevski: AFAIK `logotype` refers only to the textual part of a logo, in this case we have both so it really is a logo. `logoUri` sounds fine though, I'll just wait for Page to chime in before making changes

src/features/contracts/contracts-directory.ts
dfunckt: Containers should generally be "immutable" @sradevski -- by silently writing into the container you're making the assumption that there's arbitrarily large underlying storage available, which is not always the case and is what bind mounts and volumes are for (which make the storage requirement explicit and lets one plan accordingly).

sradevski: I don't think that's necessarily true, since the `contracts` data is essentially a cache, I don't see a difference between storing that in memory or on disk inside the container. In this case, I explicitly don't want the contract data to be persisted outside of a container's lifecycle, which is why I didn't want to use volumes. It doesn't really matter much if we use volumes TBH, I just don't see how that would be a better option here.

src/balena.sbvr
Page-: I'd never heard of `logotype` before @dfunckt but TIL `logo` is actually short for `logotype` (and also possibly for `logomark`), so my first reaction was "did you mean `logo type`?" as in the type of the logo. So fwiw I'd go with `logo`/`logo url` based on that

src/features/contracts/contracts-directory.ts
```diff
@@ -8,13 +8,82 @@ import * as stream from 'stream';
...
+			try {
+				normalizedAssets[key] = {
+					...asset,
+					url: await handleLocalAssetUrl(normalizedUrl),
+				};
+			} catch (e) {
+				console.error(
+					`Failed to normalize contract asset for url ${normalizedUrl}, excluding asset`,
+					e,
+				);
```
Page-, line 79: If this warrants a `console.error` then it should probably be a `captureException` instead so it can be forwarded to sentry as well

src/features/contracts/contracts-directory.ts
```diff
@@ -8,13 +8,82 @@ import * as stream from 'stream';
...
+				return;
+			}
+
+			// Convert from relative to absolute path for the asset file and make sure it doesn't try to access files outside of the contract folder.
+			const contractDir = path.dirname(contractFilepath);
+			const normalizedUrl = path.resolve(contractDir, asset.url);
+			if (!normalizedUrl.startsWith(contractDir)) {
+				console.error(
+					`Contract asset URL '${asset.url}' is invalid, excluding asset from contract`,
+				);
```
Page-, line 66: If this warrants a `console.error` then it should probably be a `captureException` instead so it can be forwarded to sentry as well

src/lib/utils.ts
```diff
@@ -14,6 +14,14 @@ export function getNanoTimestamp() {
 
 export const pseudoRandomBytesAsync = promisify(pseudoRandomBytes);
 
+export const isValidUrl = (url: string) => {
+	try {
+		return Boolean(new URL(url));
```
Page-, line 19: This is somewhat confusing as it implies we care about the return value of `new URL` but I think the actual expected behavior is
```suggestion
		new URL(url);
		return true;
```
as we're actually relying on `new URL` throwing for invalid URLs rather than anything to do with its return values

Page-, line 19: Relying on `new URL` doesn't necessary mean the input is a valid URL per se since it will do some coercion as according to https://nodejs.org/docs/latest-v14.x/api/url.html#url_new_url_input_base where it will attempt to coerce into a valid string, eg `{ toString: () => 'https://example.org/' }` would be valid despite not being a URL string and it will convert unicode characters into ascii using punycode, but those characters from the original string aren't strictly valid (they rely on whatever is using them being aware of how to convert them to ascii before querying dns). There's also the question of if we want to restrict to certain protocols eg https

src/features/contracts/contracts-directory.ts
```diff
@@ -8,13 +8,82 @@ import * as stream from 'stream';
...
+	await Promise.all(
+		Object.entries(assets).map(async ([key, asset]) => {
+			if (isValidUrl(asset.url)) {
+				normalizedAssets[key] = asset;
+				return;
+			}
+
+			// Convert from relative to absolute path for the asset file and make sure it doesn't try to access files outside of the contract folder.
+			const contractDir = path.dirname(contractFilepath);
+			const normalizedUrl = path.resolve(contractDir, asset.url);
```
Page-, line 62: You need to check that `asset.url` does not include anything like a `..` which might escape the contracts directory and also that no symlinks to outside of the directory can be followed

src/balena.sbvr
thgreasi: I'm also confused by `logotype` but `logouri`/`logourl` sounds fine, just in case at some point we want to use `logo` for something else.

src/features/contracts/contracts-directory.ts
thgreasi: I guess we could use `tmpfs` (?) which seems to be the docker way for all the use cases that `/tmp` offers.
See: https://docs.docker.com/storage/tmpfs/

src/features/contracts/contracts-directory.ts
```diff
@@ -8,13 +8,82 @@ import * as stream from 'stream';
...
 import type { RepositoryInfo, Contract } from './index';
+import { isValidUrl } from '../../lib/utils';
 
 const pipeline = util.promisify(stream.pipeline);
 const exists = util.promisify(fs.exists);
 const mkdir = util.promisify(fs.mkdir);
 
 const CONTRACTS_BASE_DIR = path.join(os.tmpdir(), 'contracts');
 
+const getDataUri = async (
```
thgreasi, line 19: NIT: Maybe also move this to utils?

src/features/contracts/contracts-directory.ts
```diff
@@ -8,13 +8,82 @@ import * as stream from 'stream';
...
+const handleLocalAssetUrl = async (assetUrl: string): Promise<string> => {
+	switch (path.extname(assetUrl)) {
+		case '.svg': {
+			return getDataUri(assetUrl, 'image/svg+xml');
+		}
+		case '.png': {
+			return getDataUri(assetUrl, 'image/png');
+		}
+		default:
+			return assetUrl;
```
thgreasi, line 39: Would it make more sense to just return `null` instead of the fs path?

src/features/contracts/contracts-directory.ts
```diff
@@ -8,13 +8,82 @@ import * as stream from 'stream';
...
+				return;
+			}
+
+			try {
+				normalizedAssets[key] = {
+					...asset,
+					url: await handleLocalAssetUrl(normalizedUrl),
+				};
+			} catch (e) {
+				console.error(
```
thgreasi, line 76: Should we `errors.captureException()` instead?

test/09-contracts.ts
```diff
@@ -65,6 +65,34 @@ describe('contracts', () => {
...
+
+			const contracts = await getContracts('hw.device-type');
+			const rpi3Contract = contracts.find(
+				(contract) => contract.slug === 'raspberrypi3',
+			);
+			const otherDtContract = contracts.find(
+				(contract) => contract.slug === 'other-contract-dt',
+			);
+
+			expect(rpi3Contract?.assets?.logo).to.exist;
```
thgreasi, line 84: NIT: Do we really need all those `?.` (and also repeat in every line)?
I mean we just mocked that.
If we still want to, then it might make sense to `expect().to.have.nested,property('assets.logo')`.

src/features/contracts/contracts-directory.ts
sradevski: @thgreasi I was also thinking about that, but this is only read once every 5-10 minutes and I don't think it's worth keeping it in memory (since tmpfs doesn't write anything on disk). Anyhow, we should settle on one approach and we can go with that, but I personally find just storing it on disk to be the best option for the contracts.

src/balena.sbvr
sradevski: I'll go with logo URI since URL is a bit inaccurate in this case

src/lib/utils.ts
sradevski: @Page- since we control the contracts repo I don't think we need a very strict check, but I agree that this function doesn't take care of any edge-cases. What do you think, do we keep it like this or shall I go with a more elaborate check using some package or regex?

src/features/contracts/contracts-directory.ts
sradevski: @Page- doesn't the check below this line suffice? This will resolve it into an absolute URL, and then we check if that absolute URL is within the contracts repo.

src/features/contracts/contracts-directory.ts
sradevski: yeah maybe, lets do that

test/09-contracts.ts
sradevski: Typescript wants them, and I honestly prefer those over checking if it has a nested property.

src/features/contracts/contracts-directory.ts
sradevski: The Contract spec requires that there is a URL, if you provided an asset, so we need to return a string. It can be an empty one though, let's do that.

src/lib/utils.ts
sradevski: @Page- I remember now why I did it like this, the linter complains about an unused rule. Shall we leave it as it is, or ignore the rule?

Page-: I thought the idea was that we could theoretically have other contracts repos supplied? In which case I'd like to have decent checks, especially since I can easily see this function being reused by other non-contracts code.
There is `https://www.npmjs.com/package/validator` which we use elsewhere to validate emails that has an `isURL` and might be an improvement, I'm not sure how it compares though

src/features/contracts/contracts-directory.ts
Page-: Ah, yeah it should cover the `..` case, not sure about symlinks though

sradevski: I don't think there is a way for this to be a symlink since it's just a string path, not an actual file that symlinks to something else.

src/lib/utils.ts
sradevski: We won't introduce new contract repos, but people could PR to our contracts repo (which we need to review), but I agree, let's go with a more robust check.

Page-: I think I'd ignore the rule, with a comment as to how it's used (were relying on the throw behavior)

Page-: We don't intend to introduce new contract repos today, but I could definitely imagine it in future, whether for balenahub or community supported device types or whatever so I'd prefer to play it safe. It also protects against the code being reused or copy pasted somewhere where it really does matter - I've definitely seen a pattern that is ok in one place reused in a place it's not ok before

src/features/contracts/contracts-directory.ts
Page-: You can have directory symlinks @sradevski and the path is read from the filesystem so there is definitely the possibility to resolve those symlinks, the final .svg or .png could also be a symlink itself couldn't it? Overall I'd prefer to be extra careful with anything that can read from the filesystem because it'd be really bad if someone managed to grab config variables

sradevski: Fair point, even though we control the contract repos it's better to err on the safe side, I agree. Added a check for symlinks as well

src/features/contracts/contracts-directory.ts
```diff
@@ -7,14 +7,79 @@ import * as util from 'util';
...
+				normalizedAssets[key] = asset;
+				return;
+			}
+
+			// Convert from relative to absolute path for the asset file and make sure it doesn't try to access files outside of the contract folder.
+			const contractDir = path.dirname(contractFilepath);
+			const normalizedUrl = path.resolve(contractDir, asset.url);
+			if (
+				!normalizedUrl.startsWith(contractDir) ||
+				(await (await fs.promises.lstat(normalizedUrl)).isSymbolicLink())
```
Page-, line 57: Does this check if the path included a symlink or only if the final entry is a symlink? Because it's the former we'd need to do this properly (the latter would just give a false sense of safety since we could already have escaped the contract dir). What might be a better approach is to resolve the absolute path and then compare that to the contracts dir, that way you can have symlinks that are internal to the contract dir but not ones that leave it (well, technically they can leave it if they come back in.. but I guess that's fine if you wanna do that? :S  )

sradevski: @Page- updated, let me know what you think. I'll give it a go in bob now as well, but the tests are happy

Page- (approved):

PR #526, opened 2020-12-11
Avoid updating unchanged entries when syncing a contract

Page-:
Change-type: patch

### DIFF
--- a/src/features/contracts/index.ts
+++ b/src/features/contracts/index.ts
@@ -125,8 +125,7 @@ const upsertEntries = async (
 						resource,
 						body: entry,
 						options: {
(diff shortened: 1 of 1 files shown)


thgreasi (approved):

src/features/contracts/index.ts
```diff
@@ -125,8 +125,7 @@ const upsertEntries = async (
 						resource,
 						body: entry,
 						options: {
-							$filter: { [uniqueField]: entry[uniqueField] },
-							returnResource: false,
+							$filter: { [uniqueField]: entry[uniqueField], $not: entry },
```
sradevski, line 128: TIL about the $not

PR #529, opened 2020-12-15
Export the DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES

thgreasi:
This is needed so that the cloud API can add private
device types in the device type specific config vars
definitions.

Change-type: minor


thgreasi:
@balena-ci I self-certify!

PR #531, opened 2020-12-16
Perform username/email checks in a hook rather than custom function

Page-:
Change-type: minor


Page-:
@balena-ci rebase

Page-:
@balena-ci rebase

thgreasi (approved):
LGTM :+1: 
The only case I would be curious to see how we could implement, is if we ever wanted to be able to create that root user (or anything else that's in the username whitelist), but I guess we could do that with the init scripts.

PR #532, opened 2020-12-16
Update dependencies

Page-:
Update pinejs from 14.2.6 to 14.6.0

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #534, opened 2020-12-18
Add the ability to generate a materialized pinejs config.json

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #539, opened 2020-12-28
Catch exception when resolving asset real path to prevent other contract syncing from failing

sradevski:
This will prevent syncing only the contracts that have an issue when normalizing their path, otherwise, all contract syncing for the device types would have failed.

Change-type: patch


thgreasi (approved):

sradevski:
@balena-ci rebase

PR #542, opened 2021-01-04
Update pinejs to 14.9.2

Page-:
Update pinejs from 14.9.1 to 14.9.2

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #543, opened 2021-01-04
Regenerate balena-model types to include type helpers

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #545, opened 2021-01-08
Split npm script for model typings generation to two separate ones

thgreasi:
Change-type: patch


thgreasi:
@balena-ci I self-certify!

PR #547, opened 2021-01-13
Remove unnecessary DELETE hook that tries to validate config/env vars

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #550, opened 2021-01-14
model: add some release metadata

xginn8:
Connects-to: https://github.com/balena-io/balena-io/issues/2335
Change-type: patch

### DIFF
--- a/src/balena-model.ts
+++ b/src/balena-model.ts
@@ -364,6 +364,8 @@ export interface Release {
 	start_timestamp: DateString;
 	end_timestamp: DateString | null;
 	update_timestamp: DateString;
+	is_passing_tests: boolean;
+	release_type: 'final' | 'draft';
 	release_version: string | null;
 	contract: {} | null;
 }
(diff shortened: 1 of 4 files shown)


src/balena-model.ts
```diff
@@ -364,6 +364,10 @@ export interface Release {
 	start_timestamp: DateString;
 	end_timestamp: DateString | null;
 	update_timestamp: DateString;
+	is_passing_tests: boolean;
+	release_type: 'final' | 'draft';
+	is_child_of__release: { __id: number } | [Release];
```
sradevski, line 369: Let's not add this and `pull_request_url ` just yet, the whole thing about child releases and pull requests is a bit unclear as of now. We are still not decided on how all of it will work as part of the transformers work, so I'd rather wait for us to finalize how everything will work there before we add it to our balena.io model

xginn8: sounds good, that's step 2 anyway. updated.

sradevski (approved):

PR #553, opened 2021-01-18
Separate application delete hook components into individual hooks

Page-:
This allows them to be isolated from each other and also allows them to
be executed in parallel

Change-type: patch

### DIFF
--- a/src/features/applications/hooks/backwards-compatible-delete-error.ts
+++ b/src/features/applications/hooks/backwards-compatible-delete-error.ts
@@ -1,0 +1,15 @@
+import { sbvrUtils, hooks, errors } from '@balena/pinejs';
+
+hooks.addPureHook('DELETE', 'resin', 'application', {
+	PRERUN: async (args) => {
+		const appIds = await sbvrUtils.getAffectedIds(args);
+		if (appIds.length === 0) {
+			const { odataQuery } = args.request;
+			if (odataQuery != null && odataQuery.key != null) {
+				// If there's a specific app targeted we make sure we give a 404 for backwards compatibility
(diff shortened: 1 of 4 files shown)


src/features/applications/hooks/backwards-compatible-delete-error.ts
```diff
@@ -0,0 +1,15 @@
+import { sbvrUtils, hooks, errors, permissions } from '@balena/pinejs';
+
+hooks.addPureHook('DELETE', 'resin', 'application', {
+	PRERUN: async (args) => {
+		const appIds = await sbvrUtils.getAffectedIds(args);
+		if (appIds.length === 0) {
+			const { odataQuery } = args.request;
+			if (odataQuery != null && odataQuery.key != null) {
+				// If there's a specific app targeted we make sure we give a 404 for backwards compatibility
+				throw new errors.NotFoundError('Application(s) not found.');
```
Page-, line 10: @thgreasi @dfunckt do you remember why this is a thing? Maybe we can retire it now?

thgreasi: The SDK does check for such a result:
See: https://github.com/balena-io/balena-sdk/blob/4d799ad1705ed560f0c7fd0a7dc5781e342e4e8a/lib/models/application.ts#L715-L728
See: https://github.com/balena-io/balena-sdk/blob/4d799ad1705ed560f0c7fd0a7dc5781e342e4e8a/tests/integration/models/application.spec.js#L345-L358

I'm not sure about the OData semantics, but in general that's how I've seen most REST services handling such requests.
Would it make sense to have something like this by default for all resources in pine?

thgreasi (approved):

PR #554, opened 2021-01-19
Update pinejs to 14.10.2 for performance optimization

Page-:
Update pinejs from 14.10.1 to 14.10.2

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #555, opened 2021-01-19
Revert metadata

xginn8:


thgreasi:
FYI LGTM (aka feel free to self-certify) since I don't think we need to go as far as writing the reverse migration.

xginn8:
@balena-ci I self-certify!

PR #556, opened 2021-01-20
releases: add new metadata to release model

xginn8:
filter non-final/untested releases from the state endpoint

Connects-to: https://github.com/balena-io/balena-io/issues/2335
Change-type: minor


src/balena-model.ts
```diff
@@ -366,6 +366,8 @@ export interface Release {
 	update_timestamp: DateString;
 	release_version: string | null;
 	contract: {} | null;
+	is_passing_tests: boolean;
+	release_type: 'final' | 'draft';
```
xginn8, line 370: this is actually autogenerated this time (using the `Definition` rather than a rule)

test/03-device-state-v2.ts
```diff
@@ -333,6 +333,64 @@ describe('Device State v2', () => {
 			);
 		});
 	});
+
+	describe(`Tracking latest release`, () => {
```
xginn8, line 337: @Page- i'm not sure if these tests belong in this state endpoint file, or elsewhere as a "scenario".

xginn8:
@balena-ci retest

src/features/ci-cd/hooks/track-latest-release.ts
```diff
@@ -2,19 +2,18 @@ import * as _ from 'lodash';
 
 import { sbvrUtils, hooks } from '@balena/pinejs';
 
-const updateLatestRelease = async (
-	id: number,
-	{ request, api }: hooks.HookArgs,
-) => {
-	// We only track builds that are successful
-	if (request.values.status !== 'success') {
```
xginn8, line 10: @Page- this is the only part of this PR that I'm troubled by -- this approach using params from the `request` date back to the split of open and closed APIs. I don't see any reason this approach should be different (other than not short circuiting as frequently), but i wanted to make sure I understood the approach before flipping this from draft.

test/03-device-state-v2.ts
Page-: I'm happy for these to be in tests but I'd put them as separate device pinning tests since that's what they're mainly testing rather than the device state per se (they're checking the device state response as a way to verify the pinning but that could just as easily be done via an OData GET)

src/balena-model.ts
xginn8: 100%, see https://github.com/balena-io/open-balena-api/pull/556/files#diff-6fbfd1ba14aab28155305cee9b841695255dd947b2f661fb6e375a4de53403e4R253

xginn8:
@balena-ci rebase

xginn8:
@balena-ci rebase

src/features/ci-cd/hooks/track-latest-release.ts
```diff
@@ -1,20 +1,31 @@
...
 		return;
 	}
-	const release = await api.get({
+	const [release] = await api.get({
 		resource: 'release',
-		id,
 		options: {
 			$select: 'id',
+			$top: 1,
+			$orderby: { start_timestamp: 'desc' },
```
xginn8, line 21: originally i had this using `end_timestamp`, but i think `start_timestamp` is more consistent with what folks expect.

xginn8:
blocking myself while i figure out if i want to update the rule that guarantees `release_version` uniqueness to incorporate these new metadata or not.

xginn8:
@balena-ci rebase

xginn8:
@balena-ci rebase

xginn8:
@balena-ci retest

xginn8:
@sradevski we currently have a rule enforcing uniqueness for `release version` within an app for successful releases: 

> Rule: It is necessary that each release that has a release version1 and has a status that is equal to "success" and is not invalidated, belongs to an application that owns exactly one release that has a release version2 that is equal to the release version1 and has a status that is equal to "success" and is not invalidated.

i _think_ it makes sense to extend this to `..and is passing tests and has a release type that is equal to "final"`, though paradoxically that is _less_ constrained (since we're excepting certain conditions from the uniqueness enforcement). i think that means it'd be best to not add those exceptions as of yet, and revisit if required.

src/features/ci-cd/hooks/track-latest-release.ts
```diff
@@ -1,20 +1,31 @@
...
-	id: number,
-	{ request, api }: hooks.HookArgs,
-) => {
-	// We only track builds that are successful
-	if (request.values.status !== 'success') {
+const updateLatestRelease = async ({ request, api }: hooks.HookArgs) => {
+	if (
+		!(
+			'status' in request.values ||
+			'is_passing_tests' in request.values ||
```
sradevski, line 9: You can change this to something like 
```
//This will go outside the function
const releaseStateFields = ['status', 'is_passing_tests', 'release_type', 'is_invalidated']

if(!releaseStateFields.some(field => field in request.values)) {
  return
}
```

src/features/ci-cd/hooks/track-latest-release.ts
sradevski: I think it makes sense 👍

sradevski (commented):
@xginn8 about adding the tests and type rules, I think we can leave them out for now and make sure that draft releases have a different, unique version, similar to how we do it today (when you open a PR, we add a hash to potential final release version). If we find this to be an issue we can always go back and add the rule, since as you said, it actually makes it less constrained.

src/features/ci-cd/hooks/track-latest-release.ts
xginn8: thanks, that's indeed a bit cleaner!

src/features/ci-cd/hooks/track-latest-release.ts
Page-: It might be worth adding a comment about why we order by start_timestamp

src/features/ci-cd/hooks/track-latest-release.ts
```diff
@@ -107,24 +116,23 @@ const updateLatestRelease = async (
...
 		}
 	},
 });
 
 hooks.addPureHook('POST', 'resin', 'release', {
 	POSTRUN: async (args) => {
-		// If we're creating a build then check if the latest release needs to be updated
+		// If we're creating a release then check if the latest release needs to be updated
 		const id = args.result;
 		if (id != null) {
```
Page-, line 134: There's no need to store to a var now and slightly changing the comment to explain why we have this check makes sense imo - ie don't bother checking for release updates if we didn't actually create anything
```suggestion
		// If we succesfully created a release then check if the latest release needs to be updated
		if (args.result != null) {
```

src/features/ci-cd/hooks/track-latest-release.ts
```diff
@@ -107,24 +116,23 @@ const updateLatestRelease = async (
 hooks.addPureHook('PATCH', 'resin', 'release', {
 	POSTRUN: async (args) => {
 		const { request } = args;
-		// If we're updating a build by id and setting it successful then we update the application to this build
+		// If we're updating a release by id and setting it successful then we update the application to this build
 		if (request.odataQuery != null) {
 			const keyBind = request.odataQuery.key;
 			// TODO: Support named keys
 			if (keyBind != null && 'bind' in keyBind) {
```
Page-, line 123: I think you can ditch both of these ifs since they were only as a way to get the id and we don't care about that anymore, instead you can add a `request.affectedIds.length > 0` check just to make sure we did at least update something

src/features/ci-cd/hooks/track-latest-release.ts
```diff
@@ -1,20 +1,29 @@
...
-	if (request.values.status !== 'success') {
+import { hooks } from '@balena/pinejs';
+const releaseStateFields = [
+	'status',
+	'is_passing_tests',
+	'release_type',
+	'is_invalidated',
+];
+const updateLatestRelease = async ({ request, api }: hooks.HookArgs) => {
+	if (!releaseStateFields.some((field) => field in request.values)) {
```
Page-, line 11: These fields have defaults on creation right, which means they might be specified but still potentially end up with the correct values to become the new default - is that handled correctly?

src/features/ci-cd/hooks/track-latest-release.ts
Page-: Yeah the new approach just seems like it has less short-circuiting but that makes sense since it's now possible for "track latest" to trigger a downgrade if the previously newest version becomes invalid

src/features/ci-cd/hooks/track-latest-release.ts
xginn8: @page- i may just be fried but i don't quite follow what you mean here, can you rephrase please?

Page-: So this is called in the `POST` hook, at which point the request may omit fields and therefore `field in request. values === false` however those fields have defaults in the database which means that they have an implicit value, eg `is_passing_tests` will have the implicit value `true`. I think it'd be fine in practice as I think it'd be required to specify at least `status` on create in order to alter the latest release however it'd be better to be safe to it by design (or at least by comment)

xginn8: this function is called both in the `POST` and `PATCH` hooks, which is how releases are created aiui (first a stub is POSTed, then PATCHed when more information is available). The default values provided by the DB are set in such a way that with no extra metadata they will be set as the latest release (failing open rather than closed, in a way). since this is already the expected behavior when building, i believe this is a cleaner change than making these metadata opt-out. am i still missing something here? if all the fields were to be omitted, there couldn't be a `status==success` since that's not provided as a default, and therefore the latest release couldn't be updated.

Page-: Yes that's what I'm saying, it happens to be the case that this is safe because the defaults do not match a valid latest results, but if those defaults were changed such that they did create a valid latest release we would short-circuit when we should not. Hence ideally I want the design to enforce that we're safe even if those defaults change, or at a minimum document that the defaults must not create a a valid latest release

xginn8: sorted out in https://www.flowdock.com/app/rulemotion/resin-tech/threads/Vgitvd6OIOi5BoAJUq5ClqqkvLk

src/features/ci-cd/hooks/track-latest-release.ts
```diff
@@ -107,24 +121,22 @@ const updateLatestRelease = async (
...
-		if (request.odataQuery != null) {
-			const keyBind = request.odataQuery.key;
-			// TODO: Support named keys
-			if (keyBind != null && 'bind' in keyBind) {
-				const id = sbvrUtils.resolveOdataBind(request.odataBinds, keyBind);
-				await updateLatestRelease(id, args);
-			}
+		// If we're updating a release by id and setting it successful then we update the application to this build
+		if (
+			request.odataQuery != null &&
```
Page-, line 126: There's no need to check `request.odataQuery` since we no longer use it to short circuit
```suggestion
```

src/features/ci-cd/hooks/track-latest-release.ts
```diff
@@ -107,24 +121,22 @@ const updateLatestRelease = async (
...
+			request.affectedIds &&
+			request.affectedIds?.length > 0
```
Page-, line 128: The `?` here should be unnecessary, and if not then it's dangerous due to doing `undefined > number`
```suggestion
			request.affectedIds.length > 0
```

Page- (approved):

PR #557, opened 2021-01-20
Update pinejs to 14.12.0

Page-:
Update pinejs from 14.10.2 to 14.12.0

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #559, opened 2021-01-22
Switch to using an object to specify dependent resources to delete

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #562, opened 2021-01-25
Fix potential TypeError from attempting to read `.length` of null

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #565, opened 2021-01-28
Add tests for the unique successful release commit rule

thgreasi:
Change-type: patch


thgreasi:
@balena-ci I self-certify!

PR #570, opened 2021-02-01
Support setting pg statement/query timeouts, defaulting to 60s/61s

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #572, opened 2021-02-02
Add permission checking to the supervisor proxy

thgreasi:
Change-type: patch
See: https://github.com/balena-io/balena-api/issues/2966

### DIFF
--- a/src/features/device-proxy/device-proxy.ts
+++ b/src/features/device-proxy/device-proxy.ts
@@ -4,6 +4,7 @@ import * as _ from 'lodash';
 import { sbvrUtils, permissions, errors } from '@balena/pinejs';
 import type { Filter } from 'pinejs-client-core';
 
+import type { Device } from '../../balena-model';
 import {
 	captureException,
 	handleHttpErrors,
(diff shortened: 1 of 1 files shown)


src/features/device-proxy/device-proxy.ts
```diff
@@ -184,29 +184,47 @@ export async function requestDevices({
...
+						{
+							is_connected_to_vpn: true,
+							vpn_address: { $ne: null },
+						},
+						filter,
+					],
+				},
+			},
+		})
+	).map(({ id }) => id) as number[];
```
Page-, line 206: I'd prefer this as
```suggestion
	) as Device[]).map(({ id }) => id);
```
or
```suggestion
	) as Pick<Device, 'id'>[]).map(({ id }) => id);
```
so that we type the response closer to the source and therefore benefit from the improved typings for more places (eg after the change typescript would be able to catch a typo like `.map(({ iid }) => iid)` whereas before that would be a runtime error instead)

thgreasi: Makes sense :+1:

Page- (approved):

PR #574, opened 2021-02-05
Add device family and manufacturer resources

sradevski:
I want to add maybe another test or two
Change-type: minor

### DIFF
--- a/src/features/contracts/index.ts
+++ b/src/features/contracts/index.ts
@@ -9,6 +9,7 @@ import {
 	CONTRACTS_PUBLIC_REPO_NAME,
 	CONTRACTS_PUBLIC_REPO_OWNER,
 } from '../../lib/config';
+import { captureException } from '../../infra/error-handling';
 
 const SYNC_INTERVAL = 5 * 60 * 1000;
 
@@ -92,9 +93,9 @@ const mapModel = async (
 		const contractValue =
 			_.get(contractEntry, mapper?.contractField) ?? mapper?.default;
 
-		if (mapper.refersTo) {
+		if (mapper.refersTo && contractValue != null) {
 			try {
-				const [arch] = await rootApi.get({
+				const [entry] = await rootApi.get({
 					resource: mapper.refersTo.resource,
 					options: {
 						$filter: { [mapper.refersTo.uniqueKey]: contractValue },
@@ -102,14 +103,15 @@ const mapModel = async (
 					},
 				});
 
-				mappedModel[key] = arch?.id;
+				mappedModel[key] = entry?.id ?? null;
 			} catch (err) {
 				console.error(
-					`Failed to get contract refer id for field ${key} of resource ${mapper.refersTo.resource}`,
+					`Failed to get contract refer id for field ${key} of resource ${mapper.refersTo.resource}.`,
+					err.message,
 				);
 			}
 		} else {
-			mappedModel[key] = contractValue;
+			mappedModel[key] = contractValue ?? null;
 		}
 	}
 	return mappedModel;
@@ -131,7 +133,10 @@ const upsertEntries = async (
 						resource,
 						body: entry,
 						options: {
-							$filter: { [uniqueField]: entry[uniqueField], $not: entry },
+							$filter: {
+								[uniqueField]: entry[uniqueField],
+								$not: entry,
+							},
 						},
 					});
 				}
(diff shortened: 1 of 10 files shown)


test/09_contracts.ts
```diff
@@ -22,6 +22,23 @@ const clearDeviceTypeResource = () => {
...
+	db: string;
+}) => {
+	const fromContracts = (await getContracts(contract))
+		.map((x) => x.slug)
+		.sort();
+	const fromDb = (await supertest().get(`/resin/${db}`).expect(200)).body.d
+		.map((x: { slug: string }) => x.slug)
+		.sort();
+
+	expect(fromContracts).to.eql(fromDb);
```
Page-, line 39: For the sake of 2 characters I'd find the full word more naturally readable:
```suggestion
	expect(fromContracts).to.equal(fromDb);
```

src/migrations/00043-add-belongs-to-family-dt.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "device type" 
+ADD COLUMN IF NOT EXISTS "belongs to-device family" TEXT NULL;
```
Page-, line 2: I'm assuming this is from a previous version because right now this looks like it should be a foreign key

src/features/contracts/index.ts
```diff
@@ -198,13 +202,19 @@ export const synchronizeContracts = async (contractRepos: RepositoryInfo[]) => {
...
+		'arch.sw',
+		'hw.device-manufacturer',
+		'hw.device-family',
+		'hw.device-type',
+	]) {
 		try {
 			const contracts = await getContracts(contractType);
 			await syncContractsToDb(contractType, contracts, globalSyncSettings);
 		} catch (err) {
 			console.error(
```
Page-, line 215: Should this be a `captureException` that can also make its way to sentry?

src/features/contracts/index.ts
```diff
@@ -92,24 +92,25 @@ const mapModel = async (
...
+				const [entry] = await rootApi.get({
 					resource: mapper.refersTo.resource,
 					options: {
 						$filter: { [mapper.refersTo.uniqueKey]: contractValue },
 						$select: ['id'],
 					},
 				});
 
-				mappedModel[key] = arch?.id;
+				mappedModel[key] = entry?.id;
```
Page-, line 105: In this case `mappedModel[key]` will be `undefined` but below you've explicitly added a `?? null` to make sure it uses `null` rather than `undefined` - does this need the same?

Page- (changes requested):
Request changes is for the migrations

src/features/contracts/index.ts
sradevski: yes I think it does, cause Pine complains of the value is undefined, it expects null instead.

src/features/contracts/index.ts
sradevski: Yeah I think it makes sense here, since it's the top level and if this fails it means something went wrong. For the rest, I think having it in the logs is enough, otherwise it might spam Sentry.

test/09_contracts.ts
sradevski: I agree, I just used it cause I saw it in different places like this

src/migrations/00043-add-belongs-to-family-dt.sql
sradevski: @Page- yup it was my silly mistake, fixed

src/migrations/00043-add-belongs-to-family-dt.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "device type" 
+ADD COLUMN IF NOT EXISTS "belongs to-device family" INTEGER NULL,
+ADD CONSTRAINT "device type_belongs to-device family_fkey" FOREIGN KEY ("belongs to-device family") REFERENCES "device family" ("id");
```
Page-, line 3: I'd expect this to fail because the `device family` table won't exist yet

test/09_contracts.ts
Page-: Ah I hadn't noticed it was moved code, either way I think the full word makes sense (I'd probably even find it quicker to type as well because of muscle memory)

src/migrations/00043-add-belongs-to-family-dt.sql
sradevski: I always forget that the tables need to be created manually :/

Page-: It's only in the specific case where an existing table needs to be migrated to point to the new table, if it was the new table pointing at an existing table then leaving it to be auto-created would work

Page- (approved):

src/migrations/00043-add-belongs-to-family-dt.sql
sradevski: yeah I know, I meant I forgot that the migrations run before the sbvr tables are created, but makes sense :)

PR #580, opened 2021-02-12
Update pinejs to 14.18.1

Page-:
Update pinejs from 14.15.0 to 14.18.1

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #581, opened 2021-02-15
Run tests against the exposed api version rather than internal

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #583, opened 2021-02-15
Add a `boolVar` helper for config to standardize boolean config vars

Page-:
This now requires boolean config vars must be `true` or `false`, or
not set if they're optional, previously we compared against just `true`
so any other values would be treated as `false` which could be
confusing in the case of eg `1`/`on`

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #588, opened 2021-02-26
Fix automatic slug renaming on app rename

Page-:
Change-type: minor


thgreasi (commented):
LGTM :+1: 
I would prefer a commit like "Change application slug format to have an org slug prefix" since for oB users that's probably more important.
How about also adding a migration for consistency with old data?

thgreasi (approved):

PR #589, opened 2021-02-26
Allow pinejs to enforce the at most one app with a given name rule

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

Issue #591, opened 2021-03-01
Error: Could not find service install for device


thgreasi:
See: https://github.com/balena-io/balena-api/issues/3037

PR #592, opened 2021-03-01
Fix release tracking using the new CI/CD metadata fields

thgreasi:
It looks like we previously were updating the release of the app that had the latest start_timestamp.
Draft so that I can collect some thoughts on this while I write some tests.

Resolves: #591
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/L_kD9edlawU6l8f8wvHT0FBUPG_

### DIFF
--- a/src/features/ci-cd/hooks/track-latest-release.ts
+++ b/src/features/ci-cd/hooks/track-latest-release.ts
@@ -1,52 +1,65 @@
 import * as _ from 'lodash';
+import { sbvrUtils, hooks } from '@balena/pinejs';
 
-import { hooks } from '@balena/pinejs';
 const releaseStateFields = [
 	'status',
 	'is_passing_tests',
 	'release_type',
 	'is_invalidated',
 ];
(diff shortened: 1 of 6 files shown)


Page- (dismissed):

src/features/ci-cd/hooks/track-latest-release.ts
```diff
@@ -1,138 +1,165 @@
...
-	if (release == null) {
-		return;
-	}
-	await api.patch({
-		resource: 'application',
-		id: release.belongs_to__application[0].id,
-		options: {
 			$filter: {
-				should_track_latest_release: true,
+				owns__release: {
```
thgreasi, line 57: In the rat I was originally also filtering on `should_be_running__release: true`, which while writing the tests I found that it would break new service install creation for pinned apps.

Page- (approved):

Page-:
@balena-ci rebase

PR #593, opened 2021-03-02
Added contract field to device type resources

mehalter:
One a previous Architecture call we discussed adding partials to the API for device types and releases. This adds the fields to those tables and adds a migration script.

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -57,6 +57,7 @@ async function onInitHooks() {
 		'hw.device-type': {
 			resource: 'device_type',
 			uniqueKey: 'slug',
+			includeRawContract: true,
 			map: {
 				slug: {
 					contractField: 'slug',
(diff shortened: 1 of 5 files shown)


mehalter:
After further discussion, we decided it would be cleaner to put the partials in the contract since we are already storing the contract in the database for releases and the contract is necessary for the context of the partials. We just need to add the contract to device types similar to how they are for releases.

mehalter:
I added an initial stab at the synchronization of the contracts to the database. I will talk with Stevche tomorrow about the approach. Not super happy with trying to make it fit into the current format of specifying an empty string for the `contractField`. If anyone has any suggestions feel free. I will continue thinking about it.

Page- (approved):

mehalter:
@balena-ci rebase

PR #594, opened 2021-03-03
Use already available primary key values in device state-patch

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #595, opened 2021-03-03
Optimize state-patch upserts to use only a single get

Page-:


Page-:
@balena-ci I self-certify!

PR #597, opened 2021-03-05
Update pinejs to pick up rule optimizations for deletes

Page-:
Update pinejs from 14.18.3 to 14.19.0

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #598, opened 2021-03-05
Ensure service installs exist at the point of pinning an application

Page-:
This avoids issues where you pin an application to a release for which
some devices are missing service installs. It also allows simplifying
the "should track latest release" implementation as it no longer needs
to care about creating service installs itself

Change-type: patch

### DIFF
--- a/src/features/ci-cd/hooks/service-installs.ts
+++ b/src/features/ci-cd/hooks/service-installs.ts
@@ -91,13 +91,14 @@ hooks.addPureHook('PATCH', 'resin', 'application', {
 			request.values.should_be_running__release != null &&
 			affectedIds.length !== 0
 		) {
-			// Ensure that every device of the app we've just pinned to a release has the necessary service install entries
+			// Ensure that every device of the app we've just pinned, that is not itself pinned, has the necessary service install entries
 			const devices = (await api.get({
 				resource: 'device',
 				options: {
 					$select: 'id',
 					$filter: {
 						belongs_to__application: { $in: affectedIds },
+						should_be_running__release: null,
 					},
 				},
 			})) as Array<Pick<Device, 'id'>>;
--- a/src/features/ci-cd/hooks/service-installs.ts
+++ b/src/features/ci-cd/hooks/service-installs.ts
@@ -1,6 +1,7 @@
 import * as _ from 'lodash';
 import { sbvrUtils, hooks, permissions } from '@balena/pinejs';
 import type { Filter } from 'pinejs-client-core';
+import type { Device } from '../../../balena-model';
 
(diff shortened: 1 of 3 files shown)


thgreasi (approved):
:+1: 
Seems that we were trying to combine those two steps in one w/o reason.
This single responsibility approach seems way cleaner & easier to reason.

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -1,6 +1,7 @@
 import * as _ from 'lodash';
 import { sbvrUtils, hooks, permissions } from '@balena/pinejs';
 import type { Filter } from 'pinejs-client-core';
+import { Device } from '../../../balena-model';
```
thgreasi, line 4: NIT:
```suggestion
import type { Device } from '../../../balena-model';
```

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -170,3 +171,33 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
 		}
 	},
 });
+
+hooks.addPureHook('PATCH', 'resin', 'application', {
```
thgreasi, line 175: NIT: personally I would prefer to see this hook above the device hooks, following the model in a top-down way

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -170,3 +171,33 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
+	POSTRUN: async ({ api, request }) => {
+		const affectedIds = request.affectedIds!;
+		if (
+			request.values.should_be_running__release != null &&
+			affectedIds.length !== 0
+		) {
+			// Ensure that every device of the app we've just pinned, that is not itself pinned, has the necessary service install entries
+			const devices = (await api.get({
```
thgreasi, line 183: Any reason you didn't just reuse part of the old implementation (doing a pine.get starting from the app) so that we don't do O(N) device GETs?

Page-: Mostly because we have existing code to handle devices one-by-one and I wanted to reuse it - I think there are optimizations that can be made with bundling the queries from per-device into a single large batch which will cover both the app and device hooks but I think that makes sense to be a separate piece of work.

Also these two work by separate optimization methods, the app one fetched all the service installs of all existing devices and then diff'ed that vs the expected and added the missing ones, this code fetches all the services for a device where the service is missing and then creates just those, and I'm not sure which approach actually works out better (bearing in mind that there could be a huge number of pre-existing service installs that we're fetching, and I'd expect a much higher % pre-existing than missing in general)

PR #599, opened 2021-03-05
Fixed JSON serialization of device type contracts in database

mehalter:
PineJS serializes and deserializes JSON fields automatically and serializing before pushing contracts to the database breaks this. This fixes the issue and lets PineJS handle the serialization.

### DIFF
--- a/src/features/contracts/index.ts
+++ b/src/features/contracts/index.ts
@@ -93,7 +93,7 @@ const mapModel = async (
 ) => {
 	const mappedModel: { [k: string]: any } = {};
 	if (includeRawContract) {
-		mappedModel['contract'] = JSON.stringify(contractEntry);
+		mappedModel['contract'] = contractEntry;
 	}
 	for (const key of Object.keys(map) as Array<keyof typeof map>) {
 		const mapper = map[key];
@@ -135,6 +135,10 @@ const upsertEntries = async (
 	await Bluebird.map(
 		newData,
 		async (entry: any) => {
+			const entryQuery =
+				'contract' in entry && entry.contract != null
+					? { ...entry, contract: JSON.stringify(entry.contract) }
+					: entry;
 			try {
 				if (existingData.has(entry[uniqueField])) {
 					return await rootApi.patch({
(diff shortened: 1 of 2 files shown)


test/09_contracts.ts
```diff
@@ -179,7 +179,10 @@ describe('contracts', () => {
 			expect(dbDeviceTypes).to.have.length(14);
 			expect(newDt).to.not.be.undefined;
 			expect(finDt).to.have.property('name', 'Fin');
-			expect(finDt).to.have.property('contract', JSON.stringify(finDtContract));
+			expect(finDt).to.have.deep.property(
+				'contract',
+				JSON.parse(JSON.stringify(finDtContract)),
```
mehalter, line 184: This is necessary to verify the serialization and deserialization correctly because comparing two JSON objects in Chai doesn't ignore fields that are set to "undefined", but the serialization handles this.

src/features/contracts/index.ts
```diff
@@ -135,6 +135,10 @@ const upsertEntries = async (
 	await Bluebird.map(
 		newData,
 		async (entry: any) => {
+			const entryQuery =
+				'contract' in entry && typeof entry.contract === 'object'
```
Page-, line 139: I think you should change this to
```suggestion
				'contract' in entry && entry.contract != null
```
since we would still stringify it on entry if it's a string/number/etc, it's only null that would be left as-is

mehalter: Fixed in new commit

Page- (approved):

test/09_contracts.ts
```diff
@@ -179,7 +179,10 @@ describe('contracts', () => {
 			expect(dbDeviceTypes).to.have.length(14);
 			expect(newDt).to.not.be.undefined;
 			expect(finDt).to.have.property('name', 'Fin');
-			expect(finDt).to.have.property('contract', JSON.stringify(finDtContract));
+			expect(finDt).to.have.deep.property(
```
thgreasi, line 182: @mehalter I don't think that you need to chain `.deep` here.
Deep is used if you are passing, eh, a deep property in the property name argument, eg:
```
expect(finDt).to.have.deep.property('contract.length').that.is.a('number')
```
Doesn't cause any issue here though :)

PR #600, opened 2021-03-08
Update pinejs to pick up rule optimizations for inserts

Page-:
Update pinejs from 14.19.0 to 14.20.0

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #601, opened 2021-03-08
Added optional environment variable configuration for contract repo branch

mehalter:
Add optional ability to change which branch the contracts are pulled for the API.


test/09_contracts.ts
```diff
@@ -49,8 +57,14 @@ describe('contracts', () => {
...
 
-		mockRepo('balena-io', 'contracts', 'base-contracts', true);
-		await synchronizeContracts([{ owner: 'balena-io', name: 'contracts' }]);
+		mockRepo(
+			contractRepository.owner,
+			contractRepository.name,
+			contractRepository.branch,
+			'base-contracts',
+			true,
+		);
```
thgreasi, line 66: NIT: Maybe it now makes sense to convert the arguments of this into an object, to make obvious what each arg is when invoking it and not have to manually spread all properties of the RepositoryInfo.
Eg:
```suggestion
		mockRepo({
			...contractRepository,
			filename: 'base-contracts',
			persist: true,
		});
```
Definitely doesn't have to be in this PR though.

src/features/contracts/contracts-directory.ts
```diff
@@ -81,7 +81,9 @@ const normalizeAssets = async (
 };
 
 const getArchiveLinkForRepo = (repo: RepositoryInfo) => {
-	return `https://api.github.com/repos/${repo.owner}/${repo.name}/tarball`;
+	return `https://api.github.com/repos/${repo.owner}/${repo.name}/tarball/${
+		repo.branch || ''
```
thgreasi, line 85: Most (if not all) `|| ''` could probably be `?? ''` but not a blocker.

thgreasi (commented):
LTGM, feel free to self-certify, even though I would prefer someone to check the toml & the tmpl first.

test/09_contracts.ts
mehalter: Improved this method's parameters in latest push.

src/features/contracts/contracts-directory.ts
mehalter: Ah nice catch. Fixed this in the latest push

thgreasi (approved):

PR #603, opened 2021-03-10
Validate username/email on PATCH as well as POST

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #606, opened 2021-03-10
Only try to validate username/email if they are present

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

Page-:
@balena-ci retest

PR #609, opened 2021-03-12
Cache device existence check for 5 minutes

Page-:
This will reduce the load for the common case where active devices are
not regularly deleted with the potential to briefly respond with a 401
when a device is first deleted. There may also be a slight delay for
restored devices to be able to communicate again however again that is
a very rare case.

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #612, opened 2021-03-12
Emit an error on the request stream for non-200s rather than throwing

Page-:
The throw was in an asynchronous callback and throwing there causes the
process to exit

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #613, opened 2021-03-15
Fix device existence check to be true when the device does exist

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #617, opened 2021-03-16
Update pinejs to 14.22.1

Page-:
Update pinejs from 14.21.1 to 14.22.0

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #618, opened 2021-03-16
Add support for async functions when generating rate limiter keys

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #622, opened 2021-03-23
Optimize application release pinning

thgreasi:
Brings back the optimizations for that part that we
used before #598

Change-type: patch


Page- (changes requested):
I'd rather optimize shared methods rather than having two distinct approaches

PR #625, opened 2021-03-23
Add scheduler mechanism and use it to synchronize contracts

Page-:
I brought the existing scheduler mechanism in from cloud api to use for contracts syncing and exposed it for cloud api to switch over to


Page-:
@balena-ci I self-certify!

PR #626, opened 2021-03-26
Use primitive memoization mode for build-info-facade

Page-:
This can give better performance but requires all arguments to generate
unique strings when stringified

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #627, opened 2021-03-26
Fix contracts sync job to run every 5 minutes instead of 5 seconds

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #629, opened 2021-04-01
Simplify ignored device types handling

Page-:
By handling ignored device types transparently at the point of fetching
it allows us to avoid duplicating that code and guarantees the ignoring
is always applied

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #630, opened 2021-04-02
Support/use multi-level caching that can share cache between instances

Page-:
Change-type: patch


src/infra/cache/index.ts
```diff
@@ -0,0 +1,93 @@
...
+		cacheKey?: string;
+		promise: true;
+		primitive: true;
+		preFetch?: true | number;
+		maxAge: number;
+		max?: number;
+	},
+) => {
+	const {
+		cacheKey = fn.name,
```
thgreasi, line 23: How about enforcing a `featureName.fnName` pattern?

thgreasi (commented):
This looks sooo great.
Wonder how much it will decrease our S3 bill.

src/infra/cache/index.ts
Page-: I don't think it's possible to enforce as there's no way of knowing the feature name, beyond that this is "infra" code so shouldn't really have any specific knowledge and ideally will get moved out of open-balena-api altogether as it should be generically useful

Page-:
@balena-ci I self-certify!

PR #634, opened 2021-04-07
basic implementation of target hostapp release

xginn8:
basic implementation of target hostapp release
    
Change-type: minor

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -546,6 +546,9 @@ Fact type: device is managed by service instance
 Fact type: device should be running release
 	Synonymous Form: release should be running on device
 	Necessity: each device should be running at most one release
+Fact type: device should be operated by release
+	Synonymous Form: release should operate device
+	Necessity: each device should be operated by at most one release
 Fact type: device1 is managed by device2
 	Synonymous Form: device2 manages device1
 	Necessity: each device1 is managed by at most one device2.
@@ -779,3 +782,6 @@ Rule: It is necessary that each application that owns a release1 that has a stat
 Rule: It is necessary that each release that should be running on a device, has a status that is equal to "success" and belongs to an application1 that the device belongs to.
 Rule: It is necessary that each release that should be running on an application, has a status that is equal to "success" and belongs to the application.
 Rule: It is necessary that each release that contains at least 2 images, belongs to an application that has an application type that supports multicontainer.
+Rule: It is necessary that each release that should operate a device, has a status that is equal to "success".
+Rule: It is necessary that each release that should operate a device, belongs to an application that is host and is for a device type that describes the device.
+Rule: It is necessary that each device that should be operated by a release, should be operated by a release that is not invalidated.
--- a/src/features/hostapp/hooks/index.ts
+++ b/src/features/hostapp/hooks/index.ts
@@ -1,0 +1,1 @@
+import './target-hostapp';
--- a/src/migrations/00047-add-target-hostapp.sql
+++ b/src/migrations/00047-add-target-hostapp.sql
@@ -1,0 +1,5 @@
+ALTER TABLE "device"
+ADD COLUMN IF NOT EXISTS "should be operated by-release" INT NULL;
+
+CREATE INDEX IF NOT EXISTS "device_should_be_operated_by_release_idx"
+ON "device" ("should be operated by-release");
--- a/test/fixtures/15-target-hostapps/applications.json
+++ b/test/fixtures/15-target-hostapps/applications.json
@@ -1,0 +1,28 @@
+{
+	"intel-nuc": {
+		"user": "admin",
+		"is_host": true,
+		"is_public": true,
+		"device_type": "intel-nuc",
+		"app_name": "intel-nuc"
+	},
+	"intel-nuc-esr": {
+		"user": "admin",
+		"is_host": true,
+		"is_public": true,
+		"device_type": "intel-nuc",
+		"app_name": "intel-nuc-esr"
+	},
+	"raspberrypi3": {
+		"user": "admin",
+		"is_host": true,
+		"is_public": true,
+		"device_type": "raspberrypi3",
+		"app_name": "raspberrypi3"
+	},
+	"user-app1": {
+		"user": "admin",
+		"device_type": "intel-nuc",
+		"app_name": "user-app1"
+	}
+}
--- a/src/balena-init.sql
+++ b/src/balena-init.sql
@@ -55,6 +55,8 @@ CREATE INDEX IF NOT EXISTS "device_is_running_release_idx"
 ON "device" ("is running-release");
 CREATE INDEX IF NOT EXISTS "device_should_be_running_release_idx"
 ON "device" ("should be running-release");
+CREATE INDEX IF NOT EXISTS "device_should_be_operated_by_release_idx"
+ON "device" ("should be operated by-release");
 
 -- "device config variable"."device" is the first part of an automated unique index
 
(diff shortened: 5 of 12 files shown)


src/features/devices/hooks/target-os.ts
```diff
@@ -0,0 +1,124 @@
...
+	api: sbvrUtils.PinejsClient,
+	osVersion: string,
+	deviceTypeId: string,
+) {
+	return await api.get({
+		resource: 'release',
+		options: {
+			$select: ['id', 'belongs_to__application'],
+			$filter: {
+				// TODO: maybe better to use release tags (to respect variant and version, though bleh)
```
xginn8, line 99: @thgreasi i vaguely remember discussing a plan to migrate hostapps to proper `release_version` usage. do you have a plan for that, or should this be implemented using release tags?

test/fixtures/15-target-hostapps/applications.json
```diff
@@ -0,0 +1,28 @@
...
+		"device_type": "intel-nuc",
+		"app_name": "intel-nuc"
+	},
+	"intel-nuc-esr": {
+		"user": "admin",
+		"is_host": true,
+		"is_public": true,
+		"device_type": "intel-nuc",
+		"app_name": "intel-nuc-esr"
+	},
```
xginn8, line 15: @dfunckt although ESR requires a balenaCloud subscription and plan upgrade, there's no reason to ignore that here in open-balena, right? i.e., folks could still migrate an ESR OS over to open-balena and that should be supported.

dfunckt: I think the guarantees ESR releases offer are only relevant for the cloud by definition and are otherwise just balenaOS. In that sense, there's no reason for open-balena to not be compatible with ESR -- that doesn't imply anything. However, I'm not completely sure what you're asking or what this means technically as it seems the fact that a release is ESR is just semantics implied by the `-esr` suffix in the name. Can you clarify?

xginn8: I guess my question is really if we enforce the "no downgrades" policy in a hook here (like what we do for supervisors), ESR's different versioning scheme will effectively mean a device can never go from ESR -> non-ESR via supported methods. I think there's probably a use case for ESR -> non-ESR upgrades in open-balena, whereby the "no downgrades" policy must be enforced differently.

dfunckt: I think I see. Going from ESR to regular balenaOS and vice versa is less about version downgrade/upgrade and more about migrating between different lines of software similar to, say, balenaOS for RPi and Rasbian -- they just happen to be both super compatible. I would thus encode all invariants into open-balena and think about providing the "downgrade" escape hatch separately -- it could be an action similar to changing the device type of an app.

src/balena.sbvr
```diff
@@ -545,6 +546,9 @@ Fact type: device is managed by service instance
 Fact type: device should be running release
 	Synonymous Form: release should be running on device
 	Necessity: each device should be running at most one release
+Fact type: device should have hostapp release
```
xginn8, line 549: i'm 100% open to better names for this.

dfunckt: Just rubber ducking as I don't have anything to suggest, but how would you describe a hostapp and what does it mean for a device to run one of its releases?

alexgg: I also don't have a better suggestion, but a hostapp is just an image that has some scaffolding so that mobynit can find it, mount it and pivot root into it to launch systemd. I wouldn't say it's a bootable image, as that implies bootloaders and firmware more than mobynit, but maybe mobynit bootstrapped image or similar.

dfunckt: Strawman:

```
Fact type: device is bootstrapped by release
```

Sidenote: the model should enforce that this release is of a hostapp.

xginn8: those suggestions make more sense than what I have! thanks for thinking through it. we do enforce that the related release belongs to a hostapp here: https://github.com/balena-io/open-balena-api/pull/634/files#diff-ab9e54b98b8703135b751e1127edf11c3c5e3b48e3e9892d8acc07c12767ecb5R786

xginn8: I went with `Fact type: device is initialized by release` as it felt more natural to me, thanks again for the input!

src/features/devices/hooks/target-hostapp.ts
```diff
@@ -0,0 +1,277 @@
...
+
+	const newHostappRelease = await api.get({
+		resource: 'release',
+		id: newHostappReleaseId,
+		options: {
+			$expand: {
+				release_tag: { $filter: { tag_key: 'version' }, $select: ['value'] },
+			},
+			$filter: {
+				is_invalidated: false,
```
xginn8, line 231: we shouldn't be able to _upgrade_ to an invalid release, but we can provision to one (why it's not an SBVR, but filtered here).

xginn8: do you mean express this as a rule, or something different?

src/features/devices/hooks/target-hostapp.ts
```diff
@@ -0,0 +1,277 @@
...
+								id: {
+									$in: deviceIds,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	});
```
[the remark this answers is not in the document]
xginn8: this is pretty nifty, well spotted!

src/features/devices/hooks/target-hostapp.ts
Page-: Well expressing precisely what we allow would be the first step, with the ideal outcome being that it can be expressed as an SBVR rule (and if we can't then I expect it'll be due to an element missing from our model rather than a fundamental blocker)

xginn8: this is an issue we've discussed in the past generally speaking -- that we don't want to prevent a device from coming online with a known-bad version (i.e. creates are allowed), but we want to prevent other devices from upgrading to them (updates are disallowed). this is true for both user apps as well as infrastructure apps like the supervisor or hostapp.

xginn8: discussion about converting this into a rule here: https://www.flowdock.com/app/rulemotion/resin-tech/threads/Fvz_ycKmosdRhCX61uMLmXlK0o-

Page- (commented):
LGTM outside of a bunch of TODOs

src/features/devices/hooks/target-hostapp.ts
```diff
@@ -0,0 +1,296 @@
+import {
```
Page-, line 1: Does this make sense to move into a `hostapp` feature instead of the fairly generic `devices` feature?

src/migrations/00047-add-target-hostapp.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE "device"
+ADD COLUMN IF NOT EXISTS "should be operated by-release" INT NULL;
+
+CREATE INDEX IF NOT EXISTS "device_should_be_operated_by_release_idx"
+ON "device" ("should be operated by-release");
```
Page-, line 5: This needs a balena-init.sql entry as well

src/features/devices/hooks/target-hostapp.ts
xginn8: yeah, that's fine with me

src/features/hostapp/hooks/index.ts
```diff
@@ -0,0 +1 @@
+import './target-hostapp';
```
Page-, line 1: You can remove this file altogether since the hooks are imported directly not via this index

Page- (approved):

Page- (approved):

PR #638, opened 2021-04-13
Allow composition of a release to be initially `null`

dfunckt:
This permits clients to create the release resource far earlier then currently possible — before even inspecting it — which in turn allows it to be monitored sooner.

Change-type: minor


Page- (approved):

dfunckt:
This isn't really needed -- we already have a way to represent lack of a composition with `{}`.

PR #641, opened 2021-04-15
Stop allowing cache clearing for multi-level cache

Page-:
This causes problems with the redis level of the cache because clearing
the cache requires clearing the entire redis database which is far more
aggressive than we want to support

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #642, opened 2021-04-16
initial conversion of `should track latest release` int -> enum

xginn8:
related to
https://www.flowdock.com/app/rulemotion/resin-tech/threads/Odo2hTtX2f-dnzQkTKRu5mwYZIm,
this change will allow specifying that an application can track only the
latest _stable_ release, thus allowing draft + untested release status
as opt-in release controls

Change-type: minor

### DIFF
--- a/src/balena-init.sql
+++ b/src/balena-init.sql
@@ -13,8 +13,8 @@ ALTER TABLE "device"
 ALTER COLUMN "api heartbeat state" SET DEFAULT 'unknown';
 
 ALTER TABLE "release"
-ALTER COLUMN "release type" SET DEFAULT 'final',
-ALTER COLUMN "is passing tests" SET DEFAULT 1;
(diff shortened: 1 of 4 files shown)


dfunckt:
@xginn8 just a heads up that once this is merged you’ve created a hard cutoff point beyond which open-balena cannot refer until we also bump the OData version to v7.

xginn8:
> @xginn8 just a heads up that once this is merged you’ve created a hard cutoff point beyond which open-balena cannot update its open-balena-api reference until we also bump the OData version to v7. This shouldn’t be a blocker for you, but you may have to coordinate the various parts of this work so that the bump to v7 is not significantly delayed.

Thanks for pointing that out. Since this requires a version bump and that's not something I'm too familiar with I'll definitely be asking for guidance/help if that's okay!

PR #649, opened 2021-04-29
Update dependencies

Page-:
Update pinejs from 14.22.1 to 14.22.2

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #650, opened 2021-04-29
Remove unnecessary delete hook for env var validation

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #652, opened 2021-04-30
Revert "Expose the RPi family config vars to raspberrypi3-unipi-neuron"

floion:
This reverts commit db990479be2611f24b6262e12e1dfa092473cea0.

OpenBalena is not the place to do the config vars change for the private
device types such as the UniPi Neuron so we revert this commit here.

Change-type: minor


Page- (approved):

Page-:
@balena-ci rebase

PR #654, opened 2021-04-30
Fix 'dependents' on state not being returned

toochevere:
Signed-off-by: PJ <pj@balena.io>
Change-type: patch

### DIFF
--- a/src/features/device-state/routes/state.ts
+++ b/src/features/device-state/routes/state.ts
@@ -37,6 +37,11 @@ export type App = {
 	networks: AnyObject;
 };
 
+export type Dependent = {
+	apps: AnyObject;
+	devices: AnyObject;
+};
+
 export type AppService = {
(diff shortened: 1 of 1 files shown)


xginn8:
general comment, if this failed due to a failing test in the closed API, can we move that here/replicate it?

dfunckt:
Absolutely, the more tests the better and I can't think of anything (especially in the test suite) that would cause any harm if it was open-sourced when it shouldn't, so I think there's no reason to hold back.

Page- (approved):

toochevere:
> general comment, if this failed due to a failing test in the closed API, can we move that here/replicate it?

I was thinking the same thing, but didn't want to add more to this PR than what was strictly necessary to fix the issue.  I think that there are probably a whole host of tests that can move from the closed to the open API.

PR #655, opened 2021-05-02
Updating to assigna custom role based on variable instead of default-user-role

nitishagar:
Ref: https://github.com/balena-io/balena-api/pull/2913

Change-Type: patch

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -59,6 +59,8 @@ import {
 	comparePassword,
 	registerUser,
 	setPassword,
+	GetNewUserRoleFunction,
+	setRegistrationRoleFunc,
 	validatePassword,
 	checkUserPassword,
 } from './infra/auth/auth';
@@ -272,6 +274,8 @@ export interface SetupOptions {
 	onLogin?: (
 		user: Pick<DbUser, typeof defaultFindUser$select[number]>,
 	) => PromiseLike<void> | void;
+
+	getNewUserRole?: GetNewUserRoleFunction;
 }
 
 export async function setup(app: Application, options: SetupOptions) {
@@ -360,6 +364,10 @@ export async function setup(app: Application, options: SetupOptions) {
 
 	await pine.init(app, options.config);
 	await options.onInitModel?.(app);
+
+	if (options.getNewUserRole) {
+		setRegistrationRoleFunc(options.getNewUserRole);
+	}
 
 	await import('./hooks');
 	await options.onInitHooks?.(app);
--- a/src/infra/auth/auth.ts
+++ b/src/infra/auth/auth.ts
@@ -22,6 +22,19 @@ export const userHasPermission = (
 	return user.permissions.includes(permission);
 };
 
+export type GetNewUserRoleFunction = (user: AnyObject) => string;
+
+let getNewUserRole: GetNewUserRoleFunction = () => 'default-user';
+
+export const setRegistrationRoleFunc = (
+	registerFunc: GetNewUserRoleFunction,
+) => {
+	getNewUserRole = registerFunc;
+};
+
+export const getRegistrationRole: typeof getNewUserRole = (user) =>
+	getNewUserRole(user);
+
 /**
  * A known invalid comparisson to emulate a wrong password error.
  * Used to prevent exposing information via timing attacks.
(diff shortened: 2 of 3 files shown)


dfunckt (changes requested):
There's no point declaring a role here that is not used by the core API -- unless there's use for it, it should be defined in the cloud API.

src/features/auth/hooks/assign-user-role.ts
```diff
@@ -14,7 +15,7 @@ hooks.addPureHook('POST', 'resin', 'user', {
 				req: permissions.root,
 			},
 			id: {
-				name: 'default-user',
+				name: getNewUserRoleName() || 'default-user',
```
dfunckt, line 18: It's impossible for `getNewUserRoleName()` to return a null-ish value, so:

```suggestion
				name: getNewUserRoleName(),
```

src/infra/auth/auth.ts
```diff
@@ -22,6 +23,17 @@ export const userHasPermission = (
...
+let getNewUserRole: ((user: AnyObject) => string) | undefined;
+
+export const setRegistrationRoleFunc = (
+	registerFunc: (user: AnyObject) => string,
+) => {
+	getNewUserRole = registerFunc;
+};
+
+export const getRegistrationRole = (user: AnyObject) =>
+	getNewUserRole && getNewUserRole(user);
```
thgreasi, line 35:
```suggestion
export let getNewUserRole: ((user: AnyObject) => string) = () => 'default-user';

export const setRegistrationRoleFunc = (
	registerFunc: (user: AnyObject) => string,
) => {
	getNewUserRole = registerFunc;
};
```

thgreasi (commented):
And with that ^ and Akis' comments fixed, LGTM.
Not convinced I like the new name of `assign-default-user-role.ts`, since this is still the code that assigns the default registration role, regardless of what the role provider returns.

src/infra/auth/auth.ts
```diff
@@ -22,6 +23,16 @@ export const userHasPermission = (
...
 
+let getNewUserRole: (user: AnyObject) => string = () => 'default-user';
+
+export const setRegistrationRoleFunc = (
+	registerFunc: (user: AnyObject) => string,
+) => {
+	getNewUserRole = registerFunc;
+};
+
+export const getRegistrationRole = (user: AnyObject) => getNewUserRole(user);
```
thgreasi, line 34: I guess that's just so that no-one else can modify `getNewUserRole` right?

thgreasi (commented):
LGTM

src/index.ts
```diff
@@ -272,6 +273,8 @@ export interface SetupOptions {
 	onLogin?: (
 		user: Pick<DbUser, typeof defaultFindUser$select[number]>,
 	) => PromiseLike<void> | void;
+
+	getNewUserRole?: (user: AnyObject) => string;
```
dfunckt, line 277: As per `GetUrlFunction` and `SetupFunction` above, let's give the callback a name that can be imported by clients:

```ts
  // src/infra/auth/auth.ts
  export type GetNewUserRoleFunction = (user: AnyObject) => string;
  
  // this file:
  getNewUserRole?: GetNewUserRoleFunction;
```

dfunckt (commented):
Great stuff! My comment is optional, feel free to ignore.

src/infra/auth/auth.ts
```diff
@@ -8,6 +8,7 @@ import { User } from './jwt-passport';
 
 import { getIP } from '../../lib/utils';
 import type { User as DbUser } from '../../balena-model';
+import { AnyObject } from 'pinejs-client-core';
```
Page-, line 11: This is unnecessary - it's globally available
```suggestion
```

src/infra/auth/auth.ts
```diff
@@ -22,6 +23,16 @@ export const userHasPermission = (
 	return user.permissions.includes(permission);
 };
 
+let getNewUserRole: (user: AnyObject) => string = () => 'default-user';
+
+export const setRegistrationRoleFunc = (
+	registerFunc: (user: AnyObject) => string,
```
Page-, line 29:
```suggestion
	registerFunc: typeof getNewUserRole,
```

src/infra/auth/auth.ts
Page-, line 34: I think you can inherit types like:
```suggestion
export const getRegistrationRole: typeof getNewUserRole = (user) => getNewUserRole(user);
```

dfunckt (approved):

PR #656, opened 2021-05-03
hostapps: remove rule enforcing valid hostapp releases

xginn8:
We already enforce this in the hook itself, and this rule prevents
devices that have provisioned with an invalidated hostapp release from
ever PATCHing/updating after the fact. we ideally only want to enforce
this rule at upgrade time, rather than always

Change-type: patch


xginn8:
@balena-ci rebase

xginn8:
@balena-ci rebase

xginn8:
@balena-ci I self-certify!

PR #657, opened 2021-05-03
hostapp: allow devices to provision & set target state via POST

xginn8:
Change-type: patch

### DIFF
--- a/src/features/hostapp/hooks/target-hostapp.ts
+++ b/src/features/hostapp/hooks/target-hostapp.ts
@@ -58,13 +58,16 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
 	},
 });
 
-hooks.addPureHook('PATCH', 'resin', 'device', {
-	/**
-	 * When a device checks in with it's initial OS version, set the corresponding should_be_operated_by__release resource
-	 * using its current reported version.
-	 */
+/**
+ * When a device checks in with it's initial OS version, set the corresponding should_be_operated_by__release resource
+ * using its current reported version.
+ */
+const hostappHook: hooks.Hooks = {
 	async PRERUN(args) {
-		if (args.request.values.os_version != null) {
+		if (
+			args.request.values.os_version != null &&
+			args.request.values.os_variant != null
+		) {
 			const ids = await sbvrUtils.getAffectedIds(args);
 			await setOSReleaseResource(
 				args.api,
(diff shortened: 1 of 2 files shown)


src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -55,7 +58,46 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
+			args.request.values.os_variant != null
+		) {
+			let deviceType;
+			if (!args.request.values.is_of__device_type) {
+				[deviceType] = await args.api.get({
+					resource: 'device_type',
+					options: {
+						$select: 'id',
+						$filter: {
+							slug: { $eq: args.request.values.device_type },
```
Page-, line 76: `$eq` is normally unnecessary:
```suggestion
							slug: args.request.values.device_type,
```

src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -55,7 +58,46 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
+
+			const hostappRelease = await getOSReleaseResource(
+				args.api,
+				args.request.values.os_version,
+				args.request.values.os_variant,
+				args.request.values.is_of__device_type ?? deviceType?.id,
+			);
+			if (hostappRelease && hostappRelease.length > 0) {
+				args.request.values.should_be_operated_by__release =
+					hostappRelease[0].id;
```
Page-, line 93: Can the `PATCH` hook do this the same way?

xginn8: the `PATCH` hook also enforces some extra constraints (not invalidated and not a downgrade), whereas the `POST` is less picky. do you mean specifically how the return value of `getOSReleaseResource` is used and validated?

Page-: Yeah, the other bits can be specific to the PATCH as need be but this bit of setting the default seems like it should be the same for both?

src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -55,7 +58,46 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
+						$select: 'id',
+						$filter: {
+							slug: args.request.values.device_type,
+						},
+					},
+				});
+			}
+			if (!deviceType) {
+				return;
+			}
```
Page-, line 83: Oh I just realized I made a mistake here, this'll break if there's no match
```suggestion
				const [ dt ] = await args.api.get({
					resource: 'device_type',
					options: {
						$select: 'id',
						$filter: {
							slug: args.request.values.device_type,
						},
					},
				});
			}
			if (!dt) {
				return;
			}
			deviceType = dt.id
```

xginn8:
@balena-ci I self-certify!

PR #667, opened 2021-05-07
Use `true = false` rather than `1 = 2` for generating a false result

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #669, opened 2021-05-10
hostapps: unset a target hostapp when changing device types

xginn8:
Connects-to: https://github.com/balena-io/open-balena-api/issues/668
Change-type: patch


xginn8:
the only alternative i can think of would be to force device type changes to include a HUP / selecting a new hostapp to target / automatically picking the closest one for the new device type that is greater, but all of those are clunky and this value isn't _strictly_ required yet anyway

src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -39,6 +39,22 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
+	 * TODO: changing device types presents us with a bit of a conundrum. since we cannot guarantee parity between
+	 * available OSes for each device type, we cannot just switch the target to the new device type, so instead let's just
+	 * unset the value
+	 */
+
+	async POSTPARSE(args) {
+		if (args.request.values.is_of__device_type != null) {
+			args.request.values.should_be_operated_by__release = null;
+		}
+	},
```
Page-, line 55: It'd be good to check that an explicit `should_be_operated_by__release` has not been set, so that someone can change the device type + pin to a new host release at the same time
```suggestion
	async POSTPARSE({ request }) {
		if (request.values.is_of__device_type != null && request.values.should_be_operated_by__release === undefined) {
			request.values.should_be_operated_by__release = null;
		}
	},
```

Page- (approved):

PR #670, opened 2021-05-10
Require public apps to be of a non-legacy type

thgreasi:
Also checked that there are no violating entries in prod or staging atm.

Change-type: minor
See: https://github.com/balena-io/balena-api/issues/3129


src/balena.sbvr
```diff
@@ -781,6 +781,7 @@ Fact type: config has description (Auth)
 Rule: It is necessary that each device1 that is managed by a device2, belongs to an application1 that depends on an application2 that owns the device2.
 Rule: It is necessary that each release that has a release version1 and has a status that is equal to "success" and is not invalidated, belongs to an application that owns exactly one release that has a release version2 that is equal to the release version1 and has a status that is equal to "success" and is not invalidated.
 Rule: It is necessary that each image that has a status that is equal to "success", has a push timestamp.
+Rule: It is necessary that each application that is public, has an application type that is not legacy.
```
thgreasi, line 784: Having the `is public` in the first part seems to be faster & w/ less complexity, probably b/c of the `is public` index.

PR #672, opened 2021-05-12
Refactoring in preparation for supervisor service-installs

toochevere:
Change-type:  patch

This PR preps the ground for creating service installs for the supervisor.

### DIFF
--- a/src/features/supervisor-app/hooks/supervisor-app.ts
+++ b/src/features/supervisor-app/hooks/supervisor-app.ts
@@ -1,12 +1,12 @@
 import * as semver from 'balena-semver';
 import * as _ from 'lodash';
-
 import {
 	sbvrUtils,
 	hooks,
 	permissions,
 	errors as pinejsErrors,
 } from '@balena/pinejs';
+import { Release } from '../../../balena-model';
 
 const { BadRequestError } = pinejsErrors;
 
@@ -18,8 +18,10 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
 	async PRERUN(args) {
 		if (args.request.values.supervisor_version != null) {
 			const ids = await sbvrUtils.getAffectedIds(args);
-			await setSupervisorReleaseResource(
-				args.api,
+			const { api } = args;
+
+			await setSupervisorResources(
+				api,
 				ids,
 				args.request.values.supervisor_version,
 			);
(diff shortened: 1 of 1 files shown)


src/features/supervisor-app/hooks/supervisor-app.ts
```diff
@@ -18,8 +17,10 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
 	async PRERUN(args) {
 		if (args.request.values.supervisor_version != null) {
 			const ids = await sbvrUtils.getAffectedIds(args);
-			await setSupervisorReleaseResource(
-				args.api,
+			const api = args.api;
```
dfunckt, line 20: This is purely stylistic, ignore if you want:

```suggestion
			const { api } = args;
```

src/features/supervisor-app/hooks/supervisor-app.ts
```diff
@@ -116,14 +115,30 @@ async function checkSupervisorReleaseUpgrades(
...
 			throw new BadRequestError(
 				`Attempt to downgrade supervisor, which is not allowed`,
 			);
 		}
 	}
+
+	return doingUpgrade;
+}
+
+async function getRelease(api: sbvrUtils.PinejsClient, releaseId: number) {
```
dfunckt, line 134: Can you please type the return value? This will then let you drop AnyObject from a couple of places.

src/features/supervisor-app/hooks/supervisor-app.ts
```diff
@@ -116,14 +115,30 @@ async function checkSupervisorReleaseUpgrades(
...
+	return await api.get({
+		resource: 'release',
+		id: releaseId,
```
dfunckt, line 137: As I understand it, this query could return any release and the clients calling it just assume it is from the Supervisor app. Can we improve on this?

src/features/supervisor-app/hooks/supervisor-app.ts
```diff
@@ -221,31 +236,41 @@ async function setSupervisorReleaseResource(
...
-				},
-				body: {
-					should_be_managed_by__release: supervisorRelease.id,
-				},
-			});
 		}),
 	);
 }
+
+function setSupervisorRelease(
```
dfunckt, line 260: This is a trivial function and only used from one place -- does it have to be separate?

src/features/supervisor-app/hooks/supervisor-app.ts
toochevere: Use of this method will change a bit with the next PR.  If it still seems like we are making unnecessarily wide assumptions, let's address it there?

src/features/supervisor-app/hooks/supervisor-app.ts
toochevere: The next PR will make it more clear why we are refactoring this into a separate function.  The purpose is to reduce the cyclomatic complexity of the calling method, which in the next PR will start to grow otherwise.

src/features/supervisor-app/hooks/supervisor-app.ts
```diff
@@ -35,36 +37,45 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
...
+				return;
+			}
+
+			const release = await getRelease(args.api, releaseId);
+
+			if (!release) {
+				throw new BadRequestError(
+					`Could not find a supervisor release with this ID ${releaseId}`,
+				);
+			}
```
Page-, line 67: This made more sense where it was?

src/features/supervisor-app/hooks/supervisor-app.ts
```diff
@@ -116,14 +116,33 @@ async function checkSupervisorReleaseUpgrades(
...
+		if (semver.gt(newSupervisorVersion, oldVersion)) {
+			doingUpgrade = true;
+		} else if (semver.lt(newSupervisorVersion, oldVersion)) {
 			throw new BadRequestError(
 				`Attempt to downgrade supervisor, which is not allowed`,
 			);
 		}
 	}
+
+	return doingUpgrade;
```
Page-, line 132: This is unused anywhere so is essentially adding dead code

toochevere:
The changes around this refactoring are being moved.

PR #677, opened 2021-05-25
Use the built-in nodejs uuid generation in favor of an npm module

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

thgreasi:
How about changing v as well?
See: https://github.com/balena-io/open-balena-api/blob/72eaa0c4b22ea8f82b09a0896a96360fa7546f79/src/features/registry/registry.ts#L257

Page-:
I must have missed that somehow, here it is: https://github.com/balena-io/open-balena-api/pull/679

PR #678, opened 2021-05-25
Add missing DB constraint on allowed release_type values

thgreasi:
Change-type: patch


src/migrations/00051-release-type-value-constraint.sql
```diff
@@ -0,0 +1,27 @@
+
```
Page-, line 1: It'd be nice to drop the empty line here

Page- (approved):

PR #680, opened 2021-06-01
Only delete service installs for the current app when moving

Page-:
Change-type: patch

### DIFF
--- a/src/features/ci-cd/hooks/service-installs.ts
+++ b/src/features/ci-cd/hooks/service-installs.ts
@@ -107,6 +107,43 @@ const createAppServiceInstalls = async (
 		},
 	});
 
+const deleteServiceInstallsForCurrentApp = (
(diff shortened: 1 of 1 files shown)


xginn8 (approved):
nice fix, lgtm

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -107,6 +107,43 @@ const createAppServiceInstalls = async (
...
+							s: {
+								application: {
+									$any: {
+										$alias: 'a',
+										$expr: {
+											a: {
+												owns__device: {
+													$any: {
+														$alias: 'd',
+														$expr: { d: { id: { $in: deviceIds } } },
```
Page-, line 132: I modified this to avoid needing to separately query the app ids but it is otherwise extracted from https://github.com/balena-io/open-balena-api/pull/504

PR #684, opened 2021-06-03
Further target state typing improvements

Page-:
This inverts it to have the state as one large object to make it much easier to see the format in its entirety with named aliases into that overall object where it's useful


Page-:
@balena-ci I self-certify!

PR #687, opened 2021-06-04
WIP: v3 state

Page-:


src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,437 @@
...
+		apps: {
+			[uuid: string]: {
+				/**
+				 * @deprecated to be removed in state v4
+				 */
+				id: number;
+				name: string;
+				parent_app?: string;
+				is_host?: boolean;
+				config: {
```
pipex, line 53: I thought we were keeping this config on the device level

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,437 @@
...
+								labels: {
+									[labelName: string]: string;
+								};
+								contract?: AnyObject;
+								composition?: ServiceComposition;
+							};
+						};
+					};
+				};
+				volumes?: AnyObject;
```
pipex, line 89: Should volumes go within releases?

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,437 @@
...
+	varList.forEach(({ name, value }) => {
+		if (filterFn(name)) {
+			obj[name] = value;
+		}
+	});
+};
+export const rejectUiConfig = (name: string) =>
+	!/^(BALENA|RESIN)_UI/.test(name);
+
+type LocalStateApp = StateV3['local']['apps'][string];
```
Page-, line 38:
```suggestion
type LocalStateApp = StateV3[string]['apps'][string];
```

src/features/device-state/routes/state-v3.ts
Page-: You're right, I'll get this fixed

src/features/device-state/routes/state-v3.ts
Page-: You're right, I'll get this fixed

pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,438 @@
...
+			id: release.id,
+			...(Object.keys(services).length > 0 ? { services } : undefined),
+		},
+	};
+	if (composition.networks != null) {
+		stateApp.releases[release.commit].networks = composition.networks;
+	}
+	if (composition.volumes != null) {
+		stateApp.releases[release.commit].volumes = composition.volumes;
+	}
```
thgreasi, line 197:
```suggestion
	stateApp.releases = {
		[release.commit]: {
			id: release.id,
			...(Object.keys(services).length > 0 && { services }),

			...(composition.networks != null && { networks: composition.networks }),
			...(composition.volumes != null && { volumes: composition.volumes }),
			// or maybe
			..._.pick(composition, ['networks', 'volumes'])
		},
	};
```

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,438 @@
...
+	return stateApp;
+}
+
+// These 2 config vars below are mapped to labels if missing for backwards-compatibility
+// See: https://github.com/resin-io/hq/issues/1340
+const ConfigurationVarsToLabels = {
+	RESIN_SUPERVISOR_UPDATE_STRATEGY: 'io.resin.update.strategy',
+	RESIN_SUPERVISOR_HANDOVER_TIMEOUT: 'io.resin.update.handover-timeout',
+};
```
thgreasi, line 206: I guess this can be moved to utils as a shared thing.

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,438 @@
...
+				/**
+				 * @deprecated to be removed in state v4
+				 */
+				id: number;
+				name: string;
+				class: 'fleet' | 'block' | 'app';
+				parent_app?: string;
+				is_host?: boolean;
+				releases?: {
+					[uuid: string]: {
```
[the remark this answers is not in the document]
Page-: It is more accurately a uuid than a commit and the name in the model should be updated to reflect that (it used to be a commit hash, now it's a uuid that might be a commit hash in some cases)

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,438 @@
...
+			application_config_variable: {
+				$select: ['name', 'value'],
+				$orderby: {
+					name: 'asc',
+				},
+			},
+			should_be_running__release: releaseExpand,
+		},
+	},
+	should_be_managed_by__release: {
```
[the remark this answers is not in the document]
Page-: I'd say yes, dependent devices should match regular device state and can be forwarded on, eg a dependent device could in theory just be a "normal" device that gets its updates via a single internet connected gateway device - it doesn't have to be a microcontroller or anything special

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,438 @@
...
+const stateQuery = _.once(() =>
+	api.resin.prepare<{ uuid: string }>({
+		resource: 'device',
+		id: { uuid: { '@': 'uuid' } },
+		options: {
+			$select: ['device_name', 'os_version'],
+			$expand: {
+				...deviceExpand,
+				manages__device: {
+					$select: ['uuid', 'device_name', 'belongs_to__application'],
```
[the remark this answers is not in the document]
Page-: Yes it can, good catch. This would have been a leftover from v2 dependent devices which didn't $expand

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,438 @@
...
+		},
+	},
+};
+
+const stateQuery = _.once(() =>
+	api.resin.prepare<{ uuid: string }>({
+		resource: 'device',
+		id: { uuid: { '@': 'uuid' } },
+		options: {
+			$select: ['device_name', 'os_version'],
```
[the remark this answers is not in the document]
Page-: @pipex do modern supervisors even exhibit this issue with `RESIN_HOST_LOG_TO_DISPLAY`? I'd assume we removed that var from the supervisor altogether?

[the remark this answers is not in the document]
Page-: And for that matter I assume the UI no longer shows that config var @thgreasi ?

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,438 @@
...
+			$expand: {
+				...deviceExpand,
+				manages__device: {
```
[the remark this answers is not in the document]
Page-: Nope, as I said elsewhere dependent device's state is meant to match gateway devices so we can treat them essentially the same and remove a bunch of special cases due to it

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,438 @@
...
+			},
+			should_be_running__release: releaseExpand,
+		},
+	},
+	should_be_managed_by__release: {
+		...releaseExpand,
+		$expand: {
+			...releaseExpand.$expand,
+			belongs_to__application: {
+				$select: ['id', 'uuid', 'app_name', 'is_host', 'is_of__class'],
```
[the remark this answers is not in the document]
Page-: I think there was a plan to add the host os release info as well but I can't remember precisely where that went, cc @pipex?  It's possible we kept just the one as there was talk of merging host os and supervisor into a multicontainer app

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,438 @@
...
+		};
+		apps: {
+			[uuid: string]: {
+				/**
+				 * @deprecated to be removed in state v4
+				 */
+				id: number;
+				name: string;
+				class: 'fleet' | 'block' | 'app';
+				parent_app?: string;
```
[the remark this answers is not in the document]
Page-: So it's meant to be `is_managed_by__device` (and is in the patch PR) but doesn't seem to be used in either PR - @pipex do you remember about this?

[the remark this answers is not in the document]
pipex: I think `parent_app` is meant to be for dependent devices. I suppose the dependent app has a parent app. We are not using it for regular updates anywhere on the supervisor right now.

src/features/device-state/routes/state-v3.ts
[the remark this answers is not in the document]
pipex: @Page- or now we cannot merge the supervisor and the hostOS into a multicontainer app as we don't have the deployment process in place. I think @thgreasi is right, we need to expand `should_be_operated_by__release` as well so the supervisor receives the full tri-app state. 

We will definitely need the target state of the host app for hostapp extensions which is why we started this whole process :p

src/features/device-state/routes/state-v3.ts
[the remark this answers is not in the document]
pipex: The `RESIN_HOST_LOG_TO_DISPLAY` only exists in some test files in the supervisor codebase, no references to it on any of the config code. This is the first time I'm hearing about it too

src/features/device-state/routes/state-v2.ts
```diff
@@ -197,13 +186,6 @@ function buildAppFromRelease(
 	};
 }
 
-// These 2 config vars below are mapped to labels if missing for backwards-compatibility
-// See: https://github.com/resin-io/hq/issues/1340
-const ConfigurationVarsToLabels = {
-	RESIN_SUPERVISOR_UPDATE_STRATEGY: 'io.resin.update.strategy',
```
pipex, line 203: Is this being moved somewhere else, or are you removing it entirely? This is still advertised as the main way to controlled the update strategy in our docs https://www.balena.io/docs/learn/deploy/release-strategy/update-strategies/#controlling-the-update-strategy

pipex: Ignore me, I saw the code was moved to utils

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,412 @@
...
+
+		const imgRegistry =
+			image.is_stored_at__image_location +
+			(image.content_hash != null ? `@${image.content_hash}` : '');
+
+		services[svc.service_name] = {
+			id: svc.id,
+			image_id: image.id,
+			image: formatImageLocation(imgRegistry),
+			environment,
```
pipex, line 151: The way this is setup, the environment and labels are sent twice as part of the target state. See the extract below for a test service. I am making sure the supervisor merges them correctly to use the right value, but it might be worth merging them here to reduce the payload size

```
"hello": {
    "id": 5,
    "image_id": 37,
    "image": "registry2.09e8bcd18440f7403fb526350f0350dc.bob.local/v2/c293641a12f4a18c0936e3c4d81debbb@sha256:56405843eda8dc5d6917430214e393825604f1de4927d59c5733ae777a00fbc9",
    "environment": {
      "MYVAR": "goodbye"
    },
    "labels": {},
    "composition": {
      "image": "my-custom-image",
      "environment": {
        "MYVAR": "hello"
      }
    }
  }
```

Page-: @pipex the service level `environment` key includes image_environment_variable/application_environment_variable/service_environment_variable/device_environment_variable/device_service_environment_variable whereas the composition one only includes what is in the docker-compose.yml - I think the composition ones get brought into the api as image environment variables and so would show in both but they are definitely not the definitive list

pipex: I don't understand what you are saying. I don't think that creates a problem as we know the priority, right? 

What I'm thinking is doing something like 

```
service[name]: {
    environment: {...composition.environment, ...environment}
}
``` 

which I'm [doing on the supervisor](https://github.com/balena-os/balena-supervisor/pull/1765/files#diff-c436eb4642e9d9fe975f53422f677d3b3c483de880eec12a003d401d67884f54R153-R154), but doing it on the API side would prevent sending extra data.

Page-: Everything in the composition is already sent in the upper level environment @pipex with the correct priority. The reason for that is because the variables in the composition get added to the odata model on build. However it is then also possible to remove/update those entries from the model after the fact in which case we probably don't want to use what was initially set in the composition as they were explicitly changed.

thgreasi (commented):
LGTM :+1: 
The `os_version` and `is_(managed|operated)_by__release` can potentially be done as separate minor PRs after the structure change (even though it would be nice to server the tri-app state right away ;) )

src/features/device-state/routes/state-v3.ts
thgreasi: I think I'm missing the point as well :confused: 
Why does the supervisor needs to know both the effective env var values, as well as the image env vars from the docker-compose.yml separately? What is the use case that this enables for the supervisor?

Page-: There's 2 separate things
1. the supervisor doesn't need to merge because the api already handles that
2. does the supervisor need the fully intact composition as was used to build? because if not there's probably a bunch of sections we can delete but I thought it was used for things? maybe contracts based on the composition? maybe livepush stuff? etc (I'm not really sure but if the supervisor doesn't actually care about having an intact composition we could probably delete a bunch)

pipex: > the supervisor doesn't need to merge because the api already handles that

Got it, so we can only use the top-level composition

> does the supervisor need the fully intact composition as was used to build?

It does use the whole composition, except where it overrides the values with the cloud state, i.e. environment and labels. 

I see the point though. In local mode, it would probably make sense to put the environment variables in the composition as there are no device or fleet level overrides. So from a modeling perspective, the target state I shared above is more correct, although is more wasteful.

If we want to reduce the waste that might mean we need to tradeoff modeling correctness (which imho is reasonable)
- Move the environment/labels to inside the composition
- Remove the environment/labels from the composition since they are already mirrored in the top level property

I lean more towards the second option, which still allows to provide the composition properties if we ever need to

test/03_device-state.ts
```diff
@@ -38,296 +38,315 @@ configMock['API_HEARTBEAT_STATE_TIMEOUT_SECONDS'] = TIMEOUT_SEC;
...
-		device = await fakeDevice.provisionDevice(admin, applicationId);
-
-		stateMock.getInstance()['updateDeviceModel'] = function (
-			uuid: string,
-			newState: stateMock.DeviceOnlineStates,
-		) {
-			tracker.stateUpdated(uuid, newState);
-			return updateDeviceModel.call(this, uuid, newState);
-		};
+		describe(`API heartbeat state`, () => {
```
pipex, line 78: This suite doesn't call the target state endpoint, we could take it out but it would require some code duplication

test/03_device-state.ts
```diff
@@ -38,296 +38,315 @@ configMock['API_HEARTBEAT_STATE_TIMEOUT_SECONDS'] = TIMEOUT_SEC;
...
 
-				await waitFor({ checkFn: () => statsEventSpy.callCount >= 3 });
+				[
+					{
+						tokenType: 'device API Key',
+						getActor: () => device,
+						heartbeatAfterGet: DeviceOnlineStates.Online,
+						getDevice: () => device,
+						getStateV2: () =>
+							fakeDevice.getState(device, device.uuid, stateVersion as any),
```
pipex, line 200: I could not figure out how to avoid the `stateVersion as any`, without it everything I tried got `"v2" | "v3" is not compatible with "v2"` 😒

test/test-lib/fake-device.ts
```diff
@@ -18,6 +18,11 @@ export async function getState(
 	deviceUuid: string,
 	stateVersion: 'v3',
 ): Promise<StateV3>;
+export async function getState(
```
pipex, line 21: Nice one @thgreasi, I had tried the `as const` but didn't think to add another overload to the function

thgreasi (approved):

PR #688, opened 2021-06-04
Improve/add typings for the device state patch body

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #692, opened 2021-06-08
Remove ui config from the state endpoint

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #695, opened 2021-06-10
Add build statistics to Release

toochevere:
This change is part of the larger change to add build statistics from the Builder to Releases.

Signed-off-by: Paul Jonathan Zoulin <pj@balena.io>
Change-type: minor
See:  https://www.flowdock.com/app/rulemotion/resin-tech/threads/4dsc0_Fec2ZZ3xxnZT3hkG7UMHM

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -87,6 +87,12 @@ Term: api secret
 Term: app name
 	Concept Type: Text (Type)
 
+Term: average cpu usage
+	Concept Type: Integer (Type)
+
+Term: average memory usage
+	Concept Type: Integer (Type)
+
 Term: build log
 	Concept Type: Text (Type)
 
@@ -207,6 +213,12 @@ Term: os version range
 Term: os variant
 	Concept Type: Short Text (Type)
 
+Term: peak cpu usage
+	Concept Type: Integer (Type)
+
+Term: peak memory usage
+	Concept Type: Integer (Type)
+
 Term: project type
 	Concept Type: Short Text (Type)
 
@@ -230,6 +242,9 @@ Term: release version
 
 Term: release type
 	Concept Type: Short Text (Type)
+
+Term: run time
+	Concept Type: Integer (Type)
 
 Term: scope
 	Concept Type: Short Text (Type)
@@ -379,7 +394,6 @@ Fact type: user (Auth) is member of organization
 	Synonymous Form: organization includes user (Auth)
 	Database Table Name: organization membership
 	Term Form: organization membership
-
 
 -- organization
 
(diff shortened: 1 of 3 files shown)


src/migrations/00053-build-stats.sql
```diff
@@ -0,0 +1,14 @@
...
+ADD COLUMN IF NOT EXISTS "average memory usage" INT NULL;
+
+ALTER TABLE "release"
+ADD COLUMN IF NOT EXISTS "peak cpu usage" INT NULL;
+
+ALTER TABLE "release"
+ADD COLUMN IF NOT EXISTS "average cpu usage" INT NULL;
+
+ALTER TABLE "release"
+ADD COLUMN IF NOT EXISTS "run time" INT NULL;
```
thgreasi, line 14: NIT: I think you can do:
```suggestion
ALTER TABLE "release"
ADD COLUMN IF NOT EXISTS "peak memory usage" INT NULL,
ADD COLUMN IF NOT EXISTS "peak cpu usage" INT NULL,
ADD COLUMN IF NOT EXISTS "average memory usage" INT NULL,
ADD COLUMN IF NOT EXISTS "average cpu usage" INT NULL,
ADD COLUMN IF NOT EXISTS "run time" INT NULL;
```

src/balena.sbvr
```diff
@@ -649,6 +663,16 @@ Fact type: release has release type
...
 
+Fact type: release has peak memory usage
+	Necessity: each release has at most one peak memory usage
+Fact type: release has peak cpu usage
+	Necessity: each release has at most one peak cpu usage
+Fact type: release has average memory usage
+	Necessity: each release has at most one average memory usage
+Fact type: release has average cpu usage
+	Necessity: each release has at most one average cpu usage
+Fact type: release has run time
```
thgreasi, line 674: Not sure whether we would care to have a rule to only allow this being set iff it's not still running, but we can probably add that later if we need.

thgreasi (commented):
Other than that LGTM.
Not sure whether Page might have any suggestions that would reduce the amount for Terms.

src/migrations/00053-build-stats.sql
toochevere: done

dfunckt:
It's good to default to improving open-balena-api but these fields are going to be managed by the builder hence I don't see a reason to clutter the core model. I'd consider moving this PR to balena-api.

toochevere:
> It's good to default to improving open-balena-api but these fields are going to be managed by the builder hence I don't see a reason to clutter the core model. I'd consider moving this PR to balena-api.

Will consider in arch meeting.

toochevere:
Closing this baby down and moving to `balena-api`

PR #699, opened 2021-06-30
Avoid unnecessary `getAffectedIds` calls when updating devices

Page-:
We only need the call if the device is moving between applications
which is a tiny minority of the cases in which we patch devices

Change-type: patch


Page-:
@balena-ci I self-certify!

thgreasi:
:tada:

PR #711, opened 2021-07-06
env-vars: Remove check for reserved namespaces

klutchell:
This was leftover legacy behaviour from when config vars
were handled the same way as env vars.

See: https://jel.ly.fish/8b4c23c9-705a-46bd-b446-707aeba29d45
Change-type: patch
Signed-off-by: Kyle Harding <kyle@balena.io>


klutchell:
Closed as per discussion https://jel.ly.fish/8b4c23c9-705a-46bd-b446-707aeba29d45

PR #714, opened 2021-07-13
Update pinejs and use HttpError's for the rate-limiting errors

Page-:
Update pinejs from 14.25.0 to 14.26.1

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #716, opened 2021-07-14
Support generating named provisioning keys

nitishagar:
Change-Type: minor
Signed-off-by: Nitish Agarwal 1592163+nitishagar@users.noreply.github.com

HQ: https://jel.ly.fish/e450713f-001c-4470-b571-71724aeb0869
See: https://www.flowdock.com/app/rulemotion/r-product/threads/DZt_LvQvMCOoN6p2EceSu6MG4MF
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/IPdnkCLH2G0o7DSlkSS9nRetA4f

### DIFF
--- a/src/features/api-keys/routes.ts
+++ b/src/features/api-keys/routes.ts
@@ -60,7 +60,11 @@ export const createProvisioningApiKey: RequestHandler = async (req, res) => {
 	}
 
 	try {
-		const apiKey = await $createProvisioningApiKey(req, appId);
+		const apiKey = await $createProvisioningApiKey(
+			req,
+			appId,
+			req.body.name ? { name: req.body.name } : {},
+		);
 		res.json(apiKey);
 	} catch (err) {
 		if (handleHttpErrors(req, res, err)) {
--- a/src/features/device-config/device-config.ts
+++ b/src/features/device-config/device-config.ts
@@ -10,7 +10,11 @@ import * as deviceConfig from 'balena-device-config';
 import { getUser } from '../../infra/auth/auth';
 import { captureException } from '../../infra/error-handling';
 
-import { createProvisioningApiKey, createUserApiKey } from '../api-keys/lib';
+import {
+	ApiKeyOptions,
+	createProvisioningApiKey,
+	createUserApiKey,
+} from '../api-keys/lib';
 import type { DeviceType } from '../device-types/device-types';
 
 const { BadRequestError } = errors;
@@ -49,7 +53,21 @@ export const generateConfig = async (
 			// Older ones have to use the old "user api keys"
 			return await createUserApiKey(req, (await userPromise).id);
 		}
-		return await createProvisioningApiKey(req, app.id);
+
+		const apiKeyOptions: ApiKeyOptions = {};
+
+		// Checking both req.body and req.query given both GET and POST support
+		// Ref: https://github.com/balena-io/balena-api/blob/master/src/routes/applications.ts#L95
+		const keyName: string | undefined =
+			req.body.provisioningKeyName ??
+			req.query.provisioningKeyName ??
+			undefined;
+
+		if (typeof keyName === 'string') {
+			apiKeyOptions.name = keyName;
+		}
+
+		return await createProvisioningApiKey(req, app.id, apiKeyOptions);
 	})();
 
 	// There may be multiple CAs, this doesn't matter as all will be passed in the config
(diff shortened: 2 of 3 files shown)


src/features/device-config/device-config.ts
```diff
@@ -49,7 +49,13 @@ export const generateConfig = async (
 			// Older ones have to use the old "user api keys"
 			return await createUserApiKey(req, (await userPromise).id);
 		}
-		return await createProvisioningApiKey(req, app.id);
+		return await createProvisioningApiKey(
+			req,
+			app.id,
+			req.body.provisioningKeyName
+				? { name: req.body.provisioningKeyName }
```
Page-, line 56: You should probably also add the ability to name the provisioning api key whilst going via the `/api-key/application/:appId/provisioning` and `/api-key/v1` routes, cc @thgreasi

nitishagar: Makes sense. `/api-key/v1` already has this available via body `{ name: "key-name" }`, but the `/api-key/application/:appId/provisioning` needed that in place.

Added it here.

src/features/api-keys/routes.ts
```diff
@@ -60,7 +60,13 @@ export const createProvisioningApiKey: RequestHandler = async (req, res) => {
 	}
 
 	try {
-		const apiKey = await $createProvisioningApiKey(req, appId);
+		const apiKey = await $createProvisioningApiKey(
+			req,
+			appId,
+			req.body.provisioningKeyName
```
Page-, line 66: Can you make this use `req.body.name` here please so that it can be standardized with the other api creation endpoints

test/08_api-keys.ts
```diff
@@ -20,21 +20,32 @@ describe('create provisioning apikey', function () {
...
+				applicationId: number,
+				provisioningKeyName?: string,
+			) {
 				return supertest(user)
 					.post(`/api-key/v1/`)
 					.send({
 						actorType: 'application',
 						actorTypeId: applicationId,
 						roles: ['provisioning-api-key'],
+						name: provisioningKeyName,
```
thgreasi, line 48: Let's also test that the provided name is actually set in DB.
I expect this to fail atm actually, since in the code above you are using `{ name: req.body.provisioningKeyName }`.

nitishagar: This is already existing endpoint - so will not fail the request. updating to use the `name` value instead of `provisioningKeyName`.

thgreasi: > Let's also test that the provided name is actually set in DB.

I mean similarly to how you did it in the SDK  tests.

src/features/device-config/device-config.ts
```diff
@@ -49,7 +53,20 @@ export const generateConfig = async (
...
+
+		// Checking both req.body and req.query given both GET and POST support
+		// Ref: https://github.com/balena-io/balena-api/blob/master/src/routes/applications.ts#L95
+		if (req.body.name) {
+			apiKeyOptions.name = req.body.name as string;
+		}
+
+		if (req.query.name) {
+			apiKeyOptions.name = req.query.name as string;
+		}
```
thgreasi, line 67:
```suggestion
		const keyName: string | undefined = req.body.provisioningKeyName ?? req.query.provisioningKeyName ?? undefined;
		if (typeof keyName === 'string') {
			apiKeyOptions.name = keyName;
		}
```
I strongly feel that the parameter name should be `provisioningKeyName` instead of just `name`, which better explains what that the value will be used for, and make more sense along all the rest supported parameters.
```
export interface ImgConfigOptions {
	network?: 'ethernet' | 'wifi';
	appUpdatePollInterval?: number;
	wifiKey?: string;
	wifiSsid?: string;
	ip?: string;
	gateway?: string;
	netmask?: string;
	deviceType?: string;
	version: string;
}
```

nitishagar: @Page- Is the `provisioningKeyName` fione with you on the req.body as well? I can change it back to `provisioningKeyName` along with test cases.

thgreasi: @nitishagar to be clear, I only mean using `provisioningKeyName` instead of `name` for the device config endpoint only (where this comment is). For the api key generation endpoints `name` is more appropriate, as Page suggested.

thgreasi:
@nitishagar let's also change the commit message to something like
> Support generating named provisioning keys

src/features/device-config/device-config.ts
Page-: Yes I prefer it to be `provisioningKeyName` because otherwise it's ambiguous, it's just in the case where it's eg `POST /provisioning-api-key` where `name` is unambiguous it's better as it also matches eg `POST /named-user-api-key` for a unified interface when creating api keys

nitishagar:
@thgreasi @Page- Can you take a look, I have updated based on the feedback received?

thgreasi (approved):

PR #719, opened 2021-07-16
Use `res.status(...).end()` in preference to `res.sendStatus()`

Page-:
Update pinejs from 14.26.1 to 14.27.1

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #721, opened 2021-07-19
Update pinejs and make use of pinejs' handleHttpErrors

Page-:
Update pinejs from 14.27.1 to 14.28.0

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #723, opened 2021-07-20
release versioning 2: Migrate records, set both & read only the new fields

thgreasi:
Release plan:
* step 1:
  See: https://github.com/balena-io/open-balena-api/pull/722
  See: https://github.com/balena-io/balena-api/pull/3248
  * add the new field
  * have the API set both the release_type & revision but only read the release_type field
* step 2:
  See: https://github.com/balena-io/open-balena-api/pull/723 **WEW ARE HERE**
  See: https://github.com/balena-io/balena-api/pull/3249
  * run a migration that aligns the new field values based on the old one
  * have the API set both the release_type & revision but only read the revision field
* step 3:
  See: https://github.com/balena-io/open-balena-api/pull/724
  See: https://github.com/balena-io/balena-api/pull/3250
  * stop setting the release_type field & add a translation for it
  * remove the release_type field from the sbvr
* step 4:
  See: https://github.com/balena-io/open-balena-api/pull/725
  See: https://github.com/balena-io/balena-api/pull/3251
  * remove the release_type field from the DB

Change-type: minor
HQ: https://jel.ly.fish/8ea1c390-9a85-402d-978c-4d31dcb0d235


thgreasi:
@balena-ci retest

src/balena-model.ts
```diff
@@ -393,6 +393,13 @@ export interface Release {
 	contract: {} | null;
 	is_passing_tests: boolean;
 	release_type: 'final' | 'draft';
+	is_finalized_at__date: DateString | null;
+	semver_major: number;
```
gelbal, line 397: Seems worthwhile to track these new fields in the `release history` as well. Keeping track of such metadata enables us to run analysis on these dimensions. Furthermore, we might want to expose `release history` to end users and they would appreciate such metadata.

src/migrations/00054-migrate-release-versioning-fields.sql
```diff
@@ -0,0 +1,105 @@
+UPDATE "release"
+SET "revision" = NULL
+WHERE "release type" = 'draft'
```
gelbal, line 3: Do we need a further migration for the draft releases as well to set `revision` to timestamp?

src/features/ci-cd/models/release-additions.ts
```diff
@@ -0,0 +1,159 @@
...
+	});
+
+	const createdAtTimestamp: CastNode = [
+		'Cast',
+		[
+			'Floor',
+			[
+				'Multiply',
+				['Totalseconds', ['ReferencedField', 'release', 'created at']],
+				['Number', 1000],
```
gelbal, line 44: Any particular reason for this timestamp conversion?

gelbal (commented):
Looks great and well thought out Thodori. I left couple of minor suggestions and questions (mainly due to my curiosity).

src/balena.sbvr
```diff
@@ -783,6 +807,7 @@ Rule: It is necessary that each release that has a release version1 and has a st
 Rule: It is necessary that each image that has a status that is equal to "success", has a push timestamp.
 Rule: It is necessary that each application that is public, has an application type that is not legacy.
 Rule: It is necessary that each application that owns a release1 that has a status that is equal to "success" and has a commit1, owns at most one release2 that has a status that is equal to "success" and has a commit2 that is equal to the commit1.
+Rule: It is necessary that each application that owns a release1 that has a revision, owns at most one release2 that has a semver major that is of the release1 and has a semver minor that is of the release1 and has a semver patch that is of the release1 and has a revision that is of the release1.
```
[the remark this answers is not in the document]
thgreasi: Would you prefer it this way v?
```sbvr
It is necessary that each application that owns a release1 that has a **revision1**, owns at most one release2 that has a semver major that is of the release1 and has a semver minor that is of the release1 and has a semver patch that is of the release1 and has a revision2 **that is is equal to the revision1**.
```
The `release1 that has a revision` is there in order to filter out all releases that are draft upfront, but then I decided to not name it and use `that is of release1` for consistency with the rest semver major/minor/patch checks (which also just use `that is of release1` in order to avoid having all parts of release1 and make the rule a bit shorter).

The full rule which would consistently use `equals` instead of `is of` would be huge:
```sbvr
It is necessary that each application that owns a release1 that has a revision1 and has a semver major1 and has a semver minor1 and has a semver patch1, owns at most one release2 that has a semver major2 that is equal to the semver major1 and has a semver minor2 that is equal to the semver minor1 and has a semver patch2 that equal to the semver patch1 and has a revision2 that is is equal to the revision1.
```

src/migrations/00054-migrate-release-versioning-fields.sql
```diff
@@ -0,0 +1,105 @@
...
+		parsedVersionRelease."belongs to-application",
+		-- Had to be a separate SELECT b/c we can't COALESCE REGEXP_MATCHES in the same SELECT.
+		COALESCE(parsedVersionRelease."semver major", 0) AS "semver major",
+		COALESCE(parsedVersionRelease."semver minor", 0) AS "semver minor",
+		COALESCE(parsedVersionRelease."semver patch", 0) AS "semver patch"
+	FROM (
+		SELECT r0."id",
+			r0."is finalized at-date",
+			r0."belongs to-application",
+			CAST((REGEXP_MATCHES (r0."release version", '^v?([0-9]+)(\.[0-9]+(\.[0-9]+)?)?'))[1] AS INTEGER) AS "semver major",
```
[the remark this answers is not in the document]
thgreasi: Didn't think about that!
But this should also need some extra code to remove the possibly `v` prefix and `_logstream` suffix in the major & patch cases, right?

src/migrations/00054-migrate-release-versioning-fields.sql
```diff
@@ -0,0 +1,105 @@
...
+BEGIN
+	IF EXISTS (
+		SELECT 1
+		FROM "application" AS "application.0",
+			"release" AS "release.1"
+		WHERE "release.1"."revision" IS NOT NULL
+		AND "release.1"."belongs to-application" = "application.0"."id"
+		AND (
+			SELECT COUNT(*)
+			FROM "release" AS "release.3"
```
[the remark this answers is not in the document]
thgreasi: That's actually a copy paste of the SQL that pine emits from the rule we added, so that's why I preferred to have it here "as is".

src/balena-model.ts
thgreasi: That would be in the cloud API PR @gelbal :)
You mean all the semver & the revision, or do you also care about the is_finalized_at__date ?

src/features/ci-cd/models/release-additions.ts
thgreasi: The DB's `EXTRACT(EPOCH FROM A_TIMESTAMP_HERE)` gives something like `1,627,388,902.838732` which uses a different accuracy & format than the nodejs Date's numeric representation which gives something like `1,627,389,065,266`.
The requirement for the version field of draft releases was to have that `-${timestamp}` prerelease part. Now the floor & * 1000 is strictly speaking not required, but thought it would be better if that was consistent with the js numeric representation of the `created_at` field.

src/migrations/00054-migrate-release-versioning-fields.sql
thgreasi: No, draft releases will have a NULL `revision`, which actually will be the source of truth on whether a release is draft or not.
The timestamp part will only be in the definition of the `raw` part of the `version` computed term.

I hope I didn't misunderstood your question though. So, does that make sense?

src/migrations/00054-migrate-release-versioning-fields.sql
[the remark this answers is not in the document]
gelbal: Indeed, we need to strip such prefix and suffix first. Then the current regex ignores the suffix already, right?

src/migrations/00054-migrate-release-versioning-fields.sql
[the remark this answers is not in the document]
gelbal: Aha. Interesting to see pine emitting such inconsistencies.

src/balena-model.ts
gelbal: Yeah, yeah for sure in the cloud API PR. Just wanted to share my thoughts as they appeared. Yes, I was thinking of tracking all the semver & the revision fields. Probably no need for `is_finalized_at__date` as we could derive it from the historical records. Besides, right now I cannot think of analytics reasons to keep track of that.

src/migrations/00054-migrate-release-versioning-fields.sql
gelbal: > The timestamp part will only be in the definition of the `raw` part of the `version` computed term.

This answered it for me Thodori :+1:

src/features/ci-cd/models/release-additions.ts
gelbal: Aha, good to know about the mismatch in timestamp nature between DB and JS. TIL.

src/migrations/00054-migrate-release-versioning-fields.sql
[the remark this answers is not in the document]
thgreasi: Updated to use STRING_TO_ARRAY & SUBSTRING to extract the numeric parts only.
Dropped the runtime a bit (I expected more tbh):
```
(cost=160922.60..160935.13 rows=5010 width=61) (actual time=181.354..182.159 rows=9487 loops=1)
(cost=160847.46..160859.99 rows=5010 width=61) (actual time=156.321..157.081 rows=9487 loops=1)
```
More importantly, it allowed me to squash the subquery with the outer and use COALESCE right in place.

Nice hint :+1:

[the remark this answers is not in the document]
thgreasi: > Indeed, we need to strip such prefix and suffix first. Then the current regex ignores the suffix already, right?

Yes the regex was only matching the numbers w/o caring about the suffix.

src/balena.sbvr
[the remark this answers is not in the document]
gelbal: Thanks for the explanation Thodori. Your new version reads a bit better. Honestly, if I haven't seen the SQL / TS code and your comments, I probably wouldn't understand this rule.

> The ideal would be if pine supported using computed terms in rules

Right. Indeed, such feature would help here.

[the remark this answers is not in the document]
thgreasi: TBH I prefer the consistency rather than mixing `is of` and `is equal to` :thinking: Let's see whether Page has any preference on this.

src/features/ci-cd/models/release-additions.ts
```diff
@@ -1,28 +1,159 @@
...
+						'Else',
+						[
+							'Concatenate',
+							['EmbeddedText', '+rev'],
+							['ReferencedField', 'release', 'revision'],
+						],
+					],
+				],
+				'Text',
+			],
```
Page-, line 88: I'm not sure how beneficial it really is but you could use:
```ts
function sqlConcat(
	[start, ...strings]: TemplateStringsArray,
	...nodes: Array<TextTypeNodes | string>
) {
	const concats: ConcatenateNode = ['Concatenate'];
	const addNode = (node: typeof nodes[number]) => {
		if (typeof node === 'string') {
			if (node.length > 0) {
				concats.push(['EmbeddedText', node]);
			}
		} else {
			concats.push(node);
		}
	};
	addNode(start);
	for (let i = 0; i < strings.length; i++) {
		addNode(nodes[i]);
		addNode(strings[i]);
	}

	return concats;
}
```
and do things like
```ts
sqlConcat`${['ReferencedField', 'release', 'semver major']}.${['ReferencedField', 'release', 'semver minor']}.${['ReferencedField', 'release', 'semver patch']}`
```
to get
```js
[
  'Concatenate',
  [ 'ReferencedField', 'release', 'semver major' ],
  [ 'EmbeddedText', '.' ],
  [ 'ReferencedField', 'release', 'semver minor' ],
  [ 'EmbeddedText', '.' ],
  [ 'ReferencedField', 'release', 'semver patch' ]
]

File: src/features/ci-cd/hooks/release-versioning.ts [hunk 8]
```diff
@@ -181,106 +180,100 @@ interface PatchCustomObject extends CustomObjectBase {
...
+		},
+	});
+};
+
+hooks.addPureHook('PATCH', 'resin', 'release', {
+	POSTPARSE: parseReleaseVersioningFields,
+	PRERUN: async (args) => {
+		await Promise.all([
+			setReleasesToSetRevision(args),
+			preventChangingFinalToDraft(args),
```
Page-, line 275: Do both of these rely on `parseReleaseVersioningFields`?

thgreasi: `preventChangingFinalToDraft` depends on `parseReleaseVersioningFields` setting the `custom.is_final`.
In order to make `preventChangingFinalToDraft` to not rely on `parseReleaseVersioningFields`, we would have to move it to POSTPARSE and read the `values.is_final` directly, but then that would rely on the order that the hooks run, since `parseReleaseVersioningFields` also deletes the `values.is_final` since it doesn't exist in the DB.

src/migrations/00054-migrate-release-versioning-fields.sql
```diff
@@ -0,0 +1,96 @@
...
+				WHERE pvr."id" = unparsedRelease."id"
+			)
+	) AS splitedReleaseVersion
+)
+UPDATE "release" r  SET
+	"semver major" = rr."semver major",
+	"semver minor" = rr."semver minor",
+	"semver patch" = rr."semver patch",
+	"revision" = rr."revision"
+FROM "revisioned release" rr
```
thgreasi, line 68: FYI this returns about 10k records and the select takes about 1.7 seconds.

src/migrations/00054-migrate-release-versioning-fields.sql
```diff
@@ -0,0 +1,96 @@
...
+		FROM "release" unparsedRelease
+		WHERE unparsedRelease."release type" = 'final'
+			AND NOT EXISTS (
+				SELECT 1
+				FROM "parsed version release" pvr
+				WHERE pvr."id" = unparsedRelease."id"
+			)
+	) AS splitedReleaseVersion
+)
+UPDATE "release" r  SET
```
thgreasi, line 63: FYI that the SELECT of this takes about 4.5 seconds on chartio, updating almost all 1,7m releases.
> Hash Join (cost=3334918.49..3428992.14 rows=1710430 width=61) (actual time=3078.994..4267.474 rows=1007355 loops=1)

src/features/ci-cd/models/release-additions.ts
thgreasi: @Page- I absolutely loved how the result looks like :heart: 
That's one of the JS features that I often forget.

src/balena.sbvr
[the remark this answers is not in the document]
thgreasi: With the index already in place, this takes just 30ms :)

src/migrations/00054-migrate-release-versioning-fields.sql
gelbal: Hey @thgreasi it's not immediately clear to me where this `Hash Join` happens. Can you please post the full query plan? 
(no promise if I would have a better idea though :sweat_smile:)

On a related note: What's the expected performance here? 
Sounds like 4.5 seconds is too long. But I imagine that it might go faster on the production db because of different setup.

thgreasi: @gelbal that was from a SELECT query I was doing on chartio, it should be from the join between the `r` and the `rr`.
Here is the full plan for the UPDATE as I got it from my fasttest DB:
```
Update on release r  (cost=20.77..31.07 rows=2 width=2810) (actual time=0.003..0.004 rows=0 loops=1)
  CTE parsed version release
    ->  Seq Scan on release r0  (cost=0.00..10.33 rows=1 width=28) (never executed)
          Filter: ((("release version")::text ~ '^(?:v?[0-9]+(?:\.[0-9]+(?:\.[0-9]+(?:\.[0-9]+)?(?:.logstream[0-9]?)?)?)?)$'::text) AND (("release type")::text = 'final'::text))
  ->  Hash Join  (cost=10.44..20.74 rows=2 width=2810) (actual time=0.002..0.003 rows=0 loops=1)
        Hash Cond: (r.id = rr.id)
        ->  Seq Scan on release r  (cost=0.00..10.20 rows=20 width=2746) (actual time=0.002..0.002 rows=0 loops=1)
        ->  Hash  (cost=10.42..10.42 rows=2 width=72) (never executed)
              ->  Subquery Scan on rr  (cost=10.33..10.42 rows=2 width=72) (never executed)
                    ->  WindowAgg  (cost=10.33..10.40 rows=2 width=36) (never executed)
                          ->  Sort  (cost=10.33..10.34 rows=2 width=28) (never executed)
                                Sort Key: pvr."belongs to-application", pvr."semver major", pvr."semver minor", pvr."semver patch", pvr."is finalized at-date", pvr.id
                                ->  Append  (cost=0.00..10.32 rows=2 width=28) (never executed)
                                      ->  CTE Scan on "parsed version release" pvr  (cost=0.00..0.02 rows=1 width=28) (never executed)
                                      ->  Subquery Scan on "*SELECT* 2"  (cost=0.00..10.29 rows=1 width=28) (never executed)
                                            ->  Nested Loop Anti Join  (cost=0.00..10.28 rows=1 width=28) (never executed)
                                                  Join Filter: (pvr_1.id = unparsedrelease.id)
                                                  ->  Seq Scan on release unparsedrelease  (cost=0.00..10.25 rows=1 width=16) (never executed)
                                                        Filter: (("release type")::text = 'final'::text)
                                                  ->  CTE Scan on "parsed version release" pvr_1  (cost=0.00..0.02 rows=1 width=4) (never executed)
Planning Time: 1.921 ms
Execution Time: 0.285 ms
```

and here is the respective for a query that just changes the UPDATE with a SELECT from chartio
```sql
-- everything up to here was the same, except that I removed "is finalized at-date" b/c it isn't in the external schema
SELECT *
FROM "release" r,
	"revisioned release" rr
WHERE r."id" = rr."id";
```
```
QUERY PLAN
Merge Join  (cost=2646613.98..3817498.92 rows=1835109 width=1538) (actual time=3254.764..5057.455 rows=1007716 loops=1)
  Merge Cond: (release.id = rr.id)
  CTE parsed version release
    ->  Gather  (cost=1000.00..1059884.90 rows=16430 width=20) (actual time=179.211..291.430 rows=10320 loops=1)
          Workers Planned: 4
          Workers Launched: 4
          ->  Parallel Seq Scan on release release_2  (cost=0.00..1057241.90 rows=4108 width=20) (actual time=176.554..291.249 rows=2064 loops=5)
                Filter: ((("release version")::text ~ '^(?:v?[0-9]+(?:\.[0-9]+(?:\.[0-9]+(?:\.[0-9]+)?(?:.logstream[0-9]?)?)?)?)$'::text) AND (("release type")::text = 'final'::text))
                Rows Removed by Filter: 199479
  ->  Index Scan using release_pkey on release  (cost=0.43..1092893.24 rows=1835109 width=1450) (actual time=0.017..970.756 rows=1007716 loops=1)
  ->  Sort  (cost=1586728.66..1591316.43 rows=1835109 width=24) (actual time=3254.736..3395.321 rows=1007716 loops=1)
        Sort Key: rr.id
        Sort Method: quicksort  Memory: 103304kB
        ->  Subquery Scan on rr  (cost=1322404.75..1395809.11 rows=1835109 width=24) (actual time=2262.859..2994.351 rows=1007716 loops=1)
              ->  WindowAgg  (cost=1322404.75..1377458.02 rows=1835109 width=28) (actual time=2262.858..2885.288 rows=1007716 loops=1)
                    ->  Sort  (cost=1322404.75..1326992.52 rows=1835109 width=20) (actual time=2262.845..2411.443 rows=1007716 loops=1)
                          Sort Key: pvr."belongs to-application", pvr."semver major", pvr."semver minor", pvr."semver patch", pvr.id
                          Sort Method: quicksort  Memory: 103304kB
                          ->  Append  (cost=0.00..1131485.20 rows=1835109 width=20) (actual time=179.217..1555.836 rows=1007716 loops=1)
                                ->  CTE Scan on "parsed version release" pvr  (cost=0.00..328.60 rows=16430 width=20) (actual time=179.216..293.945 rows=10320 loops=1)
                                ->  Subquery Scan on "*SELECT* 2"  (cost=1479.46..1121981.05 rows=1818679 width=20) (actual time=3.368..1181.615 rows=997396 loops=1)
                                      ->  Merge Anti Join  (cost=1479.46..1103794.26 rows=1818679 width=20) (actual time=3.367..1075.599 rows=997396 loops=1)
                                            Merge Cond: (release_1.id = pvr_1.id)
                                            ->  Index Scan using release_pkey on release release_1  (cost=0.43..1097481.01 rows=1835109 width=8) (actual time=0.020..921.823 rows=1007716 loops=1)
                                                  Filter: (("release type")::text = 'final'::text)
                                            ->  Sort  (cost=1479.03..1520.11 rows=16430 width=4) (actual time=3.343..4.986 rows=10320 loops=1)
                                                  Sort Key: pvr_1.id
                                                  Sort Method: quicksort  Memory: 868kB
                                                  ->  CTE Scan on "parsed version release" pvr_1  (cost=0.00..328.60 rows=16430 width=4) (actual time=0.001..1.196 rows=10320 loops=1)
Planning Time: 1.339 ms
Execution Time: 5117.694 ms
```

thgreasi: @gelbal  After some experiments with Page, we dropped down to 2.3 seconds!!!
```
SET JIT = 0; EXPLAIN ANALYSE
WITH "parsed version release" AS (
	SELECT r0."id",
		r0."belongs to-application", 
		CASE
			WHEN r0."release version" SIMILAR TO 'v?[0-9]+(.[0-9]+(.[0-9]+(.[0-9]+)?(_logstream[0-9]?)?)?)?' THEN
				COALESCE(CAST(SUBSTRING((STRING_TO_ARRAY(r0."release version", '.'))[1] FROM '[0-9]+') AS INTEGER), 0)
			ELSE 0
		END AS "new semver major",
		CASE
			WHEN r0."release version" SIMILAR TO 'v?[0-9]+(.[0-9]+(.[0-9]+(.[0-9]+)?(_logstream[0-9]?)?)?)?' THEN
				COALESCE(CAST(SUBSTRING((STRING_TO_ARRAY(r0."release version", '.'))[2] FROM '[0-9]+') AS INTEGER), 0)
			ELSE 0
		END AS "new semver minor",
		CASE
			WHEN r0."release version" SIMILAR TO 'v?[0-9]+(.[0-9]+(.[0-9]+(.[0-9]+)?(_logstream[0-9]?)?)?)?' THEN
				COALESCE(CAST(SUBSTRING((STRING_TO_ARRAY(r0."release version", '.'))[3] FROM '[0-9]+') AS INTEGER), 0)
			ELSE 0
		END AS "new semver patch" 
	FROM "release" r0
	WHERE r0."release type" = 'final'
	ORDER BY "belongs to-application",
		"new semver major",
		"new semver minor",
		"new semver patch",
		"id"
),
"revisioned release" AS (
	SELECT split."id",
		split."new semver major" AS "semver major",
		split."new semver minor" AS "semver minor",  
		split."new semver patch" AS "semver patch",
		(ROW_NUMBER() OVER (
			PARTITION BY split."belongs to-application", 
				split."new semver major",
				split."new semver minor",
				split."new semver patch"
			ORDER BY
				split."id" ASC
		) - 1) AS "revision"
	FROM "parsed version release" AS split
)
SELECT r."belongs to-application", rr.*
FROM "release" r,
	"revisioned release" rr
WHERE r."id" = rr."id"
```
```
QUERY PLAN
Hash Join  (cost=1196744.97..1508455.77 rows=1835108 width=28) (actual time=859.507..2243.405 rows=1007837 loops=1)
  Hash Cond: (split.id = release.id)
  ->  WindowAgg  (cost=1117318.00..1405860.55 rows=1835108 width=28) (actual time=457.167..1405.150 rows=1007837 loops=1)
        ->  Subquery Scan on split  (cost=1117318.00..1355395.08 rows=1835108 width=20) (actual time=457.152..921.196 rows=1007837 loops=1)
              ->  Gather Merge  (cost=1117318.00..1337044.00 rows=1835108 width=20) (actual time=457.149..809.078 rows=1007837 loops=1)
                    Workers Planned: 4
                    Workers Launched: 4
                    ->  Sort  (cost=1116317.94..1117464.89 rows=458777 width=20) (actual time=453.423..487.303 rows=201567 loops=5)
"                          Sort Key: release_1.""belongs to-application"", (CASE WHEN ((release_1.""release version"")::text ~ '^(?:v?[0-9]+(?:\.[0-9]+(?:\.[0-9]+(?:\.[0-9]+)?(?:.logstream[0-9]?)?)?)?)$'::text) THEN COALESCE((""substring""((string_to_array((release_1.""release version"")::text, '.'::text))[1], '[0-9]+'::text))::integer, 0) ELSE 0 END), (CASE WHEN ((release_1.""release version"")::text ~ '^(?:v?[0-9]+(?:\.[0-9]+(?:\.[0-9]+(?:\.[0-9]+)?(?:.logstream[0-9]?)?)?)?)$'::text) THEN COALESCE((""substring""((string_to_array((release_1.""release version"")::text, '.'::text))[2], '[0-9]+'::text))::integer, 0) ELSE 0 END), (CASE WHEN ((release_1.""release version"")::text ~ '^(?:v?[0-9]+(?:\.[0-9]+(?:\.[0-9]+(?:\.[0-9]+)?(?:.logstream[0-9]?)?)?)?)$'::text) THEN COALESCE((""substring""((string_to_array((release_1.""release version"")::text, '.'::text))[3], '[0-9]+'::text))::integer, 0) ELSE 0 END), release_1.id"
                          Sort Method: quicksort  Memory: 21911kB
                          Worker 0:  Sort Method: quicksort  Memory: 21996kB
                          Worker 1:  Sort Method: quicksort  Memory: 21962kB
                          Worker 2:  Sort Method: quicksort  Memory: 21795kB
                          Worker 3:  Sort Method: quicksort  Memory: 21796kB
                          ->  Parallel Seq Scan on release release_1  (cost=0.00..1073175.85 rows=458777 width=20) (actual time=0.008..336.905 rows=201567 loops=5)
"                                Filter: ((""release type"")::text = 'final'::text)"
  ->  Hash  (cost=56488.11..56488.11 rows=1835109 width=8) (actual time=398.081..398.081 rows=1007837 loops=1)
        Buckets: 2097152  Batches: 1  Memory Usage: 55753kB
        ->  Index Only Scan using release_id_belongs_to_app_idx on release  (cost=0.43..56488.11 rows=1835109 width=8) (actual time=0.014..168.863 rows=1007837 loops=1)
              Heap Fetches: 8007
Planning Time: 0.802 ms
Execution Time: 2292.437 ms
```

Page- (approved):

src/migrations/00054-migrate-release-versioning-fields.sql
gelbal: Good to hear @thgreasi. I couldn't find the time to take a look before now. Interesting to see that it's faster to repeat that `SIMILAR TO` statements instead of putting it under `WHERE`.

PR #728, opened 2021-07-30
v3 state patch endpoint

Page-:


pipex:
@balena-ci rebase

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,365 @@
...
+		/**
+		 * Used for setting gateway device of dependent devices
+		 */
+		parent_device?: number;
+		apps?: {
+			[uuid: string]: {
+				/**
+				 * We report the overall release uuid the supervisor considers the active one, even though there may be info for multiple releases.
+				 */
+				release_uuid: string;
```
pipex, line 86: I think this probably should be optional @Page-, as the supervisor reports the current release uuid only after it has successfully installed all services, which means that for the first release install, the `release_uuid` will be empty until installation finishes.

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,365 @@
...
+				releases?: {
+					[releaseUUID: string]: {
+						services?: {
+							[name: string]: {
+								image: string;
+								status: string;
+								download_progress: number;
```
pipex, line 93: This probably should be nullable as well.

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,365 @@
...
+			[uuid: string]: {
+				/**
+				 * We report the overall release uuid the supervisor considers the active one, even though there may be info for multiple releases.
+				 */
+				release_uuid: string;
+				releases?: {
+					[releaseUUID: string]: {
+						services?: {
+							[name: string]: {
+								image: string;
```
pipex, line 91: This image is meant to be the image name right? I'm not sure how reliably we can get this value, as the image could get deleted/untagged even if the container is still running. Can you get the image install just from the release uuid and serviceName?

src/features/device-state/routes/state-patch-v3.ts
Page-: Ah, it probably is optional in reality but the typings don't reflect that - I'll fix it either way

src/features/device-state/routes/state-patch-v3.ts
pipex: On second thought, this might be fine, since the supervisor can get the image name from the database, if the image is not on the database (e.g. the DB got deleted), the supervisor will recreate the entry, meaning it generally should be able to get the image name for the running service

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,365 @@
...
+								status: string;
+								download_progress: number;
+							};
```
pipex, line 94: I think the type is missing the `createdAt` property for the service

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,365 @@
...
+	'is_online',
+] as const;
+
+/**
+ * These typings should be used as a guide to what should be sent, but cannot be trusted as what actually *is* sent.
+ */
+export type StatePatchV3Body = {
+	[uuid: string]: {
+		name?: string;
+		status?: string;
```
pipex, line 53: What is this status?

pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

src/features/device-state/routes/state-patch-v3.ts
Page-: I've made it optional

src/features/device-state/routes/state-patch-v3.ts
Page-: This is the overall device status and can contain things like idle/downloading/ordered/preparing/shipping/shipped(HoD)

pipex: this doesn't exist right now on the supervisor, is it used by the API anywhere?

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,375 @@
...
+		/**
+		 * Used for setting gateway device of dependent devices
+		 */
+		parent_device?: number;
+		apps?: {
+			[uuid: string]: {
+				/**
+				 * We report the overall release uuid the supervisor considers the active one, even though there may be info for multiple releases.
+				 */
+				release_uuid?: string;
```
pipex, line 87: I don't think `release_uuid` can be empty. I tried to

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,375 @@
...
+				}
+			}
+
+			for (const [appUuid, releaseUuids] of Object.entries(appReleaseUuids)) {
+				appReleases[appUuid] = (await resinApiTx.get({
+					resource: 'release',
+					options: {
+						$select: ['id', 'commit'],
+						$filter: {
+							commit: { $in: Array.from(releaseUuids) },
```
Page-, line 211: TODO: This needs to make sure `length > 0`

pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

pipex:
Should we take this out of draft @Page-? We have tested supervisor changes along this PR and we haven't found any more issues.

pipex:
@dfunckt @thgreasi would it be possible to get your feedback on this PR when you get a chance? Is enabling the new target state endpoint for tri-app. Thank you!

pipex:
@balena-ci rebase

pipex:
@dfunckt @thgreasi light reminder to please review this PR which is blocking the tri-app and host overlays work 🙏🏻

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,387 @@
...
+						downloadProgress?: number;
+					}> = [];
+					for (const [
+						appUuid,
+						{
+							// release_uuid: isRunningReleaseUuid,
+							releases = {},
+						},
+					] of Object.entries(apps)) {
+						// // TODO: This gets the release we are running for the given app but currently we handle the user app out of band above, and ignore supervisor/os
```
Page-, line 299: Uhhh, better deal with this

pipex:
@balena-ci retest

Page-:
@balena-ci retest

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+		/**
+		 * Used for setting gateway device of dependent devices
+		 */
+		is_managed_by__device?: number;
+		apps?: {
+			[uuid: string]: {
+				/**
+				 * We report the overall release uuid the supervisor considers the active one, even though there may be info for multiple releases.
+				 */
+				release_uuid?: string;
```
dfunckt, line 65:
```suggestion
				current_release_uuid?: string;
```

?

Or:

```suggestion
				active_release_uuid?: string;
```

?

src/features/device-state/routes/state-v3.ts
```diff
@@ -0,0 +1,412 @@
...
+
+const { UnauthorizedError } = errors;
+const { api } = sbvrUtils;
+
+type LocalStateApp = StateV3[string]['apps'][string];
+type ServiceComposition = AnyObject;
+export type StateV3 = {
+	[uuid: string]: {
+		name: string;
+		is_managed_by__device?: string;
```
dfunckt, line 32:
```suggestion
		is_managed_by_device?: string;
```

Is there a reason some fields correspond directly to the model and others are renamed specifically for the target state?

src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -216,7 +217,7 @@ type StatePatchBody = {
 		is_on__commit?: string | null;
 		apps?: Array<{
 			services?: {
-				[serviceId: string]: {
+				[imageId: string]: {
```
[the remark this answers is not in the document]
pipex: > Is this correct? Does the Supervisor really send image IDs as keys in a map called services?

Yes, that is correct 

https://github.com/balena-os/balena-supervisor/blob/master/src/compose/application-manager.ts#L844-L850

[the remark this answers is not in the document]
pipex: I'm not sure about the breaking part, but this looks like just a rename on the type?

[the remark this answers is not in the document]
pipex: > With "breaking" I meant that the expectation (based on the type) seemed to be this was a service ID but apparently with this change we're now treating it as image ID

I see, since the type was not exported before and it looks like the only place that key was used (as imageId) was [here](https://github.com/balena-io/open-balena-api/blob/8de3a73e17492d10120d0b20d277d74020328d1e/src/features/device-state/routes/state-patch.ts#L427) this might not be breaking at all

> However -- just curious -- how did we end up sending image IDs from the Supervisor in map called services

Not sure about the logic of reporting the service like that. That mapping seems to be there since the module was first written in the supervisor https://github.com/balena-os/balena-supervisor/commit/cb37f7ebcc20125c6aa13f954ba38c8b91c40eda#diff-47198ccd5f456a10e7e3b9f75194ac1d47fa5e450a122cd2e56df3620bd705f8R222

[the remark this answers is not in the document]
dfunckt: Right, just the type was wrong all along then. Thanks for looking all this up!

src/features/device-state/routes/state-patch-v3.ts
Page-: The idea is for the current and target states to use a matching format and so things like `current`/`active` don't make sense, there's just a state that happens to be what the device is currently running and one that is what it should be running, and the supervisor just does what is necessary to transform one to the other. In practice there are minor differences where eg the target state doesn't specify things like `is_undervolted` because it doesn't really make sense to do so (it's not like the supervisor can control it), and the current state doesn't report back things like env vars/config because the api already knows it and it'd be a waste of bandwidth (but again it's just an optimization and it could report it and the api would just ignore it)

src/features/device-state/routes/state-v3.ts
Page-: Which ones don't correspond? The only one that comes to mind is `release_uuid`/`commit` and that's because `commit` in the model should become `release_uuid` to better represent what it actually is but that work just hasn't been done yet

src/features/device-state/routes/state-patch-v2.ts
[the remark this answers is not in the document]
Page-: Yep, I changed it because when I was looking at the supervisor I realized that it was indeed using the imageId and I updated the informational type name to match what it actually represents (it's not the first case I've had to do similar updates to match reality since the code predates the usage of typescript and so the types have to match to the unintended reality that emerged)

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+	try {
+		const body = req.body as StatePatchV3Body;
+		const custom: AnyObject = {}; // shove custom values here to make them available to the hooks
+
+		// forward the public ip address if the request is from the supervisor.
+		if (req.apiKey != null) {
+			custom.ipAddress = getIP(req);
+		}
+
+		const uuids = Object.keys(body).filter((uuid) => body[uuid] != null);
```
thgreasi, line 92: Is it valid if the supervisor sends a null-ish value for a device uuid?
I mean, should we instead throw in case `typeof body[uuid] !== 'object'`, or does that have a different meaning?

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+					if (isRunningReleaseUuid) {
+						appReleaseUuids[appUuid].add(isRunningReleaseUuid);
+					}
+					if (releases != null) {
+						for (const [releaseUuid, { services }] of Object.entries(
+							releases,
+						)) {
+							appReleaseUuids[appUuid].add(releaseUuid);
+							if (services != null) {
+								for (const [, service] of Object.entries(services)) {
```
thgreasi, line 118: I guess it could be:
```suggestion
								for (const service of Object.values(services)) {
```

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+							releases,
+						)) {
+							const release = appReleases[appUuid].find(
+								(r) => r.commit === releaseUuid,
+							);
+							if (release == null) {
+								throw new InternalRequestError();
+							}
+							for (const service of Object.values(services)) {
+								const image = images.find(
```
thgreasi, line 313: I guess we could have made the images be a dictionary, but of course no big deal...

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+									(i) =>
+										i.is_stored_at__image_location ===
+										service.image.split('@', 1)[0],
```
thgreasi, line 316: NIT: The split could be outside of the `.find()` (but not a big deal).

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+				}
+			}
+		});
+
+		await sbvrUtils.db.transaction(async (tx) => {
+			const resinApiTx = api.resin.clone({
+				passthrough: { req, custom, tx },
+			});
+
+			const waitPromises: Array<PromiseLike<any>> = [];
```
thgreasi, line 219: Would it be of a benefit if this was 
```suggestion
			const updateFns: Array<(resinApiTx: PinejsClient) => Promise<void>> = [];
```
like in in v2?

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+
+			for (const uuid of uuids) {
+				const state = body[uuid];
+
+				const { apps } = state;
+
+				let deviceBody:
+					| undefined
```
thgreasi, line 227: I think `_.pick` never returns `undefined`
```suggestion
```

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+		});
+
+		await sbvrUtils.db.transaction(async (tx) => {
+			const resinApiTx = api.resin.clone({
+				passthrough: { req, custom, tx },
+			});
+
+			const waitPromises: Array<PromiseLike<any>> = [];
+
+			for (const uuid of uuids) {
```
thgreasi, line 221: Given the few `await`s in the loop, would it make sense to `await Promise.all(uuid.map(....))` this?
PS: On the other hand I get that, given the current usage of dependent devices, this would complicate the code w/o any real perf benefit.

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+					| undefined
+					| (Pick<
+							StatePatchV3Body[string],
+							typeof v3ValidPatchFields[number]
+					  > & {
+							device_name?: string;
+							is_running__release?: number | null;
+					  }) = _.pick(state, v3ValidPatchFields);
+				let metricsBody:
+					| undefined
```
thgreasi, line 236: > I think `_.pick` never returns `undefined`
```suggestion
```

src/features/device-state/routes/state-patch-v3.ts
thgreasi, line 221: Do we need to respect the order in which the device UUIDs were send?
If not, how about iterating over the retrieved devices directly (since we already checked that the lengths do match), so that we don't have to use `const device = devices.find((d) => d.uuid === uuid);` later ?
```suggestion
			for (const device of devices) {
```
Even if the order does matter, could a `$orderby: { depends_on__device: 'asc' }` work around it?

thgreasi (commented):
Will continue tomorrow...

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+				if (apps != null) {
+					const release = appReleases[userAppUuid].find(
+						(r) => r.commit === apps[userAppUuid].release_uuid,
+					);
+					if (release) {
+						deviceBody.is_running__release = release.id;
+					}
+				}
+
+				if (!_.isEmpty(deviceBody)) {
```
thgreasi, line 267: Assuming I'm not wrong that `deviceBody` can't be `undefined`, I guess we can use the native method here, similarly to how we did before:
```suggestion
				if (Object.keys(deviceBody).length > 0) {
```

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+				const device = devices.find((d) => d.uuid === uuid);
+				if (device == null) {
+					throw new UnauthorizedError();
+				}
+				const userAppUuid = device.belongs_to__application[0].uuid;
+				if (apps != null) {
+					const release = appReleases[userAppUuid].find(
+						(r) => r.commit === apps[userAppUuid].release_uuid,
+					);
+					if (release) {
```
thgreasi, line 262: What if `apps[userAppUuid].release_uuid` has a value, but we don't find any such `release` in the DB?
Is it intended to just skip it? The v2 would have thrown I guess b/c of the FK, right :thinking:

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+				}
+
+				const device = devices.find((d) => d.uuid === uuid);
+				if (device == null) {
+					throw new UnauthorizedError();
+				}
+				const userAppUuid = device.belongs_to__application[0].uuid;
+				if (apps != null) {
+					const release = appReleases[userAppUuid].find(
+						(r) => r.commit === apps[userAppUuid].release_uuid,
```
thgreasi, line 260: What if the device was moved to a different app just before the PATCH arrived?
In that case `apps[userAppUuid]` would have been `undefined`, so I guess we can move this before the `if (apps != null) {`. Not sure whether the code should do in such a case.

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,384 @@
...
+			for (const [appUuid, releaseUuids] of Object.entries(appReleaseUuids)) {
+				appReleases[appUuid] =
+					releaseUuids.size === 0
+						? []
+						: ((await resinApiTx.get({
+								resource: 'release',
+								options: {
+									$select: ['id', 'commit'],
+									$filter: {
+										commit: { $in: Array.from(releaseUuids) },
```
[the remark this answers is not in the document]
Page-: I think so, is there any way for a device to be running a release that then gets marked as something other than successful?

src/features/device-state/routes/state-patch-v3.ts
Page-: Right, yeah I should probably update this to match those v2 changes

src/features/device-state/routes/state-patch-v3.ts
Page-: The SQL queries against the actual db will also be run in serial btw, it would only cause the processing around those queries to be concurrent

Page-: I don't think the order matters, and regardless I'd only actually expect a single one to be sent right now anyway, so I'll change this

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -0,0 +1,404 @@
...
+
+const { UnauthorizedError } = errors;
+const { api } = sbvrUtils;
+
+type LocalStateApp = StateV3[string]['apps'][string];
+type ServiceComposition = AnyObject;
+export type StateV3 = {
+	[uuid: string]: {
+		name: string;
+		is_managed_by__device?: string;
```
pipex, line 32: This field has a different name in the GET and PATCH PRs. On the GET PR it is named `parent_device` (which is what the supervisor is expecting). Is this PR just missing a rebase or you intend to change the name?

https://github.com/balena-io/open-balena-api/pull/687/files#diff-b645151f12e0231029c88685140b26ef52329c83bcabe726c1eced933329df52R32

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -0,0 +1,387 @@
...
+		cpu_id?: string;
+		is_undervolted?: boolean;
+		/**
+		 * Used for setting dependent devices as online
+		 */
+		is_online?: boolean;
+		/**
+		 * Used for setting gateway device of dependent devices
+		 */
+		is_managed_by__device?: number;
```
pipex, line 62: Same thing for this field, the supervisor PR is using `parent_device` for this field (although nothing is really writing this right now).

pipex, line 62: Same thing for this field, the supervisor PR is using `parent_device` for this field (although nothing is really writing this right now).

src/features/device-state/routes/state-patch-v3.ts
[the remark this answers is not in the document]
thgreasi: Users can patch the status, but I wasn't worried about that.
The case that I had in mind is for a user that `git push`es the same commit and the first attempt fails to build while the second succeeds. In which case when a device picks the successful one and tries to patch its state, this `.get` could (before your change) return the failed as well.

src/features/device-state/routes/state-get-v3.ts
Page-: Yeah this is intended to be `is_managed_by__device` to match the model (as we do for other names) - I've moved the state-get change to the right commit/PR now

src/features/device-state/routes/state-patch-v3.ts
Page-: Yeah this is intended to be `is_managed_by__device` to match the model (as we do for other names)

Page-: Yeah this is intended to be `is_managed_by__device` to match the model (as we do for other names)

src/features/device-state/routes/state-patch-v3.ts
Page-: Done

src/features/device-state/routes/state-patch-v3.ts
Page-: If we don't skip it then we'll return a 401 presumably and we get the issue where after moving apps the supervisor might not be able to report its current state. That said I think you're right that this is already covered with the size check

thgreasi:
Only NITs are left from my side, so LGTM :clap: 
(Would love the v3 state patch test though to no be under a v2 state patch `describe()` :grimacing: )

src/features/device-state/routes/state-patch-v3.ts
Page-: Yeah I think the earlier check would handle this

src/features/device-state/routes/state-patch-v3.ts
Page-: It's not valid but also this endpoint could theoretically be called by anything

test/test-lib/fake-device.ts
```diff
@@ -54,6 +54,12 @@ export async function patchState(
 	patchBody: StatePatchV3Body,
 	stateVersion: 'v3',
 ): Promise<void>;
+export async function patchState(
+	user: UserObjectParam,
+	deviceUuid: string,
+	patchBody: StatePatchV2Body | StatePatchV3Body,
+	stateVersion: 'v2' | 'v3',
+): Promise<void>;
```
Page-, line 62: This is essentially a duplicate of the final signature and shouldn't be necessary? @thgreasi ?

thgreasi: @Page- The final implementation signature isn't considered to be one of the overloads that we can use (iff there is any overload defined). So duplicating it here allows us to pass a variable of type `'v2' | 'v3'` like we have in the test cases.

thgreasi (approved):

PR #743, opened 2021-08-21
Add device names

chrisys:
I've done the sorting/addition in two commits so it's easier to see the changes but can squash before merge if preferred.

I thought it would be good to add some more options since with larger fleets now it's quite common to see bunches of devices all with the same name.


thgreasi (approved):

chrisys:
@balena-ci rebase

PR #745, opened 2021-08-23
Allow specifying a custom readTransaction function for state requests

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #750, opened 2021-08-24
Use native pinejs advisory lock method

thgreasi:
Change-type: patch
Depends-on: https://github.com/balena-io/pinejs/pull/495


thgreasi:
@balena-ci I self-certify!

PR #752, opened 2021-08-25
Remove BALENA_|RESIN_ prefixes

ab77:
Change-type: patch


thgreasi (commented):
LGTM, but I would change this to a minor, so that's clearer that this might be breaking for pre-existing installations (since for 0.x.x semver, minor changes are allowed to be breaking).

ab77:
> LGTM, but I would change this to a minor, so that's clearer that this might be breaking for pre-existing installations (since for 0.x.x semver, minor changes are allowed to be breaking).

Good point.

ab77:
@dfunckt @Page- I think this is safe to merge since we manually bump the base version in balena-api

PR #755, opened 2021-08-30
Trigger updates when image env vars change like for other env vars

Page-:
Change-type: patch


Page-:
@balena-ci rebase

Page-:
@balena-ci I self-certify!

PR #758, opened 2021-09-02
Inject OS image development mode flag into generated config

dfunckt:
Change-type: patch


dfunckt:
@balena-ci I self-certify!

PR #767, opened 2021-09-21
Promisify redis methods once for logs rather than wrapping each call

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #768, opened 2021-09-21
Promisify the redis prototype rather than each instance individually

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #769, opened 2021-09-21
Only schedule publishing logs when we have some in the buffer

Page-:
This avoids idle connections from polling for new logs and instead are
scheduled automatically once a log comes in

Change-type: patch


Page-:
@balena-ci rebase

Page-:
@balena-ci I self-certify!

Issue #697, opened 2021-06-21
Device filter logical operators do not work as intended


cmfcruz:
The logical operators when filtering devices using the created_at field do not work as intended. The behavior is as follows:
- `gt` works the same as `ge`.
- `le` works the same as `lt`.
- `eq` returns an empty set even when there should be a result.
- `ne` still includes the device with the same `created_at` time.

PR #770, opened 2021-09-21
Adding tests for datetime filters

fisehara:
Adding test cases for datetime ODate filters: gt, le, eq, ne

Connects-to: #697

Change-type: patch


fisehara:
The issue is solved in pinejs and pinejs modules. Here the test cases are introduced to cover the issue and demonstrate the fixing of them.

PR #771, opened 2021-09-22
Fix check if publish is already scheduled

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #784, opened 2021-10-14
Update dependencies

Page-:
Update pinejs from 14.30.3 to 14.33.0

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #794, opened 2021-10-19
Run both write log context gathering queries in the same transaction

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #820, opened 2021-10-29
Restrict supervisor release population to releases of the balena_os org

thgreasi:
The cloud API has a rule which enforces this
restriction, which means that a PATCH of a device's
supervisor_version could fail if this code returned
a release of a different public app that has the same
release_version.

Change-type: patch


thgreasi:
@balena-ci I self-certify!

PR #821, opened 2021-11-01
Change the supervisor releases to require & use the semver field [part1]

thgreasi:
Connects-to: #811
Change-type: minor
Signed-off-by: Thodoris Greasidis <thodoris@balena.io>

PR #825, opened 2021-11-01
Stop requiring a release version for the supervisor releases

thgreasi:
On top of #821, so let's get that in first.

Connects-to: #811
Change-type: minor


dfunckt (commented):
👍

thgreasi:
@balena-ci I self-certify!

PR #828, opened 2021-11-04
Revert "Avoid unnecessary conversions of buffers to strings in redis based logs"

Page-:
This doesn't seem to provide the expected performance improvement and can introduce issues (the issues are potentially the cause of not gaining a performance improvement?)


Page-:
@balena-ci I self-certify!

PR #829, opened 2021-11-05
Avoid starting a transaction if there's no updates for device-state

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #838, opened 2021-11-09
Refactor the state patch tests to use the expectResourceToMatch helper

thgreasi:
Reduces the amount of code in the tests.

Change-type: patch


thgreasi:
@balena-ci I self-certify!

PR #849, opened 2021-11-12
cache: Abstract the req permissions key normalizer

thgreasi:
Change-type: patch


src/infra/cache/permissions.ts
```diff
@@ -0,0 +1,11 @@
+import type { permissions } from '@balena/pinejs';
```
Page-, line 1: Can you rename this file `permissions-normalizer` please, when I saw the `cache/permissions` I assumed some form of permissions cache

thgreasi:
@balena-ci I self-certify!

PR #852, opened 2021-11-16
Add the option for a read-only redis endpoint for reading device logs

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #857, opened 2021-11-18
Switch to using a single shared ioredis client

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #862, opened 2021-11-19
Add support for a separate redis instance for device logs

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #865, opened 2021-11-22
Fix service installs query when changing target release of huge fleets

thgreasi:
Change-type: patch
See: https://github.com/balena-io/balena-api/issues/3445
See: https://www.flowdock.com/app/rulemotion/public-s-community/threads/dlrnw-6igCqOTWKboUm65vyDB1k

### DIFF
--- a/src/features/ci-cd/hooks/service-installs.ts
+++ b/src/features/ci-cd/hooks/service-installs.ts
@@ -1,6 +1,6 @@
 import * as _ from 'lodash';
 import { sbvrUtils, hooks, permissions } from '@balena/pinejs';
-import type { Filter } from 'pinejs-client-core';
+import type { Filter, FilterObj } from 'pinejs-client-core';
 import type {
 	Device,
 	PickDeferred,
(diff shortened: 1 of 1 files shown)


src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -10,10 +10,21 @@ import type {
 
 const createReleaseServiceInstalls = async (
 	api: sbvrUtils.PinejsClient,
-	deviceIds: number[],
+	deviceFilterOrIds: number[] | FilterObj,
 	releaseFilter: Filter,
 ): Promise<void> => {
-	if (deviceIds.length === 0) {
+	const deviceIds = Array.isArray(deviceFilterOrIds)
```
dfunckt, line 16: `deviceIds` is not used anywhere.

dfunckt: Ugh, never mind -- it is used later on and in fact we do a `Promise.map` on the array which might contain 66k IDs? :/

Page-: It's used in
```
	await Promise.all(
		deviceIds.map(async (deviceId) => {
```
below (typescript would complain if it was unused)

thgreasi: It is still used near the end of the function, where we create the service installs missing for each device.

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -59,7 +70,17 @@ const createReleaseServiceInstalls = async (
...
+				device: Array.isArray(deviceFilterOrIds)
+					? { $in: deviceFilterOrIds }
+					: {
+							$any: {
+								$alias: 'd',
+								$expr: {
+									d: deviceFilterOrIds,
+								},
+							},
+					  },
```
thgreasi, line 83: Arguably, doing a $in might be improving the performance of this query for smaller fleets, but on the other hand I ended up preferring a uniform approach depending on the provided parameter type, rather than creating a fleet size based edge case.

thgreasi:
@balena-ci I self-certify!

PR #867, opened 2021-11-22
Disable auto pipelining in subscriber mode as it makes no sense there

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #873, opened 2021-11-25
Allow device api keys to query their state as root

Page-:
This omits all the permissions checks and results in a more efficient
query but with the downside that it requires us to manually check that
this is a safe operation to allow

Change-type: patch

### DIFF
--- a/src/features/device-state/routes/state.ts
+++ b/src/features/device-state/routes/state.ts
@@ -13,8 +13,9 @@ import {
 	serviceInstallFromImage,
 	setMinPollInterval,
 } from '../utils';
-import { sbvrUtils, errors, dbModule } from '@balena/pinejs';
+import { sbvrUtils, errors, dbModule, permissions } from '@balena/pinejs';
 import { events } from '..';
+import { isApiKeyWithRole } from '../../api-keys/lib';
 
 const { UnauthorizedError } = errors;
 const { api } = sbvrUtils;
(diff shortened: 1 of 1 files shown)


src/features/device-state/routes/state.ts
```diff
@@ -329,6 +330,38 @@ const stateQuery = _.once(() =>
...
+}: {
+	uuid: string;
+	req: permissions.PermissionReq;
+	tx: Tx;
+}) => {
+	if (typeof req.apiKey?.key !== 'string') {
+		return false;
+	}
+	// Must be a device key
+	if (!(await isApiKeyWithRole(req.apiKey?.key, 'device-api-key'))) {
```
dfunckt, line 350: I  think you should be able to do

```suggestion
	if (!(await isApiKeyWithRole(req.apiKey.key, 'device-api-key'))) {
```

here, right?

src/features/device-state/routes/state.ts
```diff
@@ -329,6 +330,38 @@ const stateQuery = _.once(() =>
...
+		return false;
+	}
+	// And must have access to the device
+	const device = api.resin.prepare<{ uuid: string }>({
```
dfunckt, line 354: Shouldn't this be

```suggestion
	const device = await api.resin.prepare<{ uuid: string }>({
```

?

dfunckt (approved):

Page-:
@balena-ci I self-certify!

PR #874, opened 2021-11-26
Limit device access to service installs to those of accessible devices

thgreasi:
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/vQ8zyst61JYAgZxvXEiCctJlmZx

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -90,7 +90,7 @@ export const DEVICE_API_KEY_PERMISSIONS = [
 	`resin.device.create?belongs_to__application/any(a:a/depends_on__application/any(da:da/${ownsDevice}))`,
 	`resin.device.update?belongs_to__application/any(a:a/depends_on__application/any(da:da/${ownsDevice}))`,
 
-	'resin.service.read?application/canAccess()',
+	'resin.service.read?application/canAccess() or service_install/canAccess()',
 
 	'resin.service_install.read?device/canAccess()',
 	...writePerms(
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -19,7 +19,6 @@ export const writePerms = (resource: string, filter: string): string[] => [
 const matchesActor = 'actor eq @__ACTOR_ID';
 const matchesUser = `user/any(u:u/${matchesActor})`;
 const ownsDevice = `owns__device/any(d:d/${matchesActor})`;
-const applicationControlsDevice = `application/any(a:a/${ownsDevice} or a/depends_on__application/any(da:da/${ownsDevice}))`;
 
 export const ROLES: {
 	[roleName: string]: string[];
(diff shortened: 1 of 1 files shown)


src/lib/auth.ts
```diff
@@ -93,10 +92,10 @@ export const DEVICE_API_KEY_PERMISSIONS = [
 
 	'resin.service.read?application/canAccess()',
 
-	'resin.service_install.read?installs__service/canAccess()',
+	'resin.service_install.read?device/canAccess()',
```
thgreasi, line 95: This allowed device to read the service_installs & all device_service_environment_variables of all other devices of the application.

thgreasi, line 95: This granted to a device api key read access to all service_installs & device_service_environment_variables of all devices of the application.

src/lib/auth.ts
```diff
@@ -93,10 +92,10 @@ export const DEVICE_API_KEY_PERMISSIONS = [
...
 	...writePerms(
 		'resin.service_install',
-		`installs__service/any(s:s/${applicationControlsDevice})`,
+		`device/any(d:d/${matchesActor} or d/belongs_to__application/any(a:a/depends_on__application/any(da:da/${ownsDevice})))`,
```
thgreasi, line 98: This again granted write access to all service_installs of all device of the fleet.

src/lib/auth.ts
```diff
@@ -91,12 +90,12 @@ export const DEVICE_API_KEY_PERMISSIONS = [
 	`resin.device.create?belongs_to__application/any(a:a/depends_on__application/any(da:da/${ownsDevice}))`,
 	`resin.device.update?belongs_to__application/any(a:a/depends_on__application/any(da:da/${ownsDevice}))`,
 
-	'resin.service.read?application/canAccess()',
+	'resin.service.read?application/canAccess() or service_install/canAccess()',
```
Page-, line 93: I think you can probably ditch the `application/canAccess()`?
```suggestion
	'resin.service.read?service_install/canAccess()',
```

thgreasi: @Page- I think that we can't since if we do so, then when a device is patching its own `should_be_running__release` then the `createReleaseServiceInstalls()` will not be able to retrieve the services for which it needs to create the service installs for (except of course if we change that to use root permissions).

See: https://github.com/balena-io/open-balena-api/blob/master/src/features/ci-cd/hooks/service-installs.ts#L11-L62

See: https://github.com/balena-io/open-balena-api/pull/877

thgreasi:
Execution plan with `resin.service.read?application/canAccess() or service_install/canAccess()`:
[device state perms PR 874.txt](https://github.com/balena-io/open-balena-api/files/7621876/device.state.perms.PR.874.txt)

src/lib/auth.ts
thgreasi: @Page- even with the `resin.service.read?application/canAccess() or service_install/canAccess()`, I think the state endpoint query is quite close to the fully optimized one you ended up:
[device state perms PR 874.txt](https://github.com/balena-io/open-balena-api/files/7621876/device.state.perms.PR.874.txt)

The only diff I could spot was dependent devices related and had nothing to do with services.

Page-: Removing the branch `application/canAccess()` won't add any optimizations (it'll actually just remove them if going `application -> service`), I just thought it'd probably be unnecessary with the addition of the `service_install/canAccess()`, but you're right I hadn't considered that the service installs need to be added somehow

thgreasi: I have to say that the way of how cleverly the extra permission checks are omitted in this `application/canAccess() or service_install/canAccess()` case amazed me while I was giving this a try yesternight.

thgreasi:
@balena-ci I self-certify!

PR #875, opened 2021-11-26
Allow access to all service installs for the device-api-key's device

Page-:
Change-type: patch


Page-:
@balena-ci rebase

Page-:
#874

PR #878, opened 2021-12-01
Set download progress to null when marking image installs as deleted

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

thgreasi:
:+1:

PR #879, opened 2021-12-01
Use canAccess for device key dependent device permissions and remove device state root querying

Page-:
Update pinejs from 14.34.0 to 14.35.0

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #884, opened 2021-12-03
device-state: Drop unused device config filtering implementation

thgreasi:
Change-type: patch


thgreasi:
@dfunckt afaict this was added for https://github.com/balena-io/balena-api/commit/ca74bb6f62a5aee438aa96b5961ec3ab97eb4810 but then it wasn't used even in the final state of the code that got merged https://github.com/balena-io/balena-api/pull/2047/files .
So I guess it's fine to drop as unused atm right?

dfunckt:
It should be fine to drop, yes @thgreasi

thgreasi:
@balena-ci I self-certify!

PR #886, opened 2021-12-07
Add throttling for image install updates

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

src/lib/config.ts
```diff
@@ -243,6 +243,9 @@ export const HIDE_UNVERSIONED_ENDPOINT = boolVar(
 export const METRICS_MAX_REPORT_INTERVAL_SECONDS = Math.round(
 	intVar('METRICS_MAX_REPORT_INTERVAL', 1 * MINUTES) / 1000,
 );
+export const DOWNLOAD_PROGRESS_MAX_REPORT_INTERVAL_SECONDS = Math.round(
+	intVar('DOWNLOAD_PROGRESS_MAX_REPORT_INTERVAL_SECONDS', 30) / 1000,
```
thgreasi, line 247: So the default value is 30ms?
Should this be:
```
intVar('DOWNLOAD_PROGRESS_MAX_REPORT_INTERVAL', 30) / 1000,
```
or
```
intVar('DOWNLOAD_PROGRESS_MAX_REPORT_INTERVAL_SECONDS', 30 * SECONDS) / 1000,
```

src/features/device-state/routes/state-patch.ts
```diff
@@ -256,6 +261,57 @@ const shouldMetricsUpdate = (() => {
...
+			lastReport.status !== body.status ||
+			// Or if the release has changed
+			lastReport.is_provided_by__release !== body.is_provided_by__release ||
+			// Or if the download progress has hit a milestone...
+			// From not downloading to downloading
+			(lastReport.download_progress == null &&
+				body.download_progress != null) ||
+			// From downloading to not downloading
+			(lastReport.download_progress != null &&
+				body.download_progress == null) ||
```
thgreasi, line 299: Had to read this twice to get it.
We can combine those two to v right?
```
(lastReport.download_progress == null) != (body.download_progress == null)
```

src/features/device-state/routes/state-patch.ts
```diff
@@ -256,6 +261,57 @@ const shouldMetricsUpdate = (() => {
...
+			// Hits 100%
+			body.download_progress === 100
+		) {
+			// And we add a new entry
+			await lastImageInstallReport.set(key, {
+				download_progress: lastReport?.download_progress,
```
thgreasi, line 305: Why are we only assigning the `download_progress`?
Could/Shouldn't this be just destructuring the whole `lastReport`?

src/lib/config.ts
thgreasi: Oh I see that https://github.com/balena-io/open-balena-api/pull/887 fixed that :+1:

src/features/device-state/routes/state-patch.ts
Page-: Right I meant to add a comment to this as well, but basically we want to keep the last reported download progress if the current report doesn't include it

src/features/device-state/routes/state-patch.ts
Page-: We can combine them but I had them separate to follow the comment chain cleanly

src/features/device-state/routes/state-patch.ts
Page-: https://github.com/balena-io/open-balena-api/pull/888

thgreasi: @Page- my question was mostly why we care only about the `download_progress` and we didn't do:
```
				...lastReport,
				...body,
				updateTime: now,
```

PR #889, opened 2021-12-08
Add info about the contract that failed to parse when throwing an error

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #890, opened 2021-12-08
Limit changing the app DT only to ones of the same cpu architecture

thgreasi:
Change-type: minor
See: https://github.com/balena-io/balena-api/issues/3463


thgreasi:
@balena-ci I self-certify!

PR #896, opened 2021-12-10
patch: Add required_reboot properties to config variables

vipulgupta2048:
1. Added required_reboot and warning message properties to the schema
2. Format of warning messages needs to be decided - Decided to limit the scope of the PR and will add warning messages iteratively. 

TODO

- [ ] Some properties don't exist in this scheme file. We need to sync it up. For example: BALENA_HOST_EXTLINUX_isolcpus (Will do another PR for this)

Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipul@balena.io>

### DIFF
--- a/src/features/vars-schema/env-vars.ts
+++ b/src/features/vars-schema/env-vars.ts
@@ -16,6 +16,11 @@ const addReservedPrefixes = (array: string[]) => {
 	);
 };
 
+type ConfigVarDefinition = JSONSchema6Definition & {
+	will_reboot?: boolean;
+	warning?: string;
+};
+
 // Config variables that are allowed to be set externally
 // Note that this list will be mutated by the cloud API to add
 // extra variables as for cloud-only features.
@@ -56,7 +61,7 @@ export const INVALID_NEWLINE_REGEX = /\r|\n/;
 // Note that this list will be mutated by the cloud API to add
 // extra variables as for cloud-only features.
 export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
-	[k: string]: JSONSchema6Definition;
+	[k: string]: ConfigVarDefinition;
 } = {
 	RESIN_SUPERVISOR_CONNECTIVITY_CHECK: {
 		enum: ['false', 'true'],
@@ -84,6 +89,7 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
 		enum: ['false', 'true'],
 		description:
 			'Enable persistent logging. Only supported by supervisor versions >= v7.15.0.',
+		will_reboot: true,
 		default: 'false',
 	},
 	RESIN_SUPERVISOR_INSTANT_UPDATE_TRIGGER: {
@@ -105,6 +111,7 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
 		description:
 			'Define the PNG image to be used for the boot splash screen. Only supported by supervisor versions >= v12.3.0.',
 		maxLength: 13400, // ~10KB base64 encoded image
+		will_reboot: true,
 	},
 	BALENA_SUPERVISOR_HARDWARE_METRICS: {
 		enum: ['false', 'true'],
@@ -116,7 +123,7 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
 
 export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
 	capableDeviceTypes: string[];
-	properties: Dictionary<JSONSchema6Definition>;
+	properties: Dictionary<ConfigVarDefinition>;
 }> = [
 	{
 		capableDeviceTypes: [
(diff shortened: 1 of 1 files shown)


vipulgupta2048:
@cywang117 @20k-ultra Please feel free to decide the format of the warning message and amend any `required_reboot` property that I might have missed.

src/features/vars-schema/env-vars.ts
```diff
@@ -56,16 +56,31 @@ export const INVALID_NEWLINE_REGEX = /\r|\n/;
...
 } = {
 	RESIN_SUPERVISOR_CONNECTIVITY_CHECK: {
 		enum: ['false', 'true'],
 		description: 'Enable / Disable VPN connectivity check',
+		properties: {
+			required_reboot: false,
+			warning: '',
+		},
 		default: 'true',
 	},
```
20k-ultra, line 76: I'm not sure how much discussion was hard regarding the schema for these object but I don't think having a properties object is that useful since the description and default keys are properties as well. Additionally, I think leaving the warning message to be handled by the UI makes more sense. What do you think about this:

```suggestion
	RESIN_SUPERVISOR_CONNECTIVITY_CHECK: {
		enum: ['false', 'true'],
		description: 'Enable / Disable VPN connectivity check',
		reboots: false
		default: 'true',
	},
```

20k-ultra (commented):
If you have any questions about if a config causes a reboot reference https://github.com/balena-os/balena-supervisor/blob/master/docs/configurations.md

src/features/vars-schema/env-vars.ts
cywang117: Agree with Miguel's suggestion, although maybe `requires_reboot` instead of `reboots`, but that's up to you @vip

20k-ultra: I didn't like `requires_reboot` because if I'm telling you something requires a reboot it's implying that you have to do the reboot after changing the value but the client nor the API do the rebooting so why say "requires".  You can just say the variable is going to reboot the device. `will_reboot` is another option.

cywang117: `will_reboot` sounds good too 👍🏼

vipulgupta2048: LOL, @thgreasi felt weird too about this. I'm glad more people are having the tingling sensation about this. I like `will_reboot` too. 

@cywang117 @20k-ultra Checkout the product call for decision on this, I have marked the timestamp from when you need to hear the call from. It's important in order for us to get on the same page: https://jel.ly.fish/brainstorm-topic-find-a-short-term-solution-for-notifying-users-that-a-config-variable-will-reboot-restart-the-device-ce3e30d0-879a-4bb7-a1e0-934d21dadda4

TLDR though: 
1. We went with having an additional properties property in order not to break JSONSchema and instead extend it. (check call for the discussion)
2. Both properties in the `properties` object are warnings. Check this out, 

IF you have `will_reboot: true` then UI handles the warning and shows a default message about the device/fleet going to reboot if you make this change. 

IF you have a `warning` message then UI shows the warning when this config variable is changed. These warnings are custom messages we want to give to the user about what's going to happen. Examples: You device might disconnect, you might blow up in 60 seconds, your device will now have a persistent journal that is going to take up space. 

IF both are there and present, then we show a composite warning. Merging both of them to be shown together.

dfunckt: Thanks for the overview @vipulgupta2048.

The structure should then look like:

```suggestion
	RESIN_SUPERVISOR_CONNECTIVITY_CHECK: {
		enum: ['false', 'true'],
		description: 'Enable / Disable VPN connectivity check',
		default: 'true',
	},
```

If that config var triggered a reboot, then it should look like:

```suggestion
	RESIN_SUPERVISOR_CONNECTIVITY_CHECK: {
		enum: ['false', 'true'],
		description: 'Enable / Disable VPN connectivity check',
		properties: {
			requires_reboot: true,
		},
		default: 'true',
	},
```

If it should give out a warning when changed, then it should be:

```suggestion
	RESIN_SUPERVISOR_CONNECTIVITY_CHECK: {
		enum: ['false', 'true'],
		description: 'Enable / Disable VPN connectivity check',
		properties: {
			warning: 'Lorem ipsum ...',
		},
		default: 'true',
	},
```

The principle being that the absence of these properties cause no extra behaviour to be triggered.

> IF both are there and present, then we show a composite warning. Merging both of them to be shown together.

I don't understand what a composite warning is. Can you elaborate?

src/features/vars-schema/env-vars.ts
```diff
@@ -56,7 +56,14 @@ export const INVALID_NEWLINE_REGEX = /\r|\n/;
...
 export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
-	[k: string]: JSONSchema6Definition;
+	[k: string]:
+		| JSONSchema6Definition
+		| {
+				properties: {
+					will_reboot?: boolean;
+					warning?: string;
+				};
+		  };
```
dfunckt, line 66:
```suggestion
		JSONSchema6Definition & {
				properties?: {
					will_reboot?: boolean;
					warning?: string;
				};
		  };
```

vipulgupta2048: Thanks @dfunckt 

> | JSONSchema6Definition

Good catch on this, this is incorrect but I used a bitwise OR (`|`)(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_OR) here instead of a bitwise AND (`&`) because I feel we are looking for a union of types (addition) and using a `&` made vscode mad. So I used the single pipe `|`

The error I was getting when hovering over line 68.

```
Type '{ enum: string[]; description: string; default: string; }' is not assignable to type 'JSONSchema6Definition & { properties: { will_reboot?: boolean | undefined; warning?: string | undefined; }; }'.
  Type '{ enum: string[]; description: string; default: string; }' is not assignable to type 'JSONSchema6 & { properties: { will_reboot?: boolean | undefined; warning?: string | undefined; }; }'.
    Property 'properties' is missing in type '{ enum: string[]; description: string; default: string; }' but required in type '{ properties: { will_reboot?: boolean | undefined; warning?: string | undefined; }; }'.ts(2322)
```

Screenshot from 2021-12-13 19-13-01

dfunckt: These operators work like *logical* AND/OR, not bitwise @vipulgupta2048. Counter-intuitive, right? `&` creates the union of two types. `|` on the other hand for, say, `A | B` is "A *or* B".

The reason you get all these errors with `&` is because for most vars you don't have the `properties` attribute and it is declared as non-optional. To fix this you just need to declare it as `properties?: { ... }` (note the question mark) which also semantically better -- i.e. it's existence is optional and introduces new behaviour beyond the default.

src/features/vars-schema/env-vars.ts
vipulgupta2048: Right on @dfunckt, Thodoris also gave me similar advice and I have gone for keeping the properties truthy and only present if things exist. Check out my latest commits for that.

The composite warning is shown when both are present and truthy. For example, consider these to be the values: 

```
will_reboot: true, 
warning: "Device is going to be hot."
```

Hence, we will show a "composite" of the 2 warnings in the UI. Something like. 

"The device will reboot when you make this change. 
Warning: The device is going to be hot."

src/features/vars-schema/env-vars.ts
vipulgupta2048: > logical AND/OR

I tried to follow the docs on this one instead of what was actually happening. Hence, ended up choosing the single pipe `|`.  

Good touch on the optional properties object @dfunckt on my first pass, I marked properties of the `properties` object as optional. But this is way cleaner.

src/features/vars-schema/env-vars.ts
```diff
@@ -16,6 +16,11 @@ const addReservedPrefixes = (array: string[]) => {
 	);
 };
 
+interface ConfigVarDefinition {
```
thgreasi, line 19:
```suggestion
type ConfigVarDefinition = JSONSchema6Definition & {
```

thgreasi (dismissed):

vipulgupta2048:
@balena-ci I self-certify!

PR #899, opened 2021-12-13
patch: Add BALENA_HOST_EXTLINUX_isolcpu

vipulgupta2048:
Picked this from here: https://github.com/balena-os/balena-supervisor/blob/master/docs/configurations.md

Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipul@balena.io>

### DIFF
--- a/src/features/vars-schema/env-vars.ts
+++ b/src/features/vars-schema/env-vars.ts
@@ -267,6 +267,13 @@ export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
 					'Define the file name of the DTB to be used. Only supported by supervisor versions >= v11.14.2.',
 				will_reboot: true,
 			},
+			RESIN_HOST_EXTLINUX_isolcpus: {
+				type: 'string',
+				description:
+					'Allows to isolate CPU cores from the kernel scheduler by specifying CPU cores in the system starting from 0. Only supported by supervisor versions >= v7.25.0',
+				examples: ['0,2,3'],
+				will_reboot: true,
+			},
 		},
 	},
 	{
--- a/test/01_basic.ts
+++ b/test/01_basic.ts
@@ -129,12 +129,16 @@ describe('Basic', () => {
 			},
 			{
 				deviceType: 'jetson-nano',
-				extraConfigVarSchemaProperties: ['RESIN_HOST_EXTLINUX_fdt'],
+				extraConfigVarSchemaProperties: [
+					'RESIN_HOST_EXTLINUX_fdt',
+					'RESIN_HOST_EXTLINUX_isolcpus',
+				],
 			},
 			{
 				deviceType: 'jetson-tx2',
 				extraConfigVarSchemaProperties: [
 					'RESIN_HOST_EXTLINUX_fdt',
+					'RESIN_HOST_EXTLINUX_isolcpus',
 					'RESIN_HOST_ODMDATA_configuration',
 				],
 			},


thgreasi (commented):
LGTM :+1: but this should be a minor :)
(Self-certify after you change it)

src/features/vars-schema/env-vars.ts
```diff
@@ -267,6 +267,12 @@ export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
 					'Define the file name of the DTB to be used. Only supported by supervisor versions >= v11.14.2.',
 				will_reboot: true,
 			},
+			BALENA_HOST_EXTLINUX_isolcpus: {
+				type: 'string',
+				description:
+					'Allows to isolate CPU cores from kernel scheduler. See kernel parameters. Only supported by supervisor versions >= v7.10.0.',
```
thgreasi, line 273: Would be nice to rephrase the `See kernel parameters` to `...using/by specifying kernel parameters`.
Bonus points if you manage to get an example as well. Afaict `1,2,3` might be an example, but plz confirm that.

src/features/vars-schema/env-vars.ts
```diff
@@ -267,6 +267,12 @@ export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
 					'Define the file name of the DTB to be used. Only supported by supervisor versions >= v11.14.2.',
 				will_reboot: true,
 			},
+			BALENA_HOST_EXTLINUX_isolcpus: {
```
thgreasi, line 270: Atm we have about 3.5k records using the `RESIN_HOST_EXTLINUX_isolcpus` and none using the `BALENA_` one.
As a result using `BALENA_HOST_EXTLINUX_isolcpus` would still make their config var show as a custom one and will not show up in the schema generated UI form.
Moreover, since we state that this was added in `v7.10.0` and we only started supporting `BALENA_` in v7.25.0, this could potentially cause confusion if anyone tries to use it in v7.10.0.
See: https://github.com/balena-os/balena-supervisor/blob/master/CHANGELOG.md#v7250---2018-10-18

Maybe we should use `RESIN_HOST_EXTLINUX_isolcpus` for now and wait for the RESIN_ & BALENA_ unification to take care of all prefixes in one go.

vipulgupta2048: Right on, will change the supported by field

src/features/vars-schema/env-vars.ts
vipulgupta2048: Very less information, I found all this and pushed the example. 
- Open this link using Chrome --> https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_for_real_time/7/html/tuning_guide/isolating_cpus_using_tuned-profiles-realtime#:~:text=swap%20console%3DttyS0%2C115200n81-,isolcpus%3D0%2C4,-Isolating%20CPUs%20Using

- https://rt-labs.com/docs/p-net/linuxtiming.html#run-the-application-on-a-separate-processor-core

- https://www.suse.com/support/kb/doc/?id=000017747

vipulgupta2048: okay, this format of the example is confirmed. Check FD conversation: https://www.flowdock.com/app/rulemotion/r-supervisor/threads/g5SxblwceDmrBTT8ROiR3BQCyIm

vipulgupta2048:
@thgreasi Tests were failing because this new config variable was not part of the `extraConfigVarSchemaProperties` for specific deviceTypes. I amended that.

vipulgupta2048:
@balena-ci I self-certify!

PR #900, opened 2021-12-16
Rephrase multicontainer rule to start from the application type

Page-:
Update pinejs from 14.35.1 to 14.37.0

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #901, opened 2021-12-17
Trim .versionbot/CHANGELOG.yml so that it's < 1MB

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #904, opened 2021-12-17
Add index to optimize should be running successful release rule

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #905, opened 2021-12-20
Create service install when pinning supervisor on device creation

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #909, opened 2021-12-23
target-hostapp: Add support for detecting unified OS releases

thgreasi:
Change-type: minor
See: https://jel.ly.fish/improvement-combine-os-variants-0d96399a-660b-4d38-adca-f661378cb86e


thgreasi:
@balena-ci I self-certify!

PR #911, opened 2021-12-30
Add device configuration setting for terminating TLS on the device.

pranavpeshwe:
Introduce a device configuration variable (boolean) that allows turning ON/OFF TLS termination on the device.


ab77:
@pranavpeshwe FYI https://github.com/balena-io/environment-staging/pull/688 makes `/certs/export/chain.pem` available in balena-staging.com to the proxy container(s).

PR #919, opened 2022-01-10
features/device-types: Remove the special case for the "edge" DT

thgreasi:
It seems that atm it's not needed & on top of
that, the "edge" has been discontinued for
some time now.

Change-type: patch
See: https://github.com/balena-io-modules/resin-device-types/blob/3888375f38000b38dc7842b2540b3db3dc753d43/index.coffee#L10


thgreasi:
@balena-ci I self-certify!

PR #921, opened 2022-01-10
Rely only on the model for resolving device type aliases

thgreasi:
Change-type: minor


src/features/device-types/device-types.ts
```diff
@@ -1,10 +1,9 @@
 import * as arraySort from 'array-sort';
 import * as _ from 'lodash';
 
-import * as deviceTypesLib from '@resin.io/device-types';
+import type { DeviceType as DeviceTypeJson } from '@resin.io/device-types';
```
thgreasi, line 4: @dfunckt btw after this PR, we no longer use the code in `@resin.io/device-types` and given that the types that we import are already in custom typings in oB, we can just drop it.

dfunckt: Sounds great, why don't we just drop it with this PR?

thgreasi: Will do :+1:

thgreasi:
@balena-ci I self-certify!

PR #925, opened 2022-01-17
Use a hash tag for redis logs keys

Page-:
This allows specifying the part of the key to use for redis hashing to
ensure that related logs keys are grouped together

Change-type: minor


fisehara:
LGTM

As I didn't know about it I found a good read about these { } notation in redis key here:
https://redis.com/blog/redis-clustering-best-practices-with-keys/

fisehara (approved):
Makes sense to put every device ID in the same hash slot.
Any idea how we will know the performance increase?

PR #927, opened 2022-01-18
Remove 'balena-root-ca.service' dependency

bartversluijs:
Linked to issue https://github.com/balena-io/open-balena-api/issues/926

PR #928, opened 2022-01-18
Augment provisioning API key permissions for Post Hostapp hook

bartversluijs:
When provisioning a device, a [postHostappHook](https://github.com/balena-io/open-balena-api/blob/master/src/features/hostapp/hooks/target-hostapp.ts#L82-L119) is being run since commit https://github.com/balena-io/open-balena-api/commit/0eca892c689acf33c3d18b7f25b16ab1d90f4731 (v0.139.1).

This hook gets the OS release resource, which queries the [release, application and release tag tables](https://github.com/balena-io/open-balena-api/blob/master/src/features/hostapp/hooks/target-hostapp.ts#L193-L267). 
This is why the provisioning fails when running an openBalena instance. With those permissions added, it works fine.

Hopefully you guys will merge this!

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -8,7 +8,8 @@ export = {
 			username: 'guest',
 			password: ' ',
 			permissions: [
-				'resin.device_type.read',
+				// The "1 eq 1" is necessary so that we can do "device_type/canAccess()"
+				'resin.device_type.read?1 eq 1',
 				'resin.device_type_alias.read',
 				'resin.cpu_architecture.read',
 				'resin.device_family.read',
--- a/src/features/device-provisioning/register.ts
+++ b/src/features/device-provisioning/register.ts
@@ -53,11 +53,18 @@ export const register: RequestHandler = async (req, res) => {
 		} = req.body;
 		const deviceApiKey = req.body.api_key ?? randomstring.generate();
 
-		// Temporarily augment the api key with the ability to fetch the device we create and create an api key for it
+		/**
+		 * Temporarily augment the api key with the ability to:
+		 * - Fetch the device we create & create an api key for it
+		 * - Read the hostApp releases that should be operating the device
+		 */
 		req = augmentReqApiKeyPermissions(
 			req,
 			'resin.device.read',
 			'resin.device.create-device-api-key',
+			`resin.application.read?is_public eq true and is_host eq true and is_for__device_type/canAccess()`,
+			'resin.release.read?belongs_to__application/canAccess()',
+			`resin.release_tag.read?release/canAccess()`,
 		);
 
 		const response = await sbvrUtils.db.transaction(async (tx) => {


thgreasi:
Hi,
Can you clarify the provisioning process that you are following.
Is this error that you are getting after generating a config.json with a provisioning api key, configuring a balenaOS image & spinning up a device w/ that image?
(It shouldn't really matter, but I would love to hear which balenaOS version you have tried.)

thgreasi:
Hi,
I've created a test case for what I think you are facing, which atm fails, and it seems that this PR fixes it.
See: https://github.com/balena-io/open-balena-api/pull/929

Let me go through your proposed changes to confirm whether that's the least permissive option for fixing this and get back to you.

PS: Please clarify whether that's close to the flow you are following.

bartversluijs:
Hi @thgreasi,

Thanks for looking into this PR.
This error is indeed after generating a config.json with a provisioning API key (using `balena os configure`) & spinning up the device with that image. The logs of the Balena supervisor shows that it keeps trying to provision, but fails with a 401 error.

After a lot of debugging, I figured out it had something to do with that hook and with the missing permissions. Wasn't really clear to me at first, because it was an Unauthorized Error and not a Forbidden Error, so I thought it had something to do with the API key being invalid or something.

I've tried the following versions, all failed with the same error:
- BalenaOS v2.65.0+rev1 (dev & prod)
- BalenaOS v2.80.5+rev1 (dev & prod)
- BalenaOS v2.88.4+rev0 (dev & prod)

---

Regarding the least permissive option, I don't really know all permissions, but using read permissions for the specific tables / stores (I don't know what you call them 🙂) shouldn't do much harm. But what I can think if is adding a check whether the `os_version` and `os_variant` is set, and if they're set, adding the permissions. It's a little more secure perhaps.

Let me know what you think!

(P.S., I'll mention you in my other issue, because I need that to work as well before I can create the openBalena update 😄)

thgreasi:
Hi @bartversluijs ,
I've cherry-picked your commit into my branch at
See: https://github.com/balena-io/balena-api/pull/3515
where I restricted the permissions a bit and I'm now waiting for a review on how this works with the rest of the platform.
I hope that is fine with you. If not please let me know if you would prefer a different approach. I can perhaps merge your commit into my branch instead cheppy-picking, but I would have to check whether that's allowed in our CI process.
As an alternative that would avoid cherry-picking, is if you keep your PR fully rebased, in which case I could place my commits on top of yours directly.

Kind regards,
Thodoris

thgreasi:
Hey,
Once you rebase, feel free to squash/cherrypick into this PR the changes I did in https://github.com/balena-io/open-balena-api/pull/929/commits/f5f234af7c8f927852cb7fc01bdb6f2765802834.

bartversluijs:
Hi @thgreasi,

Thanks for looking into this. I've followed your progress closely, because we're having some issues with the current OpenBalena instance running in v0.139.0 (high CPU / load on random moments). So how the code gets merged, doesn't really matter to me, the most important thing is that it's fixed 🎉 

I'm happy to rebase it and will do that and cherry pick your change into this PR. Then you can decide how you want to merge it into the code. Once that's done and #927, I can test the upgrade process again and update our instance.

bartversluijs:
I've rebased the branch and squashed your commit and left your commit message as-is.
We're using Versionist ourselfs and when 2 commits both have `Change-type: patch`, you'd probably end up with 2 new lines in the changelog, with more or less the same description of the change.

Thought that wasn't the desired outcome, so that's why I squashed it!

thgreasi:
@bartversluijs may I ask you to amend your commit so that it changes commit hash and re-push so that circleci build is triggered?
We had the third-party PRs setting disabled before this and I just enabled it.
Maybe you can by the way change the commit message to something like `Fix hostApp release linking on /device/register endpoint` to better describe "what" it does, rather than "how".

thgreasi (approved):

thgreasi:
Thanks for reporting, investigating, debugging and opening a PR for this @bartversluijs :clap:

PR #931, opened 2022-01-20
target-hostapp: Inline the hook definitions to increase readability

thgreasi:
I guess we were re-using them in more than
one places, but that's no longer the case and
since it's the Nth time I get lost in the file b/c
of this, I decided to at last make this humble
PR.

Change-type: patch


thgreasi:
@balena-ci I self-certify!

PR #935, opened 2022-01-20
Stop using the v5 app's & device's device_type property internally

thgreasi:
Change-type: patch

### DIFF
--- a/src/features/hostapp/hooks/target-hostapp.ts
+++ b/src/features/hostapp/hooks/target-hostapp.ts
@@ -85,8 +85,8 @@ hooks.addPureHook('POST', 'resin', 'device', {
 			args.request.values.os_version != null &&
 			args.request.values.os_variant != null
 		) {
-			let deviceType = args.request.values.is_of__device_type;
-			if (!deviceType) {
+			let deviceTypeId = args.request.values.is_of__device_type;
+			if (!deviceTypeId) {
 				const [dt] = await args.api.get({
 					resource: 'device_type',
 					options: {
@@ -99,14 +99,14 @@ hooks.addPureHook('POST', 'resin', 'device', {
 				if (!dt) {
 					return;
 				}
-				deviceType = dt.id;
+				deviceTypeId = dt.id;
 			}
 
 			const hostappRelease = await getOSReleaseResource(
 				args.api,
 				args.request.values.os_version,
 				args.request.values.os_variant,
-				deviceType,
+				deviceTypeId,
 			);
 			// since this is a POST, we _know_ the device is being created and has no current/target state, so we can
 			// just append the target after determining which it is (like a preloaded app)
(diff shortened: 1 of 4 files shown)


src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -85,8 +85,9 @@ hooks.addPureHook('POST', 'resin', 'device', {
 			args.request.values.os_version != null &&
 			args.request.values.os_variant != null
 		) {
-			let deviceType = args.request.values.is_of__device_type;
-			if (!deviceType) {
+			let deviceTypeId = args.request.values.is_of__device_type;
+			if (!deviceTypeId) {
+				// TODO: Drop in v7
```
Page-, line 90: What about v7 would allow us to drop this?

src/features/device-types/hooks.ts
```diff
@@ -49,6 +50,7 @@ hooks.addPureHook('POST', 'resin', 'application', {
 	},
 });
 
+// TODO: Drop in v7
```
Page-, line 53: What about v7 would allow us to drop this?

src/features/device-types/hooks.ts
```diff
@@ -34,6 +34,7 @@ export const resolveDeviceType = async (
 	request.values.device_type = dt.slug;
 };
 
+// TODO: Drop in v7
```
Page-, line 37: What about v7 would allow us to drop this?

src/features/device-types/hooks.ts
thgreasi: Just that the major bump allows to do breaking changes.
This looked like a leftover in oB when we replaced the string `device_type` props with the `is_of/for__device_type` device_type resource FKs, which we left on purpose, so that oB doesn't break with a half-changed unreleased model.

More specifically, given that the cloud API also has translations for these exact cases:
See: https://github.com/balena-io/balena-api/blob/e291c4aad2219f7cc5b9e7bfc10e881ffe449a7c/src/models/hooks/translations/v5.ts#L703
See: https://github.com/balena-io/balena-api/blob/e291c4aad2219f7cc5b9e7bfc10e881ffe449a7c/src/models/hooks/translations/v5.ts#L768
and that this is a divergence (arguable helpful though) from how we have so far handled such changes (moving from a primitive field to an FK), it felt to me that we just forgot to drop this once we bumped oB to v6.

I might be wrong though & I don't really remember whether this was discussed and done on purpose.
cc: @dfunckt in case he remembers.

Page-: @thgreasi that doesn't change anything does it? If v6 needs it then even if we waited for v7 then dropping it here would also drop it from v6. I'm assuming what should actually happen is this should already be moved to a v6 translation and that way it'll naturally disappear when someone queries v7

thgreasi: @Page- You mean that we should drop it from here right away and just move it to the v6 translation?
I was only thinking about deprecating it here so that we pile up all breaking changes (when possible) till we release v7, but you have a point that this should be in v6.ts translations right away...

thgreasi: Let me drop the deprecation notices and actually drop these in a follow-up.

thgreasi:
@balena-ci I self-certify!

PR #938, opened 2022-01-25
Device state cleanup

Page-:


Page-:
@balena-ci I self-certify!

PR #941, opened 2022-01-31
Add `penalty` method to created rate limiters

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #942, opened 2022-01-31
Re-throw the original error on unrecognized ratelimiting consume errors

Page-:
Change-type: minor


src/infra/rate-limiting/index.ts
```diff
@@ -73,9 +73,8 @@ export const createRateLimiter = (
 						undefined,
 						headers,
 					);
-				} else {
-					throw new InternalRequestError();
 				}
+				throw e;
```
Page-, line 77: The only usage of this within open-balena-api is below and will log the error to sentry and use a 500 anyway, so this change primarily allows that sentry logging

Page-:
@balena-ci I self-certify!

PR #945, opened 2022-02-04
Stop accepting the application&device device_type POST parameter

thgreasi:
That's a breaking change for this likely forgotten "translation", which will become a v6 translation in balenaCloud.

Change-type: minor
See: https://jel.ly.fish/55d814fc-d378-44ef-b9d9-17b42b81138f
See: https://github.com/balena-io/balena-api/pull/3542


dfunckt (dismissed):
Why do you want to do this? This PR means open-balena-api with its lack of support for translations will be unusable until we expose a v7 API and have updated all clients to use that.

thgreasi:
Even though that is a breaking oB change, it will not block using oB if you are using the proper `is_of/for__device_type` fields, which is what our SDKs & CLI do. That's in the effort of
* keeping the model pure (this is the only place we are doing something like that)
* removing this apparently forgotten "translation" we added when we migrated from `device_type: string` to `is_of/for__device_type: NavigationResource<DeviceType>`

The alternative would be to keep this around with a deprecation `TODO: Drop me in v7` comment till we release v7. That's though not what we did when we changed `is_on_commit` to `is_running__release` (so in that transitional period till v6 was released, some SDK's tests stopped passing in oB, but we were fine with that).

dfunckt:
> which is what our SDKs & CLI do

Right, if that's merely affecting direct API consumers but not CLI/SDK users I'm fine with that, I did not realize this is the case and that CLI/SDK had already migrated over to using the "new" property. Good stuff 👍

thgreasi:
Well, the v6 model capable SDK never used or had typings for it (hence the UI & CLI doesn't use it)  b/c I never knew this still existied :grimacing:

thgreasi:
@balena-ci retest

thgreasi:
@balena-ci retest

thgreasi:
@balena-ci I self-certify!

PR #946, opened 2022-02-04
Draft: Separate test from init

fisehara:


fisehara:
@balena-ci retest

test/test-lib/supertest.ts
```diff
@@ -20,3 +77,30 @@ export const supertest = function (user?: string | UserObjectParam) {
 	}
 	return req as ReturnType<typeof $supertest.agent>;
 };
+
+const dropDatabaseSchema = async (): Promise<void> => {
+	if (sbvrUtils.db) {
+		await sbvrUtils.db.transaction(async (tx) => {
+			try {
+				await tx.executeSql(
```
Page-, line 85: If you're running just a single query it's simpler/preferable to use `db.executeSql` directly:
```suggestion
		try {
			await sbvrUtils.db.executeSql(
```

fisehara: Changed accordingly

PR #948, opened 2022-02-09
Support generating named device keys

nitishagar:
Change-Type: minor

See: https://github.com/balena-io/balena-ui/pull/5289/files#diff-a2b7f88dce592a5e7d89be8ff43e1a4d9f79f012f51b10e0f3836c9d0742de16R155

### DIFF
--- a/src/features/api-keys/lib.ts
+++ b/src/features/api-keys/lib.ts
@@ -94,6 +94,22 @@ const $createApiKey = async (
 	return apiKey;
 };
 
+const getKeyMetadata = (reqBody: { name?: any; description?: any }) => {
+	const { name, description } = reqBody;
+
+	if (name != null && typeof name !== 'string') {
+		throw new errors.BadRequestError('Key name should be a string value');
+	}
+
+	if (description != null && typeof description !== 'string') {
+		throw new errors.BadRequestError(
+			'Key description should be a string value',
+		);
+	}
+
+	return { name, description };
+};
+
 export const createApiKey = async (
 	actorType: string,
 	roleName: string,
@@ -102,6 +118,15 @@ export const createApiKey = async (
 	options: ApiKeyOptions = {},
 ): Promise<string> => {
 	options.apiKey ??= randomstring.generate();
+	const { name, description } = getKeyMetadata(req.body);
+
+	if (!options.name) {
+		options.name = name;
+	}
+	if (!options.description) {
+		options.description = description;
+	}
+
 	if (options.tx != null) {
 		return await $createApiKey(
 			actorType,
--- a/src/features/api-keys/routes.ts
+++ b/src/features/api-keys/routes.ts
@@ -61,11 +61,7 @@ export const createProvisioningApiKey: RequestHandler = async (req, res) => {
 	}
 
 	try {
-		const apiKey = await $createProvisioningApiKey(
-			req,
-			appId,
-			req.body.name ? { name: req.body.name } : {},
-		);
+		const apiKey = await $createProvisioningApiKey(req, appId);
 		res.json(apiKey);
 	} catch (err) {
 		if (handleHttpErrors(req, res, err)) {
@@ -94,20 +90,16 @@ export const createUserApiKey: RequestHandler = async (req, res) => {
 };
 
 export const createNamedUserApiKey: RequestHandler = async (req, res) => {
-	const { name, description } = req.body;
-	if (!name) {
-		return res.status(400).send('API keys require a name');
-	}
-
 	try {
+		if (!req.body.name) {
+			throw new errors.BadRequestError('API keys require a name');
+		}
+
 		if (req.user == null) {
 			throw new errors.BadRequestError('Must use user auth');
 		}
 
-		const apiKey = await $createNamedUserApiKey(req, req.user.id, {
-			name,
-			description,
-		});
+		const apiKey = await $createNamedUserApiKey(req, req.user.id);
 		res.json(apiKey);
 	} catch (err) {
 		if (handleHttpErrors(req, res, err)) {
--- a/test/08_api-keys.ts
+++ b/test/08_api-keys.ts
@@ -27,11 +27,13 @@ describe('create provisioning apikey', function () {
 				user: UserObjectParam | undefined,
 				applicationId: number,
 				provisioningKeyName?: string,
+				provisioningKeyDescription?: string,
 			) {
 				return supertest(user)
 					.post(`/api-key/application/${applicationId}/provisioning`)
 					.send({
 						name: provisioningKeyName,
+						description: provisioningKeyDescription,
 					});
 			},
 		},
@@ -41,6 +43,7 @@ describe('create provisioning apikey', function () {
 				user: UserObjectParam | undefined,
 				applicationId: number,
 				provisioningKeyName?: string,
+				provisioningKeyDescription?: string,
 			) {
 				return supertest(user)
 					.post(`/api-key/v1/`)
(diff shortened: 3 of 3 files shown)


test/08_api-keys.ts
```diff
@@ -88,13 +93,18 @@ describe('create provisioning apikey', function () {
...
 				});
 
 				expect(apiKeyResp).to.have.property('name');
 				expect(apiKeyResp?.name).to.be.a('string');
 				expect(apiKeyResp?.name).to.equal(`provision-key-${applicationId}`);
+				expect(apiKeyResp).to.have.property('description');
+				expect(apiKeyResp?.description).to.be.a('string');
+				expect(apiKeyResp?.description).to.equal(
+					`Sample key for application-${applicationId} description.`,
+				);
```
thgreasi, line 107: This can just be:
```suggestion
				expect(apiKeyResp).to.have.property('description', `Sample key for application-${applicationId} description.`);
```

test/08_api-keys.ts
```diff
@@ -181,6 +191,39 @@ describe('create device apikey', function () {
...
+				});
+
+				expect(apiKeyResp).to.have.property('name');
+				expect(apiKeyResp?.name).to.be.a('string');
+				expect(apiKeyResp?.name).to.equal(`SampleDeviceKey${i}`);
+				expect(apiKeyResp).to.have.property('description');
+				expect(apiKeyResp?.description).to.be.a('string');
+				expect(apiKeyResp?.description).to.equal(
+					`Sample key ${i} description.`,
+				);
```
thgreasi, line 224:
```suggestion
				expect(apiKeyResp).to.have.property('name', `SampleDeviceKey${i}`);
				expect(apiKeyResp).to.have.property('description', `Sample key ${i} description.`);
```

thgreasi (commented):
LGTM 
Just update the tests to the shorter format and self-certify :+1:

src/features/api-keys/routes.ts
```diff
@@ -34,6 +34,8 @@ export const createGenericApiKey: RequestHandler = async (req, res) => {
 
 export const createDeviceApiKey: RequestHandler = async (req, res) => {
 	const deviceId = parseInt(req.params.deviceId, 10);
+	const { name, description }: { name: string; description: string } = req.body;
```
dfunckt, line 37: You will need to do some validation on `name` and `description` before passing it onwards.

nitishagar: @dfunckt What kind of validation are we looking for?

We do similar de-structuring here: https://github.com/balena-io/open-balena-api/pull/948/files#diff-e5d9db05ea54b94163d018aac63bfaaed210c4073f518e0550d97a0a72f2fb1eR102

nitishagar: Discussed here: https://www.flowdock.com/app/rulemotion/resin-tech/threads/spWwhMW27JpeCL5byTfh3CJ1N8p

src/features/api-keys/routes.ts
```diff
@@ -117,3 +123,20 @@ export const createNamedUserApiKey: RequestHandler = async (req, res) => {
 		res.status(500).send(translateError(err));
 	}
 };
+
+const getKeyMetadata = (reqBody: { name: any; description: any }) => {
+	const { name, description } = reqBody;
+
+	if (name != null && !(typeof name === 'string' || name instanceof String)) {
+		throw Error('Key name should be a string value');
```
dfunckt, line 131: The thrown error should be `BadRequest` so that the correct response code is sent.

dfunckt (commented):
Nice. A test for a request without name/description and the correct response code would be great.

src/features/api-keys/routes.ts
```diff
@@ -1,4 +1,5 @@
 import { errors } from '@balena/pinejs';
+import { BadRequestError } from '@balena/pinejs/out/sbvr-api/errors';
```
dfunckt, line 2: We shouldn't import from a module's nested exports.

You have access to `BadRequest` via `errors` and in fact it's already used in a couple of places in this file.

test/08_api-keys.ts
```diff
@@ -181,6 +191,67 @@ describe('create device apikey', function () {
...
+				expect(apiKeyResp).to.have.property(
+					'description',
+					`Sample key ${i} description.`,
+				);
+			});
+
+			it('should return BadRequest for an apikey create request where the key name is not string', async function () {
+				await fn(this.user, this.device.id, {
+					name: 123,
+				}).expect(400);
```
dfunckt, line 252: It's nice to check the error message as well, as there's been times tests have been passing for the wrong reasons.

```suggestion
				}).expect(400, 'Key name should be a string value');
```

src/features/api-keys/lib.ts
```diff
@@ -94,6 +94,25 @@ const $createApiKey = async (
 	return apiKey;
 };
 
+const getKeyMetadata = (reqBody: { name: any; description: any }) => {
+	const { name, description } = reqBody;
+
+	if (name != null && !(typeof name === 'string' || name instanceof String)) {
```
thgreasi, line 100: I would expect this v to be enough, given how the deserialization works, which is how we do it in all other places as well
```suggestion
	if (name != null && typeof name !== 'string') {
```
That applies for description as well probably.

src/features/api-keys/lib.ts
```diff
@@ -94,6 +94,25 @@ const $createApiKey = async (
 	return apiKey;
 };
 
+const getKeyMetadata = (reqBody: { name: any; description: any }) => {
```
thgreasi, line 97: NIT:
```suggestion
const getKeyMetadata = (reqBody: { name?: any; description?: any }) => {
```

src/features/api-keys/lib.ts
```diff
@@ -102,6 +121,15 @@ export const createApiKey = async (
...
 ): Promise<string> => {
 	options.apiKey ??= randomstring.generate();
+	const { name, description } = getKeyMetadata(req.body);
+
+	if (!options.name) {
+		options.name = name;
+	}
+	if (!options.description) {
+		options.description = description;
+	}
```
thgreasi, line 131: Do we intend to block setting empty stings?
if not then we should
```suggestion
	if (options.name != null) {
		options.name = name;
	}
	if (options.description != null) {
		options.description = description;
	}
```

src/features/api-keys/lib.ts
dfunckt: We've gone through this before -- `string?` is *very* problematic because it actually has three states: `null/undefined`, `''` and some non-zero length string but for most intents and purposes the first two states mean the same thing: "i have no value". So I think ensuring empty string do not make it to the database is sensible.

src/features/api-keys/lib.ts
nitishagar: Shouldn't we block `empty` strings?

dfunckt: We've gone through this before -- `string?` is *very* problematic because it actually has three states: `null/undefined`, `''` and some non-zero length string but for most intents and purposes the first two states mean the same thing: "i have no value". So I think ensuring empty string do not make it to the database is sensible.

thgreasi: Why should we?
Have we been doing this historically?

thgreasi: Well I see we have been blocking empty stings already.... so nevermind...

dfunckt: Arguably, the correct way to handle this is to introduce a check on the term/fact type that constrains at least the name to a length greater than 3-4 characters.

thgreasi (approved):

PR #951, opened 2022-02-10
Add database name env variable

### LINKED ISSUE #949: Databasename is hardcoded
The name of the database that openbalena-api uses is hardcoded as `resin` in this file: https://github.com/balena-io/open-balena-api/blob/master/config/confd/templates/env.tmpl

It would be nice to have this as an environment variable for use with a remote postgres database.

GoodOldJack12:
This allows the configuration of the database the api connects through with the `DB_NAME` environment variable. The default is `resin` if the variable is empty. 

closes #949


dfunckt (approved):
Thanks @GoodOldJack12!

PR #952, opened 2022-02-14
Removed systemd from container

robbe-haesendonck:
To enable running open-balena-api unprivileged systemd was removed

Change-type: minor


dfunckt:
Thanks for your PR. What is your use-case?

systemd is baked into the [base image](https://github.com/balena-io-modules/open-balena-base) but we're also maintaining a "no-systemd" line which you'll need to switch to.

dfunckt:
Absolutely, I agree, systemd in balena services is legacy really and there's no longer a reason to use it. I'm just curious what prompted you to get it removed.

robbe-haesendonck:
Well, we experienced a lot of issues internally with deploying open-balena in a proper way.
Removing systemd would allow us to run it on our internal infra much easier, so that's why I'm trying to get it upstream.
So we can keep following upstream without needing to create a fork and because I think more people can benefit from this.

dfunckt:
I see, thanks.

I've got to warn you this isn't going to be straightforward, but I'm happy to help as I can.

robbe-haesendonck:
Is there anything else that needs to be changed for this to be merged?
Thanks!

robbe-haesendonck:
Allright, so tests passed. I do think I messed up a commit message somewhere as that did not pass.

robbe-haesendonck:
Thanks for the help, got everything green :laughing:

dfunckt:
The API isn't using or cares about mDNS but openBalena (and some other internal use cases) does when deploying an instance under a `.local` TLD.

robbe-haesendonck:
@dfunckt Hello, do you have any update on the mDNS problem?

dfunckt:
@robbe-haesendonck I had a chat with the team and we're in agreement that in principle removing systemd from our services is desirable. We have a few ideas how to fill the functionality gap but no implementation yet, so please bare with us. I'll get back to this as we make progress on the requirements.

robbe-haesendonck:
@dfunckt Allright, thanks for the update!

robbe-haesendonck:
@dfunckt I created 2 seperate dockerfiles, one with and one without systemd.
Would this make it suitable to be merged?

PR #963, opened 2022-02-28
Add default name and description for provisioning key (and backfill the same)

nitishagar:
Change-Type: minor

See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/_5BN_IuhmYxXSfS8xDO06s_VZlN

### DIFF
--- a/test/05_device-config.ts
+++ b/test/05_device-config.ts
@@ -1,0 +1,95 @@
+import { expect } from './test-lib/chai';
+import * as fixtures from './test-lib/fixtures';
+import { supertest } from './test-lib/supertest';
+import { version } from './test-lib/versions';
+import { sbvrUtils, permissions } from '@balena/pinejs';
+
+const { api } = sbvrUtils;
+
+describe('generate device config', function () {
+	before(async function () {
+		const fx = await fixtures.load('05-device-config');
+		this.loadedFixtures = fx;
+		this.user = fx.users.admin;
+		this.application = fx.applications.app1;
+	});
+
+	after(async function () {
+		await supertest(this.user).delete(`/${version}/api_key`).expect(200);
+		await fixtures.clean(this.loadedFixtures);
+	});
+
+	describe('using /download-config', function () {
+		it('should be able to create a provisioning key with default name and description', async function () {
+			const { body } = await supertest(this.user)
+				.get(
+					`/download-config?appId=${this.application.id}&version=v2.24.0.dev&deviceType=raspberrypi3`,
+				)
+				.expect(200)
+				.expect('content-type', /^application\/json/);
+
+			expect(body).to.have.a.property('apiKey');
+			expect(body.apiKey).to.be.a('string');
+
+			// check the name assigned
+			const apiKeyResp = await api.resin.get({
+				resource: 'api_key',
+				passthrough: {
+					req: permissions.root,
+				},
+				id: {
+					key: body.apiKey,
+				},
+				options: {
+					$select: ['name', 'description'],
+				},
+			});
+
+			expect(apiKeyResp).to.have.property(
+				'name',
+				'Automatically generated provisioning key',
+			);
+			expect(apiKeyResp).to.have.property(
+				'description',
+				'Automatically generated for an image download or config file generation',
+			);
+		});
+
+		it('should be able to create a provisioning key with specified name and description', async function () {
+			const { body } = await supertest(this.user)
+				.post('/download-config')
+				.send({
+					appId: this.application.id,
+					version: 'v2.24.0',
+					deviceType: 'raspberrypi3',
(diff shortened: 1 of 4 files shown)


thgreasi (commented):
Can we also have a test case?

src/migrations/00071-default-provisioning-key-metadata.sql
```diff
@@ -0,0 +1,21 @@
...
+UPDATE "api key"
+SET "description" = u."Automatically generated for provisioning"
+FROM "api key-has-role"
+WHERE "api key"."description" IS NULL
+AND "api key"."id" = "api key-has-role"."api key"
+AND "api key-has-role"."role" IN (
+	SELECT "id"
+	FROM "role"
+	WHERE "name" = 'provisioning-api-key'
+);
```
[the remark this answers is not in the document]
nitishagar: @thgreasi We are setting in a generic description, and this is set only for `provisioning` keys. If we don't have it set, then the table which will come up in dashboard will have a bunch of keys with empty name and description showing up the first time.

cc. @chrisys

nitishagar:
@thgreasi I updated the PR like our last time to generate if for all keys going forward rather than cover it via `/download` or `/download-config`. This will help us populate an auto gen name|description when one is not provided.

src/features/api-keys/lib.ts
```diff
@@ -121,10 +153,15 @@ export const createApiKey = async (
 	const { name, description } = getKeyMetadata(req.body);
 
 	if (!options.name) {
-		options.name = name;
+		options.name = name
+			? name
+			: `Automatically generated ${getKeyType(roleName)} key`;
```
thgreasi, line 158: This function is the basis for all api key generation endpoints as well (like `/api-key/application/:appId/provisioning`), which is also what the SDK methods call. An explicit call to those endpoints doesn't like it should mark the api keys as "Automatically generated", since there was code outside of the API that explicitly called it for a reason that we can't be sure.

Afaik we agreed to only add this `Automatically generated` placeholder only for API keys generated using `/download` & `/download-config`. Should we switch to an extra argument for defining the placeholder text rather than just on the `roleName`? (Or just an extra optional property(/ies) for the ApiKeyOptions, which though should be destructed and not be propagated into `$createApiKey()`.
PS: Sounds like the whole `KeyMetadata` & `getKeyType` code could just be replaced with a
`{ [roneName: string]: string }` dictionary.

thgreasi: Also wasn't the target to use the following text v?
> Name: Automatically generated
Description: 'Automatically generated for an image download' | 'Automatically generated during a configuration file generation'

nitishagar: I am reverting back to `provisioning` only as agreed before. I couldn't get your point here:
> An explicit call to those endpoints doesn't like it should mark the api keys as "Automatically generated", since there was code outside of the API that explicitly called it for a reason that we can't be sure.

src/features/api-keys/lib.ts
```diff
@@ -7,6 +7,25 @@ import { sbvrUtils, permissions, errors } from '@balena/pinejs';
 const { api } = sbvrUtils;
 const { BadRequestError } = errors;
 
+export class KeyMetadata {
```
thgreasi, line 10: Let's revert all changes in this file since they are not required.

src/features/api-keys/lib.ts
```diff
@@ -158,17 +178,17 @@ export type PartialCreateKey = (
...
 	createApiKey,
 	'device',
-	'device-api-key',
+	KeyMetadata.DEVICE.role,
 );
 export const createNamedUserApiKey: PartialCreateKey = _.partial(
 	createApiKey,
 	'user',
-	'named-user-api-key',
+	KeyMetadata.USER.role,
```
thgreasi, line 191: Let's revert all changes in this file since they are not required.
PS: An enum would have been enough for this.

test/05_device-config.ts
```diff
@@ -0,0 +1,58 @@
...
+		this.application = fx.applications.app1;
+	});
+
+	after(async function () {
+		await supertest(this.user).delete(`/${version}/api_key`).expect(200);
+		await fixtures.clean(this.loadedFixtures);
+	});
+
+	describe('using /download-config', function () {
+		it('should be able to create a provisioning key with default name and description', async function () {
```
thgreasi, line 23: How about adding one more for user provided values so that we have a complete scenario?

src/migrations/00071-default-provisioning-key-metadata.sql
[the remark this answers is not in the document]
nitishagar: This is removed.

thgreasi (approved):

PR #966, opened 2022-03-01
Ensure at least one device is being updated for batch vpn updates

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

Page-:
@balena-ci republish

PR #970, opened 2022-03-02
Enable registry auth requests to resolve image aliases

klutchell:
We can leverage the existing API registry auth features
to resolve org/app/release requests to the v2/randomhash format
and return the correct permissions and path in the JWT.

Change-type: patch
Signed-off-by: Kyle Harding <kyle@balena.io>

`balena-io architecture` brainstorm topic discussion.
(restricted access) https://jel.ly.fish/ab17fa5d-8aeb-4279-9e6a-a1f7b03f7009

Downstream balena-api tests:
(restricted access) https://github.com/balena-io/balena-api/pull/3580


src/features/registry/registry.ts
```diff
@@ -207,13 +216,124 @@ const resolveImageId = multiCacheMemoizee(
...
+		service: string | undefined,
+		tx: Tx,
+	): Promise<string | undefined> => {
+		try {
+			const [image] = await api.resin.get({
+				resource: 'image',
+				passthrough: { req: permissions.root, tx },
+				options: {
+					$top: 1,
+					$select: ['is_stored_at__image_location'],
```
thgreasi, line 232: NIT:
```suggestion
					$select: 'is_stored_at__image_location',
```

src/features/registry/registry.ts
```diff
@@ -207,13 +216,124 @@ const resolveImageId = multiCacheMemoizee(
...
+							$any: {
+								$alias: 'ri',
+								$expr: {
+									ri: {
+										is_part_of__release: {
+											$any: {
+												$alias: 'ipor',
+												$expr: {
+													ipor: {
+														status: 'success' as const,
```
thgreasi, line 244: you don't need this here since the API's pine clients doesn't type check model properties :)
```suggestion
														status: 'success',
```

src/features/registry/registry.ts
```diff
@@ -207,13 +216,124 @@ const resolveImageId = multiCacheMemoizee(
 	},
 );
 
+const resolveImageLocation = multiCacheMemoizee(
+	async (
+		application: string,
+		release: string | undefined,
+		service: string | undefined,
```
thgreasi, line 223: How about v so that's clear what each param holds?
```suggestion
		applicationSlug: string,
		releaseHashOrVersion: string | undefined,
		serviceName: string | undefined,
```

src/features/registry/registry.ts
```diff
@@ -207,13 +216,124 @@ const resolveImageId = multiCacheMemoizee(
...
+															{
+																ipor: {
+																	semver: release,
+																	is_final: true,
+																},
+															},
+															{
+																ipor: {
+																	raw_version: release,
+																	is_final: false,
```
thgreasi, line 280: We definitely could use a comment here to explain why we query those two different fields on each case.

src/features/registry/registry.ts
```diff
@@ -297,37 +422,79 @@ const authorizeRequest = async (
...
+				}
+
+				const imageLocation = await resolveImageLocation(
+					application,
+					release,
+					service,
+					tx,
+				);
+
+				if (imageLocation != null) {
```
thgreasi, line 469: Shouldn't it be like: 
```suggestion
				if (imageLocation == null) {
					return {
						type,
						name,
						actions: [],
					};
				}
```
Does it make to proceed with the `// request for legacy public-read appName/commit image` default case?

thgreasi (commented):
Would be nice if we also had a test case for this :)
Atm unfortunately the tests are in the cloud api :( `balena-api/test/12_registry_authentication.ts:`
Maybe we could move/copy a subset of them here, but I'm fine even with adding the test in the cloud tests.
Given how hot this endpoint is, I would also like if Page took a look as well.

src/features/registry/registry.ts
```diff
@@ -34,10 +34,17 @@ const SUPERVISOR_REPOSITORIES = /^resin\/(?:[a-zA-Z0-9]+-)+supervisor$/;
 
 const NEW_REGISTRY_REGEX = /(^(\d+)\/[\d\-]+$|^(v2\/[a-z0-9]+)(-[0-9]+)?)/;
 
+// match <org>/<app>/<semver|commit>/<service> where <semver|commit> and <service> are optional
```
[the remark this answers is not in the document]
klutchell: No, we can't use the tag field (after the colon) for redirecting to different registries since each release is a unique registry with unique permissions but the docker client doesn't request the tag or digest until after authentication has taken place.

So whatever is included as a tag is ignored (and isn't available as part of an auth scope anyway).

src/features/registry/registry.ts
klutchell: sounds good!

klutchell:
> Would be nice if we also had a test case for this :) Atm unfortunately the tests are in the cloud api :( `balena-api/test/12_registry_authentication.ts:` Maybe we could move/copy a subset of them here, but I'm fine even with adding the test in the cloud tests. Given how hot this endpoint is, I would also like if Page took a look as well.

Thanks for pointing out those tests, I was having trouble finding them. I'm happy to add new test cases to balena-api unless you think they should be moved.

src/features/registry/registry.ts
klutchell: From the tests that old format only applies to registry v1, which we don't have in balenaCloud afaik but maybe still exists in OB?

klutchell: Good catch, the registry v1 tests aren't passing because
`test_acl1/d6cd1e2bd19e03a81132a23b2025920577f84e37`
matches the same regex as
`<org>/<app>/<?version>`
so I need to figure out if that legacy path has any valid use case these days

klutchell: Keeping this change as additive for now, so if we can't resolve an image path we will fall back to the legacy `appname/commit` logic (the regex matches both unfortunately).

klutchell:
Downstream balena-api tests:
(restricted access) https://github.com/balena-io/balena-api/pull/3580

thgreasi (commented):
LGTM :heart_eyes:

klutchell:
@balena-ci I self-certify!

PR #971, opened 2022-03-03
Drop support for legacy reporting single vpn client connect/disconnects

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #972, opened 2022-03-03
Allow specifying redis hosts as a single `${host}:${port}` var

Page-:
Rather than having to specify the host/port separately

Change-type: patch


Page-:
@balena-ci I self-certify!

Page-:
@balena-ci republish

PR #974, opened 2022-03-04
Also support a single `${host}:${port}` var for `REDIS_LOGS_RO_HOST`

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #975, opened 2022-03-04
Only accept redis host vars in `${host}:${port}` pairs

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #976, opened 2022-03-08
Ensure the device logs buffer cannot be larger than the retention limit

Page-:
Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/store.ts
+++ b/src/features/device-logs/lib/store.ts
@@ -193,6 +193,11 @@ function handleStreamingWrite(
 		try {
 			// Don't flush if the backend is reporting as unavailable
 			if (buffer.length && backend.available) {
+				const limit = ctx.retention_limit;
+				if (buffer.length > limit) {
+					// Ensure the buffer cannot be larger than the retention limit
+					buffer.splice(0, buffer.length - limit);
+				}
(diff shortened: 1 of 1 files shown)


src/features/device-logs/lib/store.ts
```diff
@@ -193,6 +193,11 @@ function handleStreamingWrite(
 		try {
 			// Don't flush if the backend is reporting as unavailable
 			if (buffer.length && backend.available) {
+				const limit = ctx.retention_limit;
```
fisehara, line 196: Why is this referenced into a const here?

src/features/device-logs/lib/store.ts
```diff
@@ -193,6 +193,11 @@ function handleStreamingWrite(
...
+				if (buffer.length > limit) {
+					// Ensure the buffer cannot be larger than the retention limit
+					buffer.splice(0, buffer.length - limit);
```
fisehara, line 199: so this will only send the latest `limit-many` logs. Do we don't care about the older logs as the buffer.length will be set to 0 after publish.

src/features/device-logs/lib/store.ts
Page-: I use it twice so it was just for having a shorter reference

fisehara (approved):

PR #980, opened 2022-03-11
Use the existing redis instance for the multi-level-store

Page-:
Change-type: patch


Page-:
@balena-ci rebase

Page-:
This doesn't work because cache-manager-ioredis expects the store args to be present on the passed in redis instance which isn't feasible with a shared instance

PR #983, opened 2022-03-15
Fix random name generator

chrisys:
Adjectives after 'mean' were not being selected.

Change-type: patch
Signed-off-by: Chris Crocker-White <chriscw@balena.io>


chrisys:
@balena-ci I self-certify!

PR #984, opened 2022-03-18
Move to flowzone and utilise docker-compose to control test dependency

fisehara:
Change-type: minor


fisehara:
When this is a good approach and can be used I'm going to move it downwards to https://github.com/balena-io/pinejs.

PR #986, opened 2022-03-21
Add support for using a redis cluster for device logs

Page-:
Change-type: minor


Page-:
The redis hash tag part was previously reviewed as part of #925

Page-:
@balena-ci retest

src/features/device-heartbeat/index.ts
```diff
@@ -189,10 +189,9 @@ export class DeviceOnlineStateManager extends events.EventEmitter {
 			return;
 		}
 
-		// initialise the RedisSMQ object using our Redis client...
 		this.rsmq = new RedisSMQ({
-			host: REDIS.general.host,
-			port: REDIS.general.port,
+			// TODO: This cannot currently support a redis cluster
```
fisehara, line 193: What is 'this' refering to? logging device-heartbet to a cluster?

src/lib/config.ts
```diff
@@ -201,62 +201,95 @@ export const RATE_LIMIT_MEMORY_BACKEND = optionalVar(
...
+	if (hosts.length > 1) {
+		throw new Error(
+			`'${hostVarName}' must contain only one entry when not in cluster mode`,
+		);
 	}
-	const [host, maybePort] = hostPair.split(':');
-	const port = checkInt(maybePort);
-	if (port == null) {
-		throw new Error(`Invalid port for '${varName}': ${maybePort}`);
+	if (hosts.length > 1) {
```
fisehara, line 272:
``` suggestions
if (roHosts.length > 1) {
```

should be roHosts, otherwise it's caught always in the if before.

src/features/device-logs/lib/backends/redis.ts
```diff
@@ -155,7 +155,7 @@ export class RedisBackend implements DeviceLogsBackend {
 	}
 
 	private getKey(ctx: LogContext, suffix = 'logs') {
-		return `device:${ctx.id}:${suffix}`;
+		return `{device:${ctx.id}}:${suffix}`;
```
fisehara, line 158: Just for the reference why this is beneficial: 
https://redis.io/docs/reference/cluster-spec/#hash-tags

fisehara (changes requested):
LGTM.

Only question left, for the postgres connection env vars host and port have been separated. Should this also be considered for redis?

src/features/device-heartbeat/index.ts
Page-: RSMQ, basically we cannot initialize this with a cluster as it will fail

Page-:
@fisehara they used to be separate but I intentionally merged them so that it would be feasible to specify multiple nodes, eg
```
REDIS_HOST=10.0.0.2:6379,10.0.0.3:6380,10.0.0.4:6381
```
rather than something like
```
REDIS_HOST=10.0.0.2,10.0.0.3,10.0.0.4
REDIS_PORT=6379,6380,6381
```
or
```
REDIS_HOST_1=10.0.0.2
REDIS_PORT_1=6379
REDIS_HOST_2=10.0.0.3
REDIS_PORT_2=6380
REDIS_HOST_3=10.0.0.4
REDIS_PORT_3=6381

PR #987, opened 2022-03-22
Remove unused redis port vars

Page-:
They're now specified as `${host}:${port}` pairs in the redis host vars

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #988, opened 2022-03-22
CircleCI: Use small instances (we only need 1 core)

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1003, opened 2022-04-06
Update dependencies

Page-:
Update @balena/node-metrics-gatherer from 6.0.0 to 6.0.1

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1008, opened 2022-04-18
Consider devices with heartbeats to be out of the pre-provisioning state

thgreasi:
Change-type: minor
See: https://www.flowdock.com/app/rulemotion/r-resinos/threads/3cl3JBefZ2rV51CrrGpOz5T478d
See: https://jel.ly.fish/pattern-heartbeating-devices-without-vpn-connectivity-show-configuring-when-turned-off-f19c956

### DIFF
--- a/src/features/devices/models/device-additions.ts
+++ b/src/features/devices/models/device-additions.ts
@@ -59,8 +59,13 @@ export const addToModel = (
 	// so it's still provisioning.
 	const isPreProvisioning: AndNode = [
 		'And',
-		isOverallOffline,
+		['Equals', ['ReferencedField', 'device', 'is online'], ['Boolean', false]],
 		['NotExists', ['ReferencedField', 'device', 'last connectivity event']],
+		[
+			'Equals',
+			['ReferencedField', 'device', 'api heartbeat state'],
+			['EmbeddedText', 'unknown'],
+		],
 	];
 
 	const isPostProvisioning: EqualsNode = [


src/features/devices/models/device-additions.ts
```diff
@@ -59,8 +59,13 @@ export const addToModel = (
...
 	const isPreProvisioning: AndNode = [
 		'And',
-		isOverallOffline,
+		['Equals', ['ReferencedField', 'device', 'is online'], ['Boolean', false]],
 		['NotExists', ['ReferencedField', 'device', 'last connectivity event']],
+		[
+			'Equals',
+			['ReferencedField', 'device', 'api heartbeat state'],
+			['EmbeddedText', 'unknown'],
+		],
```
thgreasi, line 68: This change really just does whatever the comment above says...
The reasoning being that, in order for any device to have any heartbeat status != unknown, it means that the supervisor has managed to reach the API at least once, which means that the device is not really any more in pre-provisioning state.

Initial PR that introduces the previous logic
See: https://github.com/balena-io/balena-api/pull/2657/files

src/features/devices/models/device-additions.ts
```diff
@@ -59,8 +59,13 @@ export const addToModel = (
 	// so it's still provisioning.
 	const isPreProvisioning: AndNode = [
 		'And',
-		isOverallOffline,
+		['Equals', ['ReferencedField', 'device', 'is online'], ['Boolean', false]],
```
Page-, line 62: Is this redundant since having any connectivity event info would be the real tell?

thgreasi: It should probably be fine without it. I only really added it for dependent devices.

thgreasi:
@balena-ci I self-certify!

PR #1009, opened 2022-04-19
Only cache the device id for `checkDeviceExists` rather than an object

Page-:
The object is unnecessary as we only care about the id and only serves
to increases the cache memory usage

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1010, opened 2022-04-20
Add `SUPERVISOR_REPORT_INTERVAL` config var

20k-ultra:
Allows controlling how frequently a device reports its current state

Change-type: minor
Connects-to: https://github.com/balena-os/balena-supervisor/pull/1910
Signed-off-by: 20k-ultra <3946250+20k-ultra@users.noreply.github.com>


src/features/vars-schema/env-vars.ts
```diff
@@ -73,6 +77,14 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
 		description: 'Enable / Disable logs from being sent to balena API',
 		default: 'true',
 	},
+	RESIN_SUPERVISOR_REPORT_INTERVAL: {
+		type: 'integer',
+		description:
+			'Define how frequently the device reports its current state to the cloud in milliseconds',
+		default: DEFAULT_SUPERVISOR_REPORT_INTERVAL,
+		minimum: DEFAULT_SUPERVISOR_REPORT_INTERVAL,
+		maximum: 86400000,
```
20k-ultra, line 86: this line could be changed to prevent large report intervals...maybe 24 hours ?

src/features/vars-schema/env-vars.ts
```diff
@@ -73,6 +77,14 @@ export const SUPERVISOR_CONFIG_VAR_PROPERTIES: {
 		description: 'Enable / Disable logs from being sent to balena API',
 		default: 'true',
 	},
+	RESIN_SUPERVISOR_REPORT_INTERVAL: {
+		type: 'integer',
+		description:
+			'Define how frequently the device reports its current state to the cloud in milliseconds',
```
thgreasi, line 83: Shouldn't we also add the classic `Supported by supervisor versions >= v13.?.?.` at the end like we do in other places for new config vars?

src/lib/config.ts
```diff
@@ -339,6 +339,11 @@ export const DEFAULT_SUPERVISOR_POLL_INTERVAL = intVar(
 	10 * MINUTES,
 );
 
+export const DEFAULT_SUPERVISOR_REPORT_INTERVAL = intVar(
```
thgreasi, line 342: Having this as an injected env var doesn't offer much, unless you also plan to enforce this value in all state-GETs like we do for the min poll interval.
See: https://github.com/balena-io/open-balena-api/blob/fd8899bf269c998cb182bfcc8d5600b1a5ea1780/src/features/device-state/state-get-utils.ts#L6

src/features/vars-schema/env-vars.ts
thgreasi:
```suggestion
		maximum: 24 * HOURS, // import HOURS from src/lib/config.ts
```

20k-ultra:
I'm just waiting for the underlying feature that will use this to be confirmed since there seems to be some discussion about the necessity to add this feature. We might add a different config var first.

20k-ultra:
We will be prioritizing a `SUPERVISOR_METRICS_INTERVAL`

PR #1013, opened 2022-04-26
Use a shared cache for the device type json

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1018, opened 2022-04-29
Fix early exist check on service install creation from application PATCH

thgreasi:
TS was unfortunately happy with this b/c FilterObject is a `Dictionary<FilterObj | FilterArray | FilterBaseType>`, so `.length` is a valid prop and comparing with numbers is a valid value as part of FilterBaseType...
Thankfully the check wasn't `!deviceFilterOrIds.length`, b/c that would have resulted invalid target states for devices.

Change-type: patch


thgreasi:
@balena-ci I self-certify!

PR #1019, opened 2022-04-29
Optimize missing service-install detection for large fleets

thgreasi:
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/RC-voxJYGqxr_xRmyxHRtm9pLdw
See: https://www.flowdock.com/app/rulemotion/platform/threads/NZYiO9WfkuJ6JlXi_7vcYr2Qngq

### DIFF
--- a/src/features/ci-cd/hooks/service-installs.ts
+++ b/src/features/ci-cd/hooks/service-installs.ts
@@ -13,18 +13,7 @@ const createReleaseServiceInstalls = async (
 	deviceFilterOrIds: number[] | FilterObj,
 	releaseFilter: Filter,
 ): Promise<void> => {
-	const deviceIds = Array.isArray(deviceFilterOrIds)
-		? deviceFilterOrIds
-		: (
-				(await api.get({
-					resource: 'device',
-					options: {
-						$select: 'id',
-						$filter: deviceFilterOrIds,
-					},
-				})) as Array<Pick<Device, 'id'>>
-		  ).map(({ id }) => id);
-	if (deviceIds.length === 0) {
+	if (Array.isArray(deviceFilterOrIds) && deviceFilterOrIds.length === 0) {
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -65,35 +54,54 @@ const createReleaseServiceInstalls = async (
...
+						si: {
+							installs__service: serviceId,
+						},
+					},
+				},
+			},
+		},
+	}));
+
+	const devicesToAddServiceInstalls = (await api.get({
```
thgreasi, line 72: For the case that a fleet changes its target release while the two releases share the same services, this previously was returning `D * S` rows, while now it would return 0.
As a result that also reduces the `O(D * S)` complexity from the `.map`s & `_.difference` of the code running in the `Promise.all` (which in the above scenario would have resulted 0 INSERTs, since everything would match).

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -65,35 +54,54 @@ const createReleaseServiceInstalls = async (
...
+					$alias: 'si',
+					$expr: {
+						si: {
+							installs__service: serviceId,
+						},
+					},
+				},
+			},
+		},
+	}));
```
thgreasi, line 70: The tests seem to be able to run the query even when using 200 services ids (in which case the ORs and the permissions make it really big).

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -13,18 +13,7 @@ const createReleaseServiceInstalls = async (
 	deviceFilterOrIds: number[] | FilterObj,
 	releaseFilter: Filter,
 ): Promise<void> => {
-	const deviceIds = Array.isArray(deviceFilterOrIds)
-		? deviceFilterOrIds
-		: (
-				(await api.get({
```
thgreasi, line 19: ~130ms for huge fleets (while ignoring locks)

Page- (approved):

PR #1021, opened 2022-05-04
Invalidate the device type JSONs cache when a new hostApp is created

thgreasi:
This should fix the BoB test cases.

Change-type: patch


thgreasi:
@balena-ci retest

PR #1055, opened 2022-06-21
Add index to optimize hostApp app rule

thgreasi:
This should improve the "device should be operated by hostApp release of the same DT" rule.
It already is using an index on "operated by release" but since that's the most costly part of the plan, it should help making it an index only scan.
See: https://explain.dalibo.com/plan/1YE#raw (current plan)
PS: Moreover it will be consistent with the "managed by release" index as well :)

Change-type: patch


Page- (approved):

thgreasi:
Closing for now since we found that it actually made performance worse...

PR #1062, opened 2022-07-15
Remove an unnecessary `$select`ed field

Page-:
Change-type: patch


thgreasi (approved):

thgreasi (commented):
We do need the device IDs in both cases afaict, so that we know which devices to set the SV/hostAPP release to.
I agree though that we don't need the DT id.

PR #1064, opened 2022-07-15
Explicitly enable compression for ndjson

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1066, opened 2022-07-19
Change ip address and mac address data type to TEXT

fisehara:
Change-type: minor
See: https://jel.ly.fish/brainstorm-topic-4bf51224-cc3a-42de-8b68-468d3feecccf
Co-authored-by: Rose <rommims@gmail.com>
Co-authored-by: Christina Wang <cywang117@users.noreply.github.com>

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -149,7 +149,7 @@ Term: invalidation reason
 	Concept Type: Text (Type)
 
 Term: ip address
-	Concept Type: Short Text (Type)
+	Concept Type: Text (Type)
 
 Term: known issue list
 	Concept Type: Text (Type)
@@ -176,7 +176,7 @@ Term: logs channel
 	Concept Type: Short Text (Type)
 
 Term: mac address
-	Concept Type: Short Text (Type)
+	Concept Type: Text (Type)
 
 Term: memory usage
 	Concept Type: Integer (Type)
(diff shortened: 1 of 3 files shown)


cywang117:
Awaiting on-device BoB tests which will be done shortly

thgreasi:
@fisehara was this discussed in a call that you can link the PR to?
Will that lock the table?
Did you try applying this change and running an older instance against the new DB schema to confirm that it still works? (I expect it would, just asking, except maybe if you start setting longer values and rollback after that.)

fisehara:
@thgreasi I'm looping in @cywang117 @zoobot if they know the call where this was discussed. I just helped for the pairprogramming session, thus commited the changes.

cywang117:
@thgreasi Added "See: https://jel.ly.fish/brainstorm-topic-4bf51224-cc3a-42de-8b68-468d3feecccf" to initial PR. I'll link it in the commit later today.

From [Postgres 14 docs](https://www.postgresql.org/docs/current/sql-altertable.html), an ACCESS EXCLUSIVE lock will be acquired with this migration. As VARCHAR(255) can be directly converted to TEXT easily, we theorized that the lock would not be taken for a long time, however this remains to be tested. Would it be possible to deploy to staging first to monitor the migration performance? Alternatively we could migrate a seeded device table in a test environment and monitor the migration there?

cywang117:
@Page- suggested that the Supervisor always send the full ip/mac address list, but the API truncate it at the Supervisor PATCH hanlder before inserting into the database.

As a result we can keep the data type at VARCHAR(255) because we have not encountered the issue of ip/mac addresses being too long with any regularity, so we can close this PR as the database does not need to migrate if we go with this solution.

Note that users who wish to get a complete list of untruncated IP/MAC addresses can query the Supervisor API's /v1/device endpoint, either directly or proxying through the balena API.

PR #1068, opened 2022-07-21
Ip address and mac address resilient truncation

fisehara:
truncate to resolve failing device state patch from older supervisors

Change-type: minor
Co-authored-by: Rose <rose.meyers@balena.io>
Co-authored-by: Christina Wang <christina@balena.io>

### DIFF
--- a/src/features/device-state/routes/state-patch-v2.ts
+++ b/src/features/device-state/routes/state-patch-v2.ts
@@ -19,6 +19,7 @@ import {
 	v2ValidPatchFields,
 	upsertImageInstall,
 	deleteOldImageInstalls,
+	truncateShortTextFields,
 } from '../state-patch-utils';
 import type { ResolveDeviceInfoCustomObject } from '../middleware';
 
@@ -268,6 +269,13 @@ export const statePatchV2: RequestHandler = async (req, res) => {
 
 				updateFns.push(async (resinApiTx) => {
 					if (Object.keys(deviceBody).length > 0) {
+						// truncate for resilient legacy compatible device state patch so that supervisors don't fail
+						// to update b/c of length violation of 255 (SBVR SHORT TEXT type) for ip and mac address.
+						// sbvr-types does not export SHORT TEXT VARCHAR length 255 to import.
+						deviceBody = truncateShortTextFields(deviceBody, [
+							'ip_address',
+							'mac_address',
+						]);
 						// If we're updating anyway then ensure the metrics data is included
 						deviceBody = { ...deviceBody, ...metricsBody };
 						await resinApiTx.patch({
--- a/src/features/device-state/state-patch-utils.ts
+++ b/src/features/device-state/state-patch-utils.ts
@@ -39,6 +39,46 @@ export const v2ValidPatchFields: Array<
 	'download_progress',
 ];
 
+const SHORT_TEXT_LENGTH = 255;
+const ADDRESS_DELIMITER = ' ';
+
+// Truncate text at delimiters to input length or less
+const truncateText = (
+	longText: string,
+	length: number = SHORT_TEXT_LENGTH,
+	delimiter: string = ADDRESS_DELIMITER,
+): string => {
+	return longText
+		.split(delimiter)
+		.reduce((text, fragment) => {
+			const textWithFragment = text + delimiter + fragment;
+			return textWithFragment.length <= length ? textWithFragment : text;
+		}, '')
+		.trim();
+};
+
+type ValidPatchFields = Array<
+	typeof v3ValidPatchFields[number] | typeof v2ValidPatchFields[number]
+>;
+
+export const truncateShortTextFields = (
+	object: Dictionary<any>,
+	keysToTruncate: ValidPatchFields,
+) => {
+	keysToTruncate.forEach((key) => {
+		if (
+			typeof object[key] !== 'string' ||
+			object[key].length <= SHORT_TEXT_LENGTH
+		) {
+			return;
+		}
+		if (['ip_address', 'mac_address'].includes(key)) {
+			object[key] = truncateText(object[key]);
+		}
+	});
+	return object;
+};
+
 export const metricsPatchFields = [
 	'memory_usage',
 	'memory_total',
--- a/src/features/device-state/routes/state-patch-v3.ts
+++ b/src/features/device-state/routes/state-patch-v3.ts
@@ -21,6 +21,7 @@ import {
 	deleteOldImageInstalls,
 	upsertImageInstall,
 	shouldUpdateMetrics,
+	truncateShortTextFields,
 } from '../state-patch-utils';
 
 const { BadRequestError, UnauthorizedError, InternalRequestError } = errors;
@@ -311,6 +312,13 @@ export const statePatchV3: RequestHandler = async (req, res) => {
 			}
 
 			if (Object.keys(deviceBody).length > 0) {
+				// truncate for resilient legacy compatible device state patch so that supervisors don't fail
+				// to update b/c of length violation of 255 (SBVR SHORT TEXT type) for ip and mac address.
+				// sbvr-types does not export SHORT TEXT VARCHAR length 255 to import.
+				deviceBody = truncateShortTextFields(deviceBody, [
+					'ip_address',
+					'mac_address',
+				]);
 				// If we're updating anyway then ensure the metrics data is included
 				deviceBody = { ...deviceBody, ...metricsBody };
 				updateFns.push(async (resinApiTx) => {
--- a/test/03_device-state.ts
+++ b/test/03_device-state.ts
@@ -481,6 +481,50 @@ mockery.registerMock('../src/lib/config', configMock);
 			);
 		});
 
+		it('should accept addresses longer than 255 chars and truncate at space delimiters', async () => {
+			const generateValidAddress = (
+				addr: string,
+				truncLen: number,
+				delimiter: string = '',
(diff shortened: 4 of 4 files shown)


src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -268,6 +269,16 @@ export const statePatchV2: RequestHandler = async (req, res) => {
...
 				updateFns.push(async (resinApiTx) => {
 					if (Object.keys(deviceBody).length > 0) {
+						// truncate for resilient legacy compatible device state patch so that supervisors don't fail
+						// to update b/c of length violation of 255 (SBVR SHORT TEXT type) for ip and mac address.
+						// sbvr-types does not export SHORT TEXT VARCHAR length 255 to import.
+						if (deviceBody.ip_address) {
+							deviceBody.ip_address = deviceBody.ip_address?.slice(0, macIpAddressesTruncationLength);
+						}
+						if (deviceBody.mac_address) {
+							deviceBody.mac_address = deviceBody.mac_address?.slice(0, macIpAddressesTruncationLength);
```
cywang117, line 279: Instead of `macIpAddressesTruncationLength`, should slice at the last index of the space character smaller than the truncation length, to prevent partial mac/ip addresses from being stored

zoobot: could also split and join

src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -268,6 +269,22 @@ export const statePatchV2: RequestHandler = async (req, res) => {
 
 				updateFns.push(async (resinApiTx) => {
 					if (Object.keys(deviceBody).length > 0) {
+						// truncate for resilient legacy compatible device state patch so that supervisors don't fail
+						// to update b/c of length violation of 255 (SBVR SHORT TEXT type) for ip and mac address.
+						// sbvr-types does not export SHORT TEXT VARCHAR length 255 to import.
+						if (deviceBody.ip_address) {
+							deviceBody.ip_address = deviceBody.ip_address?.slice(
```
thgreasi, line 276: since you are inside the `if`, I expect that the `?.` is not needed.
```suggestion
							deviceBody.ip_address = deviceBody.ip_address.slice(
```

src/features/device-state/state-patch-utils.ts
```diff
@@ -39,6 +39,8 @@ export const v2ValidPatchFields: Array<
 	'download_progress',
 ];
 
+export const macIpAddressesTruncationLength = 255;
```
thgreasi, line 42: How about exporting a function rather than the length?
It could also be doing a `['ip_address', 'mac_address'].forEach(...)`.
Or it could be made even more generic like `truncateShortTextFields<T extends object>(object: T, keys: Array<keyof T>)`, which could allow us to use it in more place (eg maybe for metrics).

cywang117: Made slightly more generic with https://github.com/balena-io/open-balena-api/commit/dd736ce43d85bc640f9b08a479ed783f0f56ed9d

src/features/device-state/routes/state-patch-v2.ts
cywang117: Updated logic to account for this with https://github.com/balena-io/open-balena-api/commit/dd736ce43d85bc640f9b08a479ed783f0f56ed9d

cywang117:
I'm not able to run `fasttest` locally and haven't been able to figure out why, so kindly requesting a re-review from @thgreasi @fisehara @zoobot. Thanks!

src/features/device-state/state-patch-utils.ts
```diff
@@ -39,6 +39,44 @@ export const v2ValidPatchFields: Array<
...
 
+const SHORT_TEXT_LENGTH = 255;
+const ADDRESS_DELIMITER = ' ';
+
+// Truncate space-delimited addresses at the space character to SHORT_TEXT_LENGTH
+const truncateIpMacAddresses = (address: string): string => {
+	return address
+		.split(ADDRESS_DELIMITER)
+		.reduce((addresses, addr) => {
+			return addresses.length + ADDRESS_DELIMITER.length + addr.length <
```
pranavpeshwe, line 50: Can we not use the full SHORT_TEXT_LENGTH here? IOW, why `<` and not `<=` ?

I'd worked on the support ticket that led to creation of this issue hence checking out the fix. Hope I'm not missing something obvious here.

fisehara: No specific reason to not use the full length of SHORT_TEXT_LENGTH

fisehara:
@cywang117 actually when the GH CI checks don't complain the fasttests shouldn't complain either. The tests are executed in the CI.

LGTM. Looks great, thanks for taking over.

src/features/device-state/state-patch-utils.ts
cywang117: If the current address string length is equal to (`<=`) SHORT_TEXT_LENGTH, it would add the next address in (`addresses + ADDRESS_DELIMITER + addr`), making the total length greater than 255. 

Also, if we do something like `return address.length < SHORT_TEXT_LENGTH ? addresses + DELIMITER + addr` (this is what I assume is meant by using the full SHORT_TEXT_LENGTH), it may be that `addresses + DELIMITER + addr` will be greater than 255 chars already. Thus we need to have the `addresses.length + ADDRESS_DELIMITER.length + addr.length` lookahead logic to make sure we never pass 255 chars.

Okay with you @pranavpeshwe @fisehara, or is there something I'm missing?

pranavpeshwe: @cywang117 Thanks for replying. I did not get the explanation, though.
What I meant to ask is, could it not have been ` return addresses.length + ADDRESS_DELIMITER.length + addr.length <= SHORT_TEXT_LENGTH  ? addresses + ADDRESS_DELIMITER + addr : addresses;` ?  Why are we limiting ourselves to only 254 characters out of the available 255 by using `<` instead of `<=` in the check?

cywang117: Ahh I see.. I think you're right and it should be `<=`. Changing it with the next force push

cywang117:
@zoobot Thanks for the review. I'll address your points individually:
- SHORT_TEXT_LENGTH --> TEXT_LENGTH: This was a conscious decision due to [Pine translating SBVR of Concept Type: Short Text into VARCHAR(255)](https://github.com/balena-io/open-balena-api/blob/16ac5bbe4797a1c5036181ba34bc1ab0a2d77254/src/balena.sbvr#L179). Might be better to keep it as-is to keep the association clear.

- shorterByOneWord: May not necessarily work as it only cuts off the last word, but with addresses we want to truncate the last N words such that the truncated text is shorter than 255 and no addresses are partial. However I do see your point about making the function more generic, and [here](https://github.com/balena-io/open-balena-api/pull/1068/files#diff-d9d7682a3f0225f18bf8ab8e8cc159fdb3df3bf481cc92e0ac3b17a6a8328ab4R46-R55) is my latest attempt in https://github.com/balena-io/open-balena-api/commit/262280c9859594d7c06ee19fe38d99f6a038fa96. I've removed the `.trim()` because it's not necessary given the logic in the ternary. Let me know if there are any questions & we can continue discussing.

cywang117:
@zoobot I see, thanks for clarifying! I definitely did not read the full function, and just read the function name 😅

cywang117:
@balena-ci rebase

cywang117:
@balena-ci I self-certify!

cywang117:
It looks like I don't have write access to the repo. Could someone with write access approve please?

fisehara:
@cywang117 sure :)
@balena-ci I self-certify!

fisehara:
@balena-ci I self-certify!

cywang117:
Thanks @fisehara !

src/features/device-state/state-patch-utils.ts
```diff
@@ -39,6 +39,46 @@ export const v2ValidPatchFields: Array<
...
+	keysToTruncate.forEach((key) => {
+		if (
+			typeof object[key] !== 'string' ||
+			object[key].length <= SHORT_TEXT_LENGTH
+		) {
+			return;
+		}
+		if (['ip_address', 'mac_address'].includes(key)) {
+			object[key] = truncateText(object[key]);
+		}
```
thgreasi, line 77: @fisehara @cywang117 I guess there is no need to have both the `keysToTruncate` & the `['ip_address', 'mac_address']` here, right?
B/c it look like an unnecessary safeguard on whether the invoker used shorttext props or not.
Given how the source looks today, maybe we could just do:
```suggestion
const default_short_text_fields_to_trancate = ['ip_address', 'mac_address'];
export const truncateShortTextFields = (
	object: Dictionary<any>,
	keysToTruncate: ValidPatchFields = default_short_text_fields_to_trancate,
) => {
	for (const key of keysToTruncate) {
		if (
			typeof object[key] !== 'string' ||
			object[key].length <= SHORT_TEXT_LENGTH
		) {
			return;
		}
		object[key] = truncateText(object[key]);
```
so that this array doesn't have to get repeated 3 times and we have a single definition. Less GC as well :)
WRT the `for...of`, Page recently replaced tons of `.forEach` b/c it's slightly faster, so let's do it here as well for consistency.

cywang117: See https://github.com/balena-io/open-balena-api/pull/1085 for the application of the changes suggested here.

PR #1079, opened 2022-08-02
Optimize device logs in the case that loki is disabled

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1081, opened 2022-08-02
Device v3 compatible fleet default state

### LINKED ISSUE #1071: API generates virtual fleet target state
- Custom endpoint for fleet target state v3 in open-balena-api (no computed terms solution for now)
  - specific target state JSON format should be kept / no OData data structure desired as state result
- endpoint should provide creating 'pinned' fleet target state for a given fleet release ID
- fleet target state endpoint shall only be used for device images that contain OS with overlay2 and supervisor version >= v13

For now the fleet target state will not deliver the supervisor app state (as the device target state does): 
>  The supervisor app is not necessary as part of the target state on this version. The OS preloads the supervisor image through a different process and the supervisor target state (device.should_be_managed_by__release) is only assigned on provision, when the supervisor reports state for the first time.

fisehara:
Default fleet state compatible to device state v3

resolves #1071

Change-type: minor
See: https://jel.ly.fish/thread-77a17ab4-730c-4725-955c-4c719771b8c0

### DIFF
--- a/src/features/device-state/index.ts
+++ b/src/features/device-state/index.ts
@@ -9,6 +9,7 @@ import { stateV2 } from './routes/state-get-v2';
 import { stateV3 } from './routes/state-get-v3';
 import { statePatchV2 } from './routes/state-patch-v2';
 import { statePatchV3 } from './routes/state-patch-v3';
+import { fleetStateV3 } from './routes/fleet-state-get-v3';
 
 export {
 	setReadTransaction,
@@ -45,6 +46,11 @@ export const setup = (app: Application) => {
 		statePatchV2,
 	);
 	app.patch('/device/v3/state', middleware.authenticatedApiKey, statePatchV3);
+	app.get(
+		'/device/v3/fleet/:fleetUuid/state',
+		middleware.authenticated,
+		fleetStateV3,
+	);
 };
 
 export interface Events {
--- a/src/features/device-state/routes/fleet-state-get-v3.ts
+++ b/src/features/device-state/routes/fleet-state-get-v3.ts
@@ -1,0 +1,175 @@
+import type { RequestHandler } from 'express';
+import type { Request } from 'express';
+
+import * as _ from 'lodash';
+import {
+	captureException,
+	handleHttpErrors,
+} from '../../../infra/error-handling';
+import { readTransaction } from '../state-get-utils';
+import { Expand } from 'pinejs-client-core';
+
+import { sbvrUtils, errors } from '@balena/pinejs';
+import {
+	buildAppFromRelease,
+	getConfig,
+	releaseExpand,
+	StateV3,
+} from './state-get-v3';
+const { api } = sbvrUtils;
+const { UnauthorizedError } = errors;
+
+type FleetStateV3 = {
+	[uuid: string]: Omit<StateV3[string], 'is_managed_by__device' | 'apps'> & {
+		apps: {
+			[uuid: string]: Omit<
+				StateV3[string]['apps'][string],
+				'is_managed_by__device'
+			>;
+		};
+	};
+};
+
+const fleetExpand: Expand = {
+	application_config_variable: {
+		$select: ['name', 'value'],
+		$orderby: {
+			name: 'asc',
+		},
+	},
+	application_environment_variable: {
+		$select: ['name', 'value'],
+	},
+	should_be_running__release: releaseExpand,
+	service: {
+		$select: ['id', 'service_name'],
+		$expand: {
+			service_environment_variable: {
+				$select: ['name', 'value'],
+			},
+			service_label: {
+				$select: ['label_name', 'value'],
+			},
+			is_built_by__image: {
+				$select: ['id', 'is_stored_at__image_location'],
+			},
+		},
+	},
+};
+
+const stateQuery = _.once(() =>
+	api.resin.prepare<{ uuid: string }>({
+		resource: 'application',
+		id: { uuid: { '@': 'uuid' } },
+		options: {
+			$expand: fleetExpand,
+		},
+	}),
+);
+
+const releaseQuery = _.once(() =>
+	api.resin.prepare<{ commit: string; fleetId: number }>({
+		resource: 'release',
+		options: {
+			$filter: {
+				commit: { '@': 'commit' },
+				belongs_to__application: { '@': 'fleetId' },
+				status: 'success',
+			},
+			...releaseExpand,
+		},
+	}),
+);
+
+const getFleet = async (req: Request, uuid: string) => {
+	const fleet = await readTransaction((tx) =>
+		stateQuery()({ uuid }, undefined, { req, tx }),
+	);
+
+	if (fleet == null) {
+		throw new UnauthorizedError();
+	}
+
+	return fleet;
+};
+
+const getSuccessfulReleaseForFleetAndCommit = async (
+	req: Request,
+	commit: string,
+	fleetId: number,
+) => {
+	const [release] = await readTransaction((tx) =>
+		releaseQuery()({ commit, fleetId }, undefined, { req, tx }),
+	);
+
+	if (release == null) {
+		throw new UnauthorizedError();
+	}
+
+	return release;
+};
+
+const getFleetAppsForState = (
+	fleet: AnyObject,
+	release: AnyObject | undefined,
+	config: Dictionary<string>,
+): FleetStateV3[string]['apps'] => {
+	return {
+		[fleet.uuid]: {
+			id: fleet.id,
+			name: fleet.app_name,
+			is_host: fleet.is_host,
+			class: fleet.is_of__class,
+			...(release != null && {
+				releases: buildAppFromRelease(undefined, fleet, release, config),
+			}),
+		},
+	};
+};
+
+const getFleetStateV3 = async (
+	req: Request,
+	uuid: string,
+	releaseUuid: string | undefined = undefined,
+): Promise<FleetStateV3> => {
+	const fleet = await getFleet(req, uuid);
+	const config = getConfig(undefined, fleet);
+	let release: AnyObject | undefined;
+	if (releaseUuid) {
+		release = await getSuccessfulReleaseForFleetAndCommit(
+			req,
+			releaseUuid,
+			fleet.id,
+		);
+	} else {
+		release = fleet.should_be_running__release[0];
+	}
+	const apps = getFleetAppsForState(fleet, release, config);
+
+	return {
+		[uuid]: {
+			name: fleet.app_name,
+			apps,
+			config,
+		},
+	};
+};
+
+export const fleetStateV3: RequestHandler = async (req, res) => {
+	const { fleetUuid } = req.params;
+	const { releaseUuid } = req.body;
+	if (!fleetUuid) {
+		return res.status(400).end();
+	}
+
+	try {
+		const state = await getFleetStateV3(req, fleetUuid, releaseUuid);
+		res.json(state);
+	} catch (err) {
+		if (handleHttpErrors(req, res, err)) {
+			return;
+		}
+		captureException(err, 'Error getting default flee state', { req });
+		res.status(500).end();
+	}
+};
--- a/src/features/device-state/routes/state-get-v3.ts
+++ b/src/features/device-state/routes/state-get-v3.ts
@@ -83,8 +83,8 @@ export type StateV3 = {
 	};
 };
 
-function buildAppFromRelease(
-	device: AnyObject,
+export function buildAppFromRelease(
+	device: AnyObject | undefined,
 	application: AnyObject,
 	release: AnyObject,
 	config: Dictionary<string>,
@@ -108,24 +108,32 @@ function buildAppFromRelease(
 		// extract the per-image information
 		const image = ipr.image[0];
 
-		const si = serviceInstallFromImage(device, image);
+		const si = serviceInstallFromImage(device ?? application, image);
 		if (si == null) {
 			throw new Error(
-				`Could not find service install for device: '${
+				`Could not find service install for device or application: '${
 					application.uuid
 				}', image: '${image?.id}', service: '${JSON.stringify(
 					image?.is_a_build_of__service,
-				)}', service installs: '${JSON.stringify(device.service_install)}'`,
+				)}', service: '${
+					device
+						? JSON.stringify(device.service_install)
+						: JSON.stringify(application.service)
+				}'`,
 			);
 		}
-		const svc = si.service[0];
-
+		const svc = si.service?.[0] ?? si;
 		const environment: Dictionary<string> = {};
 		varListInsert(ipr.image_environment_variable, environment);
 		varListInsert(application.application_environment_variable, environment);
 		varListInsert(svc.service_environment_variable, environment);
-		varListInsert(device.device_environment_variable, environment);
-		varListInsert(si.device_service_environment_variable, environment);
+
+		if (device?.device_environment_variable) {
+			varListInsert(device.device_environment_variable, environment);
+		}
+		if (si?.device_service_environment_variable) {
+			varListInsert(si.device_service_environment_variable, environment);
+		}
 
 		const labels: Dictionary<string> = {};
 		for (const { label_name, value } of [
@@ -178,7 +186,7 @@ function buildAppFromRelease(
 	};
 }
 
-const releaseExpand = {
+export const releaseExpand = {
 	$select: ['id', 'commit', 'composition'],
 	$expand: {
 		contains__image: {
@@ -334,19 +342,27 @@ const getDevice = async (req: Request, uuid: string) => {
 	return device;
 };
 
-const getConfig = (device: AnyObject) => {
+export const getConfig = (
+	device: AnyObject | undefined,
+	application: AnyObject = device?.belongs_to__application[0],
+) => {
 	const config: Dictionary<string> = {};
 
 	// add any app-specific config values...
-	const userAppFromApi: AnyObject = device.belongs_to__application[0];
-	varListInsert(
-		userAppFromApi.application_config_variable,
-		config,
-		rejectUiConfig,
-	);
+
+	if (application) {
+		varListInsert(
+			application.application_config_variable,
+			config,
+			rejectUiConfig,
+		);
+	}
 
 	// override with device-specific values...
-	varListInsert(device.device_config_variable, config, rejectUiConfig);
+	if (device) {
+		varListInsert(device.device_config_variable, config, rejectUiConfig);
+	}
+
 	filterDeviceConfig(config);
 	setDefaultConfigVariables(config);
 
--- a/test/21_fleet-target-state.ts
+++ b/test/21_fleet-target-state.ts
@@ -1,0 +1,250 @@
+import * as mockery from 'mockery';
+import { expect } from 'chai';
+import { supertest } from './test-lib/supertest';
+import { version } from './test-lib/versions';
+import * as configMock from '../src/lib/config';
+import * as fixtures from './test-lib/fixtures';
+import { Application, User } from '../src/balena-model';
+
+const POLL_MSEC = 2000;
+const TIMEOUT_SEC = 1;
+
+// @ts-expect-error mock the value for the default poll interval...
+configMock['DEFAULT_SUPERVISOR_POLL_INTERVAL'] = POLL_MSEC;
+
+// @ts-expect-error mock the value for the timeout grace period...
+configMock['API_HEARTBEAT_STATE_TIMEOUT_SECONDS'] = TIMEOUT_SEC;
+
+// register the mocks...
+mockery.registerMock('../src/lib/config', configMock);
+
+// test fleet default state helper
+const appMatcherFunc = (svc: AnyObject, fxProp: AnyObject) =>
+	svc.hasOwnProperty('environment') && !!fxProp;
+const svcMatcherFunc = (svc: AnyObject, fxProp: AnyObject) =>
+	svc.id === fxProp.service.__id;
+const imgMatcherFunc = (svc: AnyObject, fxProp: AnyObject) =>
+	svc.image_id === fxProp.release_image.image.__id;
+
+const checkServiceProperties = (
+	fixture: AnyObject,
+	fleetStateReleasesServices: AnyObject,
+	propertyCollectionName: string,
+	propKeyIdentifier: string,
+	matcherFunction: (svc: AnyObject, fxProperty: AnyObject) => boolean,
+) => {
+	for (const fxProperty of Object.values(fixture)) {
+		for (const svc of Object.values(fleetStateReleasesServices)) {
+			expect(svc).to.haveOwnProperty(propertyCollectionName);
+			if (matcherFunction(svc, fxProperty)) {
+				expect(svc[propertyCollectionName]).to.include({
+					[fxProperty[propKeyIdentifier]]: fxProperty.value,
+				});
+			} else {
+				expect(svc[propertyCollectionName]).to.not.include({
+					[fxProperty[propKeyIdentifier]]: fxProperty.value,
+				});
+			}
+		}
+	}
+};
+
+(['v3'] as const).forEach((stateVersion) =>
+	describe(`Fleet State ${stateVersion}`, function () {
+		let fx: fixtures.Fixtures;
+		let application: Application;
+		let admin: User;
+		let releases: AnyObject;
+		let fleetStateEndpoint: string;
+
+		before(async function () {
+			fx = await fixtures.load('21-fleet-target-state');
+			admin = fx.users.admin;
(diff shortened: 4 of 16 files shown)


aethernet (commented):
looks fine to me, but i lack the experience to approve

zoobot (commented):
Looks good, has what we need for preloading! :)

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,372 @@
...
+		[release.commit]: {
+			id: release.id,
+			...(Object.keys(services).length > 0 ? { services } : undefined),
+			...(composition.networks != null && { networks: composition.networks }),
+			...(composition.volumes != null && { volumes: composition.volumes }),
+		},
+	};
+}
+
+const releaseExpand = {
```
Page-, line 191: Can this be shared with device-state-v3?

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,372 @@
...
+import { Expand } from 'pinejs-client-core';
+
+import { sbvrUtils, errors } from '@balena/pinejs';
+import { Application } from '../../../balena-model';
+const { api } = sbvrUtils;
+const { UnauthorizedError } = errors;
+
+type LocalStateApp = FleetStateV3[string]['apps'][string];
+type ServiceComposition = AnyObject;
+export type FleetStateV3 = {
```
Page-, line 27: Can this be shared with device-state-v3? I think you can do it with a bunch of `{ ...DeviceStateV3['...], x: ... }` until you get to the `Exclude<x, 'is_managed_by__device'>` if you really want, although since that property is optional anyway you shouldn't need to have to remove it?

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,372 @@
...
+		$select: ['name', 'value'],
+		$orderby: {
+			name: 'asc',
+		},
+	},
+	application_environment_variable: {
+		$select: ['name', 'value'],
+	},
+	should_be_running__release: releaseExpand,
+	application__has__service_name: {
```
Page-, line 228:
```suggestion
	service: {
```

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,372 @@
...
+		$select: ['id', 'service_name'],
+		$expand: {
+			application__has__service_name__has__name: {
```
Page-, line 231:
```suggestion
			service environment variable: {
```

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,372 @@
...
+				$select: ['name', 'value'],
+			},
+			application__has__service_name__has__label_name: {
```
Page-, line 234:
```suggestion
			service label: {
```

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,372 @@
...
+		try {
+			composition = JSON.parse(release.composition);
+		} catch {
+			composition = {};
+		}
+	}
+	for (const ipr of release.contains__image as AnyObject[]) {
+		// extract the per-image information
+		const image = ipr.image[0];
+		const svc = serviceBuildByImage(fleet, image);
```
Page-, line 126: Any reason this can't use `serviceInstallFromImage` the same as for device-state?

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,372 @@
...
+	} else {
+		serviceId = image.is_a_build_of__service;
+	}
+
+	return fleet.application__has__service_name.find(
+		(fleetService: AnyObject) => fleetService.id === serviceId,
+	);
+};
+
+function buildFleetAppFromRelease(
```
Page-, line 104: I think we should be able to share this entire function with device-state-v3 by just making device an optional parameter there (or I guess not optional but rather `AnyObject | undefined` would make more sense so you have to explicitly pass undefined)

Page- (changes requested):
Overall I'd like to see at least half of the code from this PR disappear due to being shared with device state since ideally they should be identical and duplicating the code with minor/no differences is going to be a maintenance nightmare

src/features/device-state/routes/fleet-state-get-v3.ts
fisehara: yes, thx!

src/features/device-state/routes/fleet-state-get-v3.ts
fisehara: serviceInstallFromImage accesses the device 
```
	return _.find(device.service_install, (si) => si.service[0].id === id);
```
I can add a switch that the input is either device | fleet and if the input has `service_install` (for device input) I take it otherwise I check for existence of 'service' (for fleet / application as input)

src/features/device-state/routes/fleet-state-get-v3.ts
fisehara: thx

src/features/device-state/routes/fleet-state-get-v3.ts
fisehara: thx

src/features/device-state/routes/fleet-state-get-v3.ts
fisehara: thx

src/features/device-state/routes/fleet-state-get-v3.ts
fisehara: Solved with:
```
type FleetStateV3 = {
	[uuid: string]: Omit<StateV3[string], 'is_managed_by__device' | 'apps'> & {
		apps: {
			[uuid: string]: Omit<
				StateV3[string]['apps'][string],
				'is_managed_by__device'
			>;
		};
	};
};
```

src/features/device-state/routes/fleet-state-get-v3.ts
fisehara: yep, applied also for config.. Thx

src/features/device-state/index.ts
```diff
@@ -45,6 +46,12 @@ export const setup = (app: Application) => {
 		statePatchV2,
 	);
 	app.patch('/device/v3/state', apiKeyMiddleware, statePatchV3);
+	app.get('/device/v3/fleet-state/:fleetUuid', apiKeyMiddleware, fleetStateV3);
```
pipex, line 49: I know this was discussed, so is just a nitpick, but how about

`/device/v3/state/fleet/:fleetUuid`

or

``/device/v3/state/from-fleet/:fleetUuid`

Using `fleet-state` is not totally correct, since in the end the endpoint returns the device state as inherited from the fleet state.

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,173 @@
...
+		release = await getReleaseForCommit(req, releaseUuid);
+	} else {
+		release = getReleaseForFleet(fleet);
+	}
+	const apps = getFleetAppsForState(fleet, release, config);
+
+	const state: FleetStateV3 = {
+		[uuid]: {
+			name: fleet.app_name,
+			apps,
```
pipex, line 150: can a fleet have multiple apps? Will it be able to have multiple apps in the future? Since the `uuid` is used both to index the return from this method and the return from `getFleetAppsForState`, won't this result in a redundant object?

```
{
    abc: {
        name: 'my-fleet",
        apps: {
             abc: {
                id: 1,
                name: 'my-fleet'
                releases: {}
             }
        }
    }
}
```

Is this on purpose?

src/features/device-state/index.ts
zoobot: /device/v3/state/fleet/:fleetUuid that one is clear, imo.

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -108,24 +108,32 @@ function buildAppFromRelease(
 		// extract the per-image information
 		const image = ipr.image[0];
 
-		const si = serviceInstallFromImage(device, image);
+		const si = serviceInstallFromImage(device ?? application, image);
```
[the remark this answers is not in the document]
pipex: Oh, sorry, I see that you are checking the `device` below to add device environment variables and other things. 

I don't really like this pattern, but I since so few things change in the function body depending on whether is an application or a device that I think it's fine.

src/features/device-state/routes/fleet-state-get-v3.ts
fisehara: @pipex 
Yes, a fleet may consists of multiple apps, namely userApps, supervisor and hostApp. 
Thus, now the fleet consists only of it's own user application

src/features/device-state/routes/state-get-v3.ts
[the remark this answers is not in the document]
fisehara: The idea is an as close as possible overlap with the device target state v3 code. Hence, we decided to go this way instead of introducing redundant code.

src/features/device-state/index.ts
fisehara: I'm also in favour of `/device/v3/state/fleet/:fleetUuid`

src/features/device-state/index.ts
```diff
@@ -45,6 +46,12 @@ export const setup = (app: Application) => {
...
+	app.get(
+		'/device/v3/fleet-state/:fleetUuid/release/:releaseUuid',
```
dfunckt, line 51: I know there was discussion about how the endpoints should look like and the result was what you have, but seeing them now I think the following is much better:

```suggestion
	app.get('/device/v3/fleet/:fleetUuid/state', apiKeyMiddleware, fleetStateV3);
	app.get(
		'/device/v3/fleet/:fleetUuid/release/:releaseUuid/state',
```

Because:

- it does not imply that the endpoint is about "fleet state" when a specific release is requested (i.e. the second endpoint)
- it follows a little more closely the original format of the state endpoint, where the URL ends with `/state`.

src/features/device-state/index.ts
aethernet: cf this conversation : https://github.com/balena-io/open-balena-api/pull/1081#discussion_r1011501666

src/features/device-state/index.ts
aethernet: adopted by @fisehara (https://balena.zulipchat.com/#narrow/stream/332163-_all/topic/curious.20about.20preloading.20efforts/near/308394690)

aethernet:
Had to rebase and clear conflict.

aethernet:
rebased and resolved conflicts again

Page- (changes requested):
It looks like there were some issues whilst rebasing that reverted a load of stuff - can you fix and re-request a review please

.github/actions/test/action.yml
```diff
@@ -17,5 +17,5 @@ runs:
       shell: bash
       run: |
         # TODO: Drop the custom test action once we have a better way to
-        # generate the materialized config, other than passing an env var.
+        # generate the meterialized onfig, other than passing an env var.
```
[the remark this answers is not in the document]
aethernet: Not completely done, i reverted too far.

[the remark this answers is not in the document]
fisehara: @aethernet I have the un-rebased commit on my machine and can do the rebase to latest mater locally. No issue!

fisehara:
@aethernet just in case, here is a master-rebased work copy of this (here) PR
https://github.com/balena-io/open-balena-api/tree/fisehara/1071-virtual-fleet-target-state

aethernet:
@fisehara thanks, I got it back :)

src/features/device-state/index.ts
```diff
@@ -45,6 +46,16 @@ export const setup = (app: Application) => {
...
+	app.get(
+		'/device/v3/fleet/:fleetUuid/state',
+		middleware.authenticated,
+		fleetStateV3,
+	);
+	app.get(
+		'/device/v3/fleet/:fleetUuid/release/:releaseUuid/state',
+		middleware.authenticated,
+		fleetStateV3,
+	);
```
thgreasi, line 58: Given that the releaseUuid is used like an optional parameter shouldn't we just make it an optional query parameter rather than having 2 different endpoints?
On principle having `/device/v3/fleet/:fleetUuid/state` & `/device/v3/fleet/:fleetUuid/release/:releaseUuid/state` returning the same result sounds a bit sketchy. I also think that from REST semantics pov `/device/v3/fleet/:fleetUuid/release/:releaseUuid/state` might not be that correct, since it sounds like we are asking for the target state of the release of that fleet, while what we really are asking is the target state of the fleet when configured to run the provided release (imo that points that the release should be a parameter).

fisehara:
Don't merge, I figured out that the test don't run as expected. Am working on the test to run in completeness and test properly

fisehara:
@aethernet just a heads up, the test fixture loaders where linking wrong ids between image environment variables and images. They are connected through the `is part of release` table and the id where from the linkage table instead of the real image id.
Ready to review :)

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,174 @@
...
+		throw new UnauthorizedError();
+	}
+
+	return release[0] ?? {};
+};
+
+const getReleaseForFleet = (fleet: AnyObject): AnyObject | undefined => {
+	if (fleet.should_be_running__release[0] != null) {
+		return fleet.should_be_running__release[0];
+	}
```
Page-, line 111: Can remove the redundant if
```suggestion
	return fleet.should_be_running__release[0];
```

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -270,7 +276,13 @@ const stateQuery = _.once(() =>
...
 		options: {
 			$select: ['device_name', 'os_version'],
-			$expand: deviceExpand,
+			$expand: {
+				...deviceExpand,
+				manages__device: {
+					$select: ['uuid', 'device_name'],
+					$expand: deviceExpand,
+				},
+			},
```
Page-, line 285: Can you revert this - I'm pretty sure it's a failed rebase artifact
```suggestion
			$expand: deviceExpand,
```

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -298,6 +310,13 @@ const getStateV3 = async (req: Request, uuid: string): Promise<StateV3> => {
 		},
 	};
 
+	for (const depDev of device.manages__device as AnyObject[]) {
+		state[depDev.uuid] = {
+			name: depDev.device_name,
+			apps: getUserAppState(depDev, getConfig(depDev)),
+		};
+	}
+
```
Page-, line 319: Can you revert this - I'm pretty sure it's a failed rebase artifact
```suggestion
```

src/features/device-state/routes/fleet-state-get-v3.ts
fisehara: Thanks!

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,174 @@
...
+};
+
+const stateQuery = _.once(() =>
+	api.resin.prepare<{ uuid: string }>({
+		resource: 'application',
+		id: { uuid: { '@': 'uuid' } },
+		options: {
+			$expand: {
+				...fleetExpand,
+			},
```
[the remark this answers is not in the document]
fisehara: :+1:

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,174 @@
...
+const releaseQuery = _.once(() =>
+	api.resin.prepare<{ commit: string }>({
+		resource: 'release',
+		options: {
+			$filter: {
+				commit: { '@': 'commit' },
+			},
+			...releaseExpand,
+		},
+	}),
```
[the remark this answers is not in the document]
fisehara: Oh I see, yes! Thanks

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,174 @@
...
+);
+
+const getFleet = async (req: Request, uuid: string) => {
+	const fleet = await readTransaction((tx) =>
+		stateQuery()({ uuid }, undefined, { req, tx }),
+	);
+
+	if (!fleet) {
```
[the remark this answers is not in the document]
fisehara: For sure, thanks!

src/features/device-state/routes/state-get-v3.ts
fisehara: done.

src/features/device-state/routes/state-get-v3.ts
fisehara: 👍

src/features/device-state/routes/fleet-state-get-v3.ts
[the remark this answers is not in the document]
fisehara: Checking it, it's not possible as commit is not a unique key for release resource:
```
Fact type: release has commit
	Necessity: each release has exactly one commit.
```

From my understanding commit cannot be used as an ID field though and needs to be filtered.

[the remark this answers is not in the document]
Page-: You're right @fisehara, which also means that this query is invalid as it can return many results and it just picks the first one that happens to be returned which could easily be the wrong one and be a big problem

Page- (changes requested):

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -334,20 +340,25 @@ const getDevice = async (req: Request, uuid: string) => {
...
 		config,
 		rejectUiConfig,
 	);
 
 	// override with device-specific values...
-	varListInsert(device.device_config_variable, config, rejectUiConfig);
-	filterDeviceConfig(config, device.os_version);
+	if (device) {
+		varListInsert(device.device_config_variable, config, rejectUiConfig);
+		filterDeviceConfig(config, device.os_version);
```
Page-, line 360: Should we ask the fleet config to pass in an os version? As from the comment on `filterDeviceConfig` it seems important that it is run

fisehara: @Page- I see what you mean. On first sight I'd argue that the image that goes through this process is for devices greater than the os version constraint represented in the filterDeviceConfig. Hence, this filter can be applied always?

@aethernet Please add your insights on this for better clarity.

aethernet: As the filter only remove one var (`configVars.RESIN_HOST_LOG_TO_DISPLAY`) if OS is below v2 I think we're safe. 

From a preloading pov, we'll never preload v1 OS, so we're good (to be more precise preloading requires `overlay2` I don't remember the minimal os version shipping with an overlay2 engine but far above v2.

fisehara: @aethernet The check is for os semver >= 2.0.0 This would basically mean we always have to clean this up

aethernet: indeed sorry misread the sign, it should be always applied

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -334,20 +340,25 @@ const getDevice = async (req: Request, uuid: string) => {
...
 		config,
 		rejectUiConfig,
 	);
 
 	// override with device-specific values...
-	varListInsert(device.device_config_variable, config, rejectUiConfig);
-	filterDeviceConfig(config);
+	if (device) {
+		varListInsert(device.device_config_variable, config, rejectUiConfig);
+		filterDeviceConfig(config);
```
Page-, line 360: This should be outside the `if`

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -108,24 +108,30 @@ function buildAppFromRelease(
...
 		const environment: Dictionary<string> = {};
 		varListInsert(ipr.image_environment_variable, environment);
 		varListInsert(application.application_environment_variable, environment);
 		varListInsert(svc.service_environment_variable, environment);
-		varListInsert(device.device_environment_variable, environment);
-		varListInsert(si.device_service_environment_variable, environment);
+
+		if (device) {
+			varListInsert(device.device_environment_variable, environment);
+			varListInsert(si.device_service_environment_variable, environment);
```
Page-, line 133: Should this be outside the if?

src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -0,0 +1,180 @@
...
+	);
+
+	if (release == null) {
+		throw new UnauthorizedError();
+	}
+
+	return release;
+};
+
+const getReleaseForFleet = (fleet: AnyObject): AnyObject | undefined => {
```
Page-, line 112: I'd probably in-line this, it's only used in one place so it'll be clearer if it's in-line and will automatically be typed nicely as we improve query typings (the explicit types for this function call will undo any typing improvements outside of the function)

test/21_fleet-target-state.ts
```diff
@@ -0,0 +1,251 @@
+import * as _ from 'lodash';
```
Page-, line 1: This is unused?
```suggestion
```

test/21_fleet-target-state.ts
```diff
@@ -0,0 +1,251 @@
...
+
+const checkServiceProperties = (
+	fixture: AnyObject,
+	fleetStateReleasesServices: AnyObject,
+	propertyCollectionName: string,
+	propKeyIdentifier: string,
+	matcherFunction: (svc: AnyObject, fxProperty: AnyObject) => boolean,
+) => {
+	for (const fxProperty of Object.values(fixture)) {
+		for (const svc of Object.values<any>(fleetStateReleasesServices)) {
```
Page-, line 38: Why the `any`?

src/features/device-state/routes/state-get-v3.ts
fisehara: Oh right, thanks!

src/features/device-state/routes/fleet-state-get-v3.ts
fisehara: Thanks for the hint about the added value on the typings!

test/21_fleet-target-state.ts
fisehara: Yep

test/21_fleet-target-state.ts
fisehara: Yes thanks, leaving it without the template type as any is inferred from the AnyObject type.

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -334,19 +342,25 @@ const getDevice = async (req: Request, uuid: string) => {
...
+export const getConfig = (
+	device: AnyObject | undefined,
+	application: AnyObject = device?.belongs_to__application[0],
+) => {
 	const config: Dictionary<string> = {};
 
 	// add any app-specific config values...
-	const userAppFromApi: AnyObject = device.belongs_to__application[0];
+
 	varListInsert(
```
Page-, line 353: I think it'd make sense to use the same method for both missing app and missing device, eg
```ts
	if (application) {
		varListInsert(application.application_config_variable, config, rejectUiConfig);
	}

File: src/features/device-state/routes/fleet-state-get-v3.ts [hunk 27]
```diff
@@ -0,0 +1,176 @@
...
+	const apps = getFleetAppsForState(fleet, release, config);
+
+	const state: FleetStateV3 = {
+		[uuid]: {
+			name: fleet.app_name,
+			apps,
+			config,
+		},
+	};
+	return state;
```
Page-, line 156: You can simplify a bit by directly returning, and that will mean the type is inferred from the function return type rather than needing to be explicitly typed
```suggestion
	return {
		[uuid]: {
			name: fleet.app_name,
			apps,
			config,
		},
	};
```

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -334,19 +342,27 @@ const getDevice = async (req: Request, uuid: string) => {
...
-	const userAppFromApi: AnyObject = device.belongs_to__application[0];
-	varListInsert(
-		userAppFromApi.application_config_variable,
-		config,
-		rejectUiConfig,
-	);
+
+	if (application) {
+		varListInsert(
+			application?.application_config_variable ?? [],
```
Page-, line 355:
```suggestion
			application.application_config_variable,
```

Page- (approved):

PR #1082, opened 2022-08-04
Add caching for the trust proxy function when trusting by ip

Page-:
Change-type: patch


src/lib/config.ts
```diff
@@ -403,8 +403,22 @@ if (trustProxy === 'true') {
 	// If it's 'true' enable it
 	trustProxyValue = true;
 } else if (trustProxy.includes('.') || trustProxy.includes(':')) {
-	// If it looks like an ip use as-is
-	trustProxyValue = trustProxy;
+	// If it looks like an ip the compile the trust function directly and memoize it, since the trust
```
thgreasi, line 406:
```suggestion
	// If it looks like an ip then compile the trust function directly and memoize it, since the trust
```
?

thgreasi (approved):

PR #1084, opened 2022-08-06
vars-update-trigger: Fix 42P01 error when the affected envs are too many

thgreasi:
Change-type: patch
See: https://github.com/balena-io/balena-api/issues/3742
See: https://sentry.io/organizations/balena/issues/3475598791/
See: https://www.flowdock.com/app/rulemotion/public-s-community/threads/GrL4sR7imv_sfdTqZsgu80lMvpg

### DIFF
--- a/src/features/vars-schema/hooks/vars-update-trigger.ts
+++ b/src/features/vars-schema/hooks/vars-update-trigger.ts
@@ -1,5 +1,6 @@
 import { sbvrUtils, hooks } from '@balena/pinejs';
-import type { Filter } from 'pinejs-client-core';
+import * as _ from 'lodash';
+import type { FilterObj } from 'pinejs-client-core';
 
 import { captureException } from '../../../infra/error-handling';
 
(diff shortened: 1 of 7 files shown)


src/features/vars-schema/hooks/vars-update-trigger.ts
```diff
@@ -9,14 +10,21 @@ interface CustomObject {
...
 const addEnvHooks = (
 	resource: string,
 	buildFilter: (
 		args: hooks.HookArgs & {
 			tx: Tx;
 		},
-	) => Promise<Filter | undefined>,
+	) => Promise<
+		| FilterObj
+		| [affectedIds: number[], filterBuilder: (ids: number[]) => FilterObj]
```
thgreasi, line 25: Also tried to use differentiating props like
```suggestion
		| Filter
		| {
			affectedIds: number[],
			filterBuilder: (ids: number[]) => Filter
		  }
```
but `typeof filter.filterBuilder === 'function'` wasn't enough for TS to figure out the type of affectedIds properly as well :(

src/features/vars-schema/hooks/vars-update-trigger.ts
```diff
@@ -9,14 +10,21 @@ interface CustomObject {
 	affectedDevices?: number[];
 }
 
+// See: https://www.postgresql.org/docs/13/protocol-message-formats.html#:~:text=The%20number%20of%20parameter%20values%20that%20follow
+const MAX_SAFE_SQL_BINDS = Math.pow(2, 16) - 1;
```
Page-, line 14: You'll probably want to chunk slightly below the limit to allow some binds for other things like computed terms/permissions/etc

thgreasi: @Page- They way I was thinking about it, was that in such a case the test cases would let us know, but I obviously forgot the case that the computed term might be bC only.
Will drop it by 100 to give it some extra room and we can change it later if we need to.

thgreasi: Oh I should also switch it to use the exponential operator `**` !!

thgreasi:
@balena-ci I self-certify!

PR #1086, opened 2022-08-10
Allow trusted ips to make http connections with x-forwarded-proto

Page-:
Change-type: minor

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -403,6 +403,16 @@ function fixProtocolMiddleware(skipUrls: string[] = []): Handler {
 		if (req.protocol === 'https' || skipUrls.includes(req.url)) {
 			return next();
 		}
+		if (req.headers['x-forwarded-for'] == null) {
+			const trust = req.app.get('trust proxy fn') as ReturnType<
+				typeof import('proxy-addr').compile
+			>;
+			if (trust(req.socket.remoteAddress!, 0)) {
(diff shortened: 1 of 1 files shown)


fisehara:
@Page- Is this for calls to the API that are not going through a reverse-proxy?
So the request is not coming from the reverse-proxy (as this would have the XFF header) but it's coming from an internal IP.
The trusted (not proxy) IPs needs to be specified in the process.ENV.TRUST_PROXY?
I'm wondering if the TRUST_PROXY env is still a meaningful name for these trusted IPs, or if they should be given in a different ENV var?

Page-:
>Is this for calls to the API that are not going through a reverse-proxy?

yes @fisehara, eg for https://github.com/balena-io/open-balena-vpn/pull/281 so that the vpn requests can go direct to the api rather than having to go via the AWS load balancer

> The trusted (not proxy) IPs needs to be specified in the process.ENV.TRUST_PROXY?

It's ones we're willing to trust as proxies and hence trust their `x-forwarded-*` headers, so they're capable of setting the `x-forwarded-proto` to https anyway and we'll believe them so this just allows them to make direct connections for their own purposes without having to jump through the hoop of adding the `x-forwarded-proto` when the request originates from them

fisehara:
@Page- LGTM, you want to have another review, or should I approve?

Page-:
@balena-ci I self-certify!

PR #1089, opened 2022-08-16
Remove more unnecessary headers for supervisor requests

Page-:
Supervisor target state requests are usually cached and as such the
majority of the bandwidth is used for sending headers, by removing all
unnecessary headers it allows us the reduce the idle bandwidth from
supervisors by a significant percentage

Change-type: patch

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -321,23 +321,22 @@ export async function setup(app: Application, options: SetupOptions) {
 		fixProtocolMiddleware(['/ping'].concat(options.skipHttpsPaths || [])),
 	);
 
-	app.use((_req, res, next) => {
-		res.set('X-Frame-Options', 'DENY');
-		res.set('X-Content-Type-Options', 'nosniff');
-		next();
-	});
-
 	app.use((req, res, next) => {
-		let origin = req.get('Origin');
 		const userAgent = req.get('User-Agent');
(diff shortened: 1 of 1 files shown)


dfunckt (commented):
👍 

This is likely a mistake from when we split open-balena-api codebase from balena-api because the intent is pretty clear from the comment. Quite costly, presumably.

dfunckt:
I’d like Felipe or @alexgg to comment on the removal of the date header as I think it may be used as a time sync shortcut, but could be wrong.

Page-:
@dfunckt the `/connectivity-check` header is what balenaOS uses for the date header which is unaffected by this change (I had previously removed the date header from it and had to add it back in, hence how I know :P )

Page-:
Also it's not a mistake from the splitting or anything like that, it's just when those other headers were added they were just not put behind the same supervisor guard

pipex:
This looks great to me. The supervisor explicitely uses a couple of headers from the API (`retry-after`, `etag`). 

Regarding the `Date` header, we have thought about using that as a time synchronization mechanism for the supervisor to allow to replace API actions by target state alternatives, e.g. instead of the `reboot` action, we could have a `lastBootGt: X` target state with `X` a date, but that would require some sort of synchronization with the API time. This feature has not yet been researched, so I think removing the `Date` is fine for now.

Page-:
@balena-ci I self-certify!

PR #1090, opened 2022-08-17
Remove misleading date header removal for supervisor requests

Page-:
Since the date header is automatically added when you set the status
you need to remove it after the fact (or disable nodejs adding it
altogether) so it was not having any effect here other than adding
confusion

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1092, opened 2022-08-22
Remove unnecessary CORS headers from non-OPTIONS requests

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1094, opened 2022-08-22
Use a windows+dev-only friendly husky install method

Page-:
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "prepack": "npm run build",
     "test": "IMAGE_NAME=test-open-balena-api ./automation/test.sh",
     "posttest": "docker rmi test-open-balena-api",
-    "prepare": "[ \"$HUSKY\" = 0 ] || husky install"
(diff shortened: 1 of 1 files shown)


Page-:
@balena-ci I self-certify!

package.json
```diff
@@ -26,7 +26,7 @@
     "prepack": "npm run build",
     "test": "IMAGE_NAME=test-open-balena-api ./automation/test.sh",
     "posttest": "docker rmi test-open-balena-api",
-    "prepare": "[ \"$HUSKY\" = 0 ] || husky install"
+    "prepare": "node -e \"try { require('husky').install() } catch (e) {if (e.code !== 'MODULE_NOT_FOUND') throw e}\""
```
thgreasi, line 29: @Page- When I saw that in husky's docs, I was pretty sure you would hate it, but then it seems to be overall better for the cases you mentioned :)

PR #1107, opened 2022-08-29
Correct gpio example env var

cywang117:
Corrected from `"gpio=19=op,dh","gpio=0-25=a2"` which will result in
`gpio=gpio=...` in config.txt being applied by the Supervisor.

If `"gpio=19=op,dh","gpio=0-25=a2"` is set via the dashboard as in the old example, the `config.txt` value will formatted by the Supervisor as `gpio=gpio=19=op` which is invalid.

Change-type: patch
Signed-off-by: Christina Wang <christina@balena.io>


pipex:
Good catch @cywang117. LGTM

Here is the RPI documentation for reference https://www.raspberrypi.com/documentation/computers/config_txt.html#gpio

pipex (approved):
LGTM

PR #1108, opened 2022-08-30
Add logs throttling when exceeding the `LOGS_WRITE_BUFFER_LIMIT`

Page-:
This limits the number of logs we will accept per flush interval with a
warning when logs are dropped

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #1112, opened 2022-09-08
Add logo field to organization model

nitishagar:
Change-Type: minor

See: https://jel.ly.fish/improvement-image-logo-icon-upload-s3-org-fleet-01163f0


src/balena-init.sql
```diff
@@ -24,6 +24,9 @@ ALTER COLUMN "semver prerelease" SET DEFAULT '',
 ALTER COLUMN "semver build" SET DEFAULT '',
 ALTER COLUMN "variant" SET DEFAULT '';
 
+ALTER TABLE "organization"
+ADD COLUMN IF NOT EXISTS "logo" TEXT NULL;
+
```
thgreasi, line 29: That's exactly how the org's sql is defined already.
Since you are not adding any special index or a specific default value, my guess is that you don't need this.

PR #1114, opened 2022-09-14
Prevent unauthenticated requests from proceeding in token auth

dfunckt:
Authentication failure due to "no credentials provided" and due to "invalid credentials provided" are currently treated in exactly the same way: we just let the request proceed and rely on (no) permissions to limit access to resources. This is problematic (it can result in redundant work being done in the API) and misleading ("I've provided creds but I get back no results").

This change separates the latter case and responds immediately with 401, as it should.

This is however a workaround and ideally it would be Pine that makes this distinction and throws an exception appropriately but this would be a breaking change to existing code.

Chat thread: https://jel.ly.fish/thread-79f9d489-8af5-42dd-b45f-84cef2d379a9

Change-type: patch

### DIFF
--- a/src/features/registry/middleware.ts
+++ b/src/features/registry/middleware.ts
@@ -1,13 +1,21 @@
 import * as BasicAuth from 'basic-auth';
 import type { RequestHandler } from 'express';
 
+import { errors } from '@balena/pinejs';
+
 import { retrieveAPIKey } from '../../infra/auth/api-keys';
 
+import { TOKEN_AUTH_BUILDER_TOKEN } from '../../lib/config';
+
+export class NoAuthProvidedError extends errors.UnauthorizedError {}
+
+export class InvalidAuthProvidedError extends errors.UnauthorizedError {}
+
 // Resolves permissions and populates req.user object, in case an api key is used
-// in the password field of a basic authentication header
+// in the password field of a basic authentication header. Also works with JWTs.
 export const basicApiKeyAuthenticate: RequestHandler = async (
 	req,
-	_res,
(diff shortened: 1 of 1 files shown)


src/features/registry/middleware.ts
```diff
@@ -16,9 +16,13 @@ export const basicApiKeyAuthenticate: RequestHandler = async (
...
-		await retrieveAPIKey(req, undefined);
-		next();
+		const apiKey = await retrieveAPIKey(req, undefined);
+		if (apiKey != null) {
+			next();
+		} else {
+			res.status(401).end();
+		}
 	} catch (err) {
-		next(err);
```
Page-, line 22: Isn't there an error handling middleware to handle this error already?

dfunckt: Likely, but I chose to handle it here explicitly like we do on the other auth-related middleware.

dfunckt:
I converted the PR to draft to prevent merging until I'm able to fully determine the impact this will have -- see JF chat thread.

src/features/registry/middleware.ts
Page-: I'd rather have the shared code handle it if such exists so that it can be handled correctly for all cases, eg this should at a minimum have a `handleHttpErrors` call, and should probably be cautious about directly responding with the error message - if we don't know the error type is one that is safe to return to the user then it could leak anything, eg if somehow a db password got into the error message due to it being an error connecting to the db or something

dfunckt:
@balena-ci I self-certify!

PR #1115, opened 2022-09-16
Update pinejs to support filtered $count inside $orderby

thgreasi:
Update @balena/pinejs from 14.50.0 to 14.51.0

Change-type: minor


thgreasi:
@balena-ci I self-certify!

PR #1119, opened 2022-09-29
Force suts runtime platform to linux/amd64

fisehara:
For aarch64 dev hosts explicitly specifying the platform lets docker compose run the image in emulated amd64 environment.

open-balena-base image is amd64 only. Explicitly naming the architecture so that it's executed in amd64 mode

Change-type: patch


fisehara:
obsolete

PR #1124, opened 2022-10-18
Adding forked Express.js compression lib

dsevastianov:
to support brotli compression

Change-type: patch


package.json
```diff
@@ -36,6 +36,7 @@
     "@balena/es-version": "^1.0.1",
     "@balena/node-metrics-gatherer": "^6.0.3",
     "@balena/pinejs": "^14.51.0",
+    "@nicksrandall/compression": "^1.0.3",
```
Page-, line 39: I'd lean towards https://www.npmjs.com/package/compression-next as the repo it's from the "more promising" PR (whatever that means as I don't think anything will be added regardless), but more importantly to me the repo is forked from `expression` and so it's possible to see the diff between the two easily for vetting how out of date/diverged it is. It also has slightly more usage but it's minimal either way
```suggestion
    "compression-next": "^1.0.3",
```

Page- (approved):

PR #1131, opened 2022-10-28
Send a byte back periodically from the log-stream endpoint

pipex:
Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/store.ts
+++ b/src/features/device-logs/lib/store.ts
@@ -234,11 +234,6 @@ function handleStreamingWrite(
 		}
 	}
 
-	// https://stackoverflow.com/a/44780406/1559300
-	req.on('socket', (s) => {
-		s.setKeepAlive(true, 240 * 1000);
-	});
-
 	let publishScheduled = false;
 
 	async function tryPublish() {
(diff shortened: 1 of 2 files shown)


src/features/device-logs/lib/store.ts
```diff
@@ -220,6 +220,18 @@ function handleStreamingWrite(
...
 
+	// NOTE: this is testing code to test logging behavior with cloudflare
+	// these timers need to be defined in a non-harcoded way
+
+	// Wait for the client to finish the grace period and connect the buffers
+	setTimeout(ping, 5 * 1000);
+
+	function ping() {
+		// Send a single byte back every 10 seconds
+		res.write(new Uint8Array([1]));
```
Page-, line 231: I think creating a buffer directly probably makes more sense, ie
```suggestion
		res.write(Buffer.alloc(1, 1));
```
which creates a size 1 buffer and pre-allocates it with the value 1 (default is 0): https://nodejs.org/docs/latest-v18.x/api/buffer.html#static-method-bufferallocsize-fill-encoding

pipex:
@balena-ci rebase

pipex:
This was meant as a possible solution of an issue with logging on cloudflare but it had no impact on the solution. The logging problem has been solved so there is no longer need for this.

PR #1139, opened 2022-11-09
Enable esModuleInterop to ease the transition to ES modules

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1142, opened 2022-11-15
Mark loki-grpc-client as an optional dependency

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1143, opened 2022-11-15
state-get-v2/v3: Do not count expired api key requests as heartbeats

thgreasi:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346009-aspect.2Fcustomer-success/topic/Nebra.20bills

### DIFF
--- a/src/features/device-state/routes/state-get-v2.ts
+++ b/src/features/device-state/routes/state-get-v2.ts
@@ -314,6 +314,8 @@ const stateQuery = _.once(() =>
 
 const getStateV2 = async (req: Request, uuid: string): Promise<StateV2> => {
 	const device = await getDevice(req, uuid);
+	// At this point we are sure that the api key is valid and not expired.
+	events.emit('get-state', uuid, { apiKey: req.apiKey });
 	const config = getConfig(device);
 
 	const userApp = getUserAppForState(device, config);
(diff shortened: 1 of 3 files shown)


src/features/device-state/routes/state-get-v2.ts
```diff
@@ -314,6 +314,8 @@ const stateQuery = _.once(() =>
 
 const getStateV2 = async (req: Request, uuid: string): Promise<StateV2> => {
 	const device = await getDevice(req, uuid);
+	// At this point we are aure that the api key is valid and not expired.
```
Page-, line 317:
```suggestion
	// At this point we are sure that the api key is valid and not expired.
```

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -283,6 +283,8 @@ const stateQuery = _.once(() =>
 
 const getStateV3 = async (req: Request, uuid: string): Promise<StateV3> => {
 	const device = await getDevice(req, uuid);
+	// At this point we are aure that the api key is valid and not expired.
```
Page-, line 286:
```suggestion
	// At this point we are sure that the api key is valid and not expired.
```

thgreasi:
@balena-ci I self-certify!

PR #1144, opened 2022-11-15
Add a Synonymous Form for service & image installs

thgreasi:
Change-type: minor
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/device.20installs.20image.2Fservice.20Synonymous.20Form

### DIFF
--- a/src/balena-model.ts
+++ b/src/balena-model.ts
@@ -207,6 +207,7 @@ export interface Image {
 	content_hash: string | null;
 	contract: {} | null;
 	device__installs__image?: ImageInstall[];
+	is_installed_on__device?: ImageInstall[];
 	is_downloaded_by__device?: GatewayDownload[];
 	image__is_downloaded_by__device?: GatewayDownload[];
 	image__is_part_of__release?: ImageIsPartOfRelease[];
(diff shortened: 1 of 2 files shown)


thgreasi:
@balena-ci I self-certify!

src/balena-model.ts
```diff
@@ -253,7 +254,7 @@ export interface Application {
 	uuid: string;
 	is_public: boolean;
 	application__has__env_var_name?: ApplicationEnvironmentVariable[];
-	application__has__name?: ApplicationConfigVariable[];
+	application__has__config_var_name?: ApplicationConfigVariable[];
```
Page-, line 257: Do we not have a test that verifies this file is up to date?

thgreasi: Looks like we don't. Let me add one while this gets merged.
Moreover I think that the generated property name is not correct either and that abstract-sql-to-typscript should have been using the Term Form when available, right?

thgreasi: Actually, we do have a test in bC, so let me copy it here.

PR #1145, opened 2022-11-15
Add a test that auto-generated model types are up to date

thgreasi:
Change-type: patch


thgreasi:
@balena-ci I self-certify!

PR #1150, opened 2022-11-16
Only fetch data if necessary for patching v3 current device state

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1151, opened 2022-11-16
Fix failed tests not failing CI

thgreasi:


fisehara (approved):
Fixes the issues.

NIT: I don't like the package.json invoking shell script invoking pacakge.json script again.

PR #1153, opened 2022-11-17
Tests: pin redis to 7.x

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1156, opened 2022-11-18
Remove dependent device permissions

Page-:
These are not used in practice and removing them can greatly simplify permission queries

Change-type: major


Page-:
@balena-ci I self-certify!

PR #1157, opened 2022-11-18
Improve check to only handle dependent devices when necessary

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1158, opened 2022-11-21
Update `loki-grpc-client` to version 2

codewithcheese:
Version 2 uses `grpc-js` instead of `grpc`
Moved `loki-grpc-client` from `optionalDependencies` `dependencies` since it should no longer cause install fails

Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/backends/loki.ts
+++ b/src/features/device-logs/lib/backends/loki.ts
@@ -7,10 +7,9 @@ import {
 	createOrgIdMetadata,
 	Direction,
 	EntryAdapter,
-	IGrpcClientAsync,
-	promisifyClient,
 	PusherClient,
 	PushRequest,
+	PushResponse,
 	QuerierClient,
 	QueryRequest,
 	QueryResponse,
@@ -99,7 +98,7 @@ function backoff<T extends (...args: any[]) => any>(
 export class LokiBackend implements DeviceLogsBackend {
 	private subscriptions: EventEmitter;
 	private querier: QuerierClient;
-	private pusher: IGrpcClientAsync;
+	private pusher: PusherClient;
 	private tailCalls: Map<string, ClientReadableStream<TailResponse>>;
 
 	constructor() {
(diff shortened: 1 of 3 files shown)


src/features/device-logs/lib/backends/loki.ts
```diff
@@ -210,11 +206,23 @@ export class LokiBackend implements DeviceLogsBackend {
 		const pushRequest = new PushRequest();
 		pushRequest.setStreamsList(streams);
 		const startAt = Date.now();
-		return this.pusher
-			.push(pushRequest, createOrgIdMetadata(String(appId)), {
-				deadline: Date.now() + PUSH_TIMEOUT,
-			})
-			.finally(() => updateLokiPushDurationHistogram(Date.now() - startAt));
+		return new Promise<any>((resolve, reject) => {
```
Page-, line 209: This `any` should be changed to whatever the type of `response` below is

src/features/device-logs/lib/backends/loki.ts
```diff
@@ -210,11 +206,23 @@ export class LokiBackend implements DeviceLogsBackend {
...
+			this.pusher.push(
+				pushRequest,
+				createOrgIdMetadata(String(appId)),
+				{
+					deadline: Date.now() + PUSH_TIMEOUT,
```
Page-, line 214: It'd make sense to reuse the date we just created to avoid some work:
```suggestion
					deadline: startAt + PUSH_TIMEOUT,
```

src/features/device-logs/lib/backends/loki.ts
```diff
@@ -210,11 +206,23 @@ export class LokiBackend implements DeviceLogsBackend {
...
+				},
+				(err, response) => {
+					updateLokiPushDurationHistogram(Date.now() - startAt);
```
Page-, line 217: Can you move this back into a `.finally` on the promise itself, it's best to have the minimum in a callback so that you get back to promise land ASAP and avoid issues with error handling in callbacks

codewithcheese:
@Page- thanks for the review, good points. I have addressed them, ready for review.

Page- (approved):

fisehara:
@codewithcheese thanks for updating it to use grpc-js. With you changes I clearly see, that I took a total wrong direction when checking grpc-js.

codewithcheese:
@fisehara unfortunately, the GRPC tooling for JS is a bit of a mess and poorly documented, was not straight forward to setup in the first place. What direction did you take?

fisehara:
@codewithcheese I started with exchanging grpc with grpc-js manually which didn't turn out straight forward. Using the updated client didn't come to my mind 🤦

PR #1161, opened 2022-11-21
Make public apps & releases accessible by guests

thgreasi:
This also fixes devices not being able to create
service installs for their supervisor when the
supervisor app's DT doesn't match the
device's.

Discussed as an ad-hoc item in the 2022-12-01 arch call.
See: https://balena.zulipchat.com/#narrow/stream/345748-aspect.2Farchitecture/topic/architecture.20brainstorms

Change-type: patch

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -8,11 +8,20 @@ export = {
 			username: 'guest',
 			password: ' ',
 			permissions: [
+				// core model permissions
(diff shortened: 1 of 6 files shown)


thgreasi:
@balena-ci I self-certify!

thgreasi:
@balena-ci I self-certify!

PR #1164, opened 2022-11-23
Fix device requests to gateway_download failing with 401

thgreasi:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/345889-loop.2Fbalena-os/topic/Python.20SDK.20doesn't.20work.20with.20device.20API.20key


thgreasi:
@balena-ci I self-certify!

PR #1165, opened 2022-11-23
Avoid some unnecessary work in device-heartbeat

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1166, opened 2022-11-23
Separate the image install cache and download progress throttle timers

Page-:
This allows for caching unchanged image install info for longer than the duration of the download progress throttle

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #1167, opened 2022-11-23
Test delay between loki.publish read back.

fisehara:
Running fasttest with parallel `stress --cpu 8 -m 8` will fail on test cases 13 (loki).
With the artificial delays the test cases run successful always

Change-type: patch


fisehara:
Not discovered behavior any more

PR #1171, opened 2022-11-25
Check the model typings as part of the sut

thgreasi:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/348930-product-os.2Fflowzone/topic/.E2.9C.94.20Docker.20jobs
See: https://github.com/balena-io/balena-api/pull/3905


thgreasi:
@balena-ci I self-certify!

PR #1138, opened 2022-11-09
Migration: Normalize release contract fields

joshbwlng:
Change-type: patch
Signed-off-by: Josh Bowling <josh@monarci.com>

PR #1172, opened 2022-11-25
Finalize release contract async migration

joshbwlng:
Change-type: patch
Signed-off-by: Josh Bowling <josh@monarci.com>

### DIFF
--- a/src/migrations/0074-normalize-release-contract.async.ts
+++ b/src/migrations/0074-normalize-release-contract.async.ts
@@ -17,5 +17,5 @@ export = {
 	backoffDelayMS: 4000,
 	errorThreshold: 15,
 	asyncBatchSize: 1000,
-	finalize: false,
+	finalize: true,
 };


fisehara (approved):
LGTM

thgreasi:
I think this should have been a draft PR with a:
```
Depends-on: #1138
```
and we should have only merged it once #1138 was already released to prod & have migrated all records, since atm we have to be careful to first deploy the one, then wait and only then deploy this.

fisehara:
@thgreasi this migration will not affect our prod. It's mentioned in #1138 that it's a two step PR for clarity. But it needs to be in there already.

thgreasi:
Ok, so as discussed this doesn't break anything, but it is confusing imo to have the migration finalized in oB but not in cloud.
For consistency I would only merge this once the cloud API is ready to have the migration finalized as well.

joshbwlng:
@thgreasi Thanks for the input, I'll keep that in mind next time.

PR #1175, opened 2022-11-28
Disable changing device name via the state PATCH

pipex:
While the state endpoint allows patching the device name since v2, that feature hasn't really been used by supervisor until supervisor v13 when that feature was introduced by mistake. Because of a race condition between supervisors v13.0.0 and v14.0.2, the device name may get either an empty value or the value 'local', this can cause confusion in users or affect the user workflows that depend on having distinct device names.

This commit removes the device name from the list of allowed patch variables for the v3 state endpoint.

Change-type: patch


pipex:
The outcome of an internal discussion on this PR is that the patch name feature should be removed and we should add it back on a future `POST` endpoint that will replace the ODATA calls made by the supervisor on provisioning

pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

PR #1176, opened 2022-11-29
Simplify the device host & SV app permissions to rely only on is_public

thgreasi:
Atm `is_host eq 1 or is_public eq 1 === is_public eq 1`, so the is_host check is redundant.
It seem that the `is_host eq 1` is a leftover since before we added `is_public` and adopted its semantics in host app permissions.

Change-type: patch
See: https://github.com/balena-io/open-balena-api/pull/645


thgreasi:
@balena-ci I self-certify!

PR #1177, opened 2022-11-29
Improve service install create permission

thgreasi:


src/lib/auth.ts
```diff
@@ -68,7 +70,7 @@ export const DEVICE_API_KEY_PERMISSIONS = [
 	'resin.device_type.read?describes__device/canAccess()',
 	`resin.device.read?${matchesActor}`,
 	`resin.device.update?${matchesActor}`,
-	'resin.application.read?owns__device/canAccess() or (is_public eq true and is_for__device_type/any(dt:dt/describes__device/canAccess()))',
+	'resin.application.read?owns__device/canAccess() or (is_public eq true and is_for__device_type/canAccess())',
```
thgreasi, line 73: This is needed since this device permission only grants access to the hostApp atm and not the supervisor app (b/c of the `is_for__device_type/any(dt:dt/describes__device...)` part). In the cloud API it atm works b/c public apps are in the guest permissions.

An alternative to this is to add public apps in the guest permissions in oB as well
See: https://github.com/balena-io/open-balena-api/pull/1161

thgreasi: Arch call decided we should do the alt

thgreasi:
@balena-ci I self-certify!

thgreasi:
@balena-ci I self-certify!

PR #1190, opened 2022-12-09
CI: Replace pgsanity with libpg-query

joshbwlng:
Replacing pgsanity as it hasn't been updated in several years and is unable to properly parse JSON operators.

Change-type: patch
Signed-off-by: Josh Bowling <josh@monarci.com>

---

libpg-query repo: https://github.com/pyramation/libpg-query-node
Issue regarding pgsanity's inability to parse JSON operators: https://github.com/markdrago/pgsanity/issues/29

While libpg-query returns a full parse tree for the given SQL statement, I only real care that it throws when it fails to parse the SQL its given. This is similar to what we currently get from pgsanity, making it essentially a drop-in replacement that is more active (updated last month vs pgsanity being updated ~5 years ago) and is able to parse more recent operators etc.

Also adding an extra block to the SQL validity test to check async migration TypeScript files.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,6 @@ FROM balena/open-balena-base:v14.1.0 as runtime
 
 EXPOSE 80
 
-RUN apt update \
-	&& apt install libecpg-dev python3-pip \
-	&& rm -rf /var/lib/apt/lists/* \
-	&& pip3 install --no-cache-dir setuptools \
-	&& pip3 install --no-cache-dir pgsanity
-
 COPY package.json package-lock.json /usr/src/app/
 RUN HUSKY=0 npm ci --unsafe-perm --production && npm cache clean --force
 
--- a/docker-compose.test-custom.yml
+++ b/docker-compose.test-custom.yml
@@ -110,6 +110,7 @@ services:
         - /usr/src/app/node_modules/@mapbox/node-pre-gyp
         - /usr/src/app/node_modules/bcrypt
         - /usr/src/app/node_modules/grpc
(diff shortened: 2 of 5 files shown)


fisehara (commented):
looks real good to me. 
I wonder if we want to move the migraton sanity check into pinejs and give it just the model configuration  That we we may not need it in open-balena-api and balena-api

joshbwlng:
@fisehara I was struggling with that myself. We did something similar in AutumnDB, add testing utilities to the lowest part of the stack that made sense and then have consumers at the higher parts of the stack use the shared utility. Is that what basically what you mean?

fisehara:
@joshbwlng Yes right. Have it in pinejs and let the model / config checked by pinejs. Basically, pinejs will run them, right?

joshbwlng:
@fisehara Yeah, more or less. I'm not sure what that would look like exactly in pinejs, but I'm wondering if that could wait for a later PR? My main goal here is to just replace the outdated pgsanity, and ensure that async migration sql is checked (when possible). I agree that it might make sense to abstract shared testing utilities/methods, but I'd rather not let that block this work unless its super quick and easy to implement in pinejs.

fisehara:
@joshbwlng it's absolutely fine to have it on a later PR and move it down the stack into pinjs later. Just wondered, if there is anything that would block it from this current perspective. 
From my side LGTM

joshbwlng:
@balena-ci rebase

joshbwlng:
@balena-ci I self certify!

PR #1191, opened 2022-12-09
Update @balena/pinejs to 14.53.6

fisehara:
Update @balena/pinejs from 14.53.5 to 14.53.6

Change-type: patch


fisehara:
@balena-ci I self-certify!

PR #1194, opened 2022-12-13
Fix the autogenerated constraint names

thgreasi:
Connects-to: #1188
Change-type: patch
See: https://github.com/balena-io/balena-api/pull/3937


thgreasi:
@balena-ci I self-certify!

PR #1195, opened 2022-12-13
Optimize the device should be running release rule

thgreasi:
[From 1.2s](https://explain.dalibo.com/plan/82c78c9f6a2d280f)
[To 293ms](https://explain.dalibo.com/plan/846h39c73a9a828d)

Apparently PG can now (after the COUNT(*) =0 change or after the PG 14 bump) reason that we are joining the `device.should_be_running__release` with the `release.id`s and only does one `FROM release` (since there is no `release2` reference in the execution plan):
```sql
SELECT (
	SELECT COUNT(*)
	FROM "device" AS "device.0",
		"release" AS "release.1"
	WHERE "device.0"."should be running-release" = "release.1"."id"
	AND NOT EXISTS (
		SELECT 1
		FROM "release" AS "release.2",
			"application" AS "application.4"
		WHERE "device.0"."belongs to-application" = "application.4"."id"
		AND "release.2"."belongs to-application" = "application.4"."id"
		AND "release.2"."status" = 'status'
		AND "release.2"."status" IS NOT NULL
		AND "device.0"."should be running-release" = "release.2"."id"
	)
) = 0 AS "result";
```
this was not the case when I originally optimized this rule.

Change-type: patch


thgreasi:
Ooops, I used `status` rather than `success`...

PR #1198, opened 2022-12-14
Allow device API keys to delete their own device_tags

thgreasi:
Change-type: minor
See: https://balena.zulipchat.com/#narrow/stream/346010-channel.2Fsupport/topic/Deleting.20Tags.20from.20Deivce


thgreasi:
@balena-ci I self-certify!

PR #1199, opened 2022-12-14
Add configuration variables for HDMI port 1

pipex:
Raspberry PI 4 has a second which can be configured adding the suffix :1 to configurations in config.txt. This commit enables those variables on the API.

Change-type: patch

### DIFF
--- a/src/features/vars-schema/env-vars.ts
+++ b/src/features/vars-schema/env-vars.ts
@@ -56,7 +56,8 @@ export const BLOCKED_NAMES = addReservedPrefixes([
 	'HOST_LOG_TO_DISPLAY',
 ]);
 
-export const INVALID_CHARACTER_REGEX = /^\d|\W/;
+const INVALID_ALPHANUMERIC_REGEX = /^\d|\W/;
+export const INVALID_CHARACTER_REGEX = /^[\d:]|[^A-Za-z0-9_:]/;
 export const INVALID_NEWLINE_REGEX = /\r|\n/;
 
 const getDefinitionWithMinimumSupervisorVersion = (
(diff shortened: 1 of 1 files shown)


pipex:
@balena-ci rebase

src/features/vars-schema/env-vars.ts
```diff
@@ -343,6 +343,38 @@ export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
...
+	...getDefinitionWithMinimumSupervisorVersion(
+		{
+			'14.4.7': [
+				'raspberrypi4-64',
+				'raspberrypi400-64',
+				'raspberrypicm4-ioboard',
+			],
+		},
+		{
+			'BALENA_HOST_CONFIG_hdmi_force_hotplug:1': {
```
thgreasi, line 355: Have you tested this to confirm whether any part of the stack might freak out by the `:`?

pipex: I am working on this, the supervisor freaks out, but I will create a new version that accepts it, hence the limiting by version. I need to see if the dashboard will accept the change.

pipex: Ok, just tested, I need to change the `device_config_variable` resource, because I'm getting `Error!Request error: Configuration variable names can only contain alphanumeric characters and underscores.`

src/features/vars-schema/env-vars.ts
```diff
@@ -343,14 +343,50 @@ export const DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES: Array<{
...
-const checkVarName = (type: string, name: string) => {
-	if (INVALID_CHARACTER_REGEX.test(name)) {
+const checkVarName = (
+	type: string,
+	name: string,
+	invalidChars = INVALID_CHARACTER_REGEX,
+) => {
+	if (invalidChars.test(name)) {
 		throw new BadRequestError(
 			`${type} names can only contain alphanumeric characters and underscores.`,
```
Page-, line 391: If we change the regex then we also need a different error message

src/features/vars-schema/env-vars.ts
```diff
@@ -363,6 +403,14 @@ const checkVarName = (type: string, name: string) => {
 	}
 };
 
+const checkAlphaNumeric = (type: string, name: string) => {
+	if (/^\d|\W/.test(name)) {
```
Page-, line 407: In-lining just this one case feels weird. I'd say either declare it like `INVALID_CHARACTER_REGEX` or in-line both and have their names document the regex purpose

src/features/vars-schema/env-vars.ts
```diff
@@ -56,7 +56,7 @@ export const BLOCKED_NAMES = addReservedPrefixes([
 	'HOST_LOG_TO_DISPLAY',
 ]);
 
-export const INVALID_CHARACTER_REGEX = /^\d|\W/;
+export const INVALID_CHARACTER_REGEX = /^(\d|:)|[^A-Za-z0-9_:]/;
```
Page-, line 59: It's best to avoid the capturing group as it's unnecessary, either with `[...]`
```suggestion
export const INVALID_CHARACTER_REGEX = /^[\d:]|[^A-Za-z0-9_:]/;
```
or mark it as a non-capturing group `(?:...)`
```suggestion
export const INVALID_CHARACTER_REGEX = /^(?:\d|:)|[^A-Za-z0-9_:]/;
```

pipex:
Keeping this as draft as I need to update the supervisor version once this merges https://github.com/balena-os/balena-supervisor/pull/2090

thgreasi (approved):

src/features/vars-schema/env-vars.ts
```diff
@@ -56,7 +56,8 @@ export const BLOCKED_NAMES = addReservedPrefixes([
 	'HOST_LOG_TO_DISPLAY',
 ]);
 
-export const INVALID_CHARACTER_REGEX = /^\d|\W/;
+const INVALID_ALPHANUMERIC_REGEX = /^\d|\W/;
+export const INVALID_CHARACTER_REGEX = /^[\d:]|[^A-Za-z0-9_:]/;
```
Page-, line 60: Is this going to mess up the UI @thgreasi since it uses it via the `/config/vars` `invalidRegex` property

thgreasi: Oh Indeed @Page- we need to export the new regex as well and update `/config/vars` to preserve the old key and add a new to define the config var regex :facepalm:

pipex: I don't understand, `/config/vars` validates the configuration variable format no? Then I think the export is ok. Moreover, I tested the UI with this new OB version and without this export the configuration menu failed to set the right values

PR #1205, opened 2022-12-21
Unify API key handling across `retrieveAPIKey` and middleware

Page-:
This unifies the code to avoid duplicate work in the case of mixing and matching the two methods where you may have prefetched and api key but then end up doing a full fetch elsewhere anyway

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1206, opened 2022-12-21
Avoid calling `retrieveAPIKey` when possible in `getUser`

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1210, opened 2022-12-23
Auth middleware

Page-:


src/infra/auth/middleware.ts
```diff
@@ -22,7 +25,15 @@ export const authenticatedMiddleware: RequestHandler = async (
...
+ * Note: This is specifically *user* credentials, device keys/application provisioning keys do not count
+ */
+export const fullyAuthenticatedUser: RequestHandler = async (
+	req,
+	res,
+	next,
+) => {
 	try {
 		await getUser(req, undefined);
 		next();
```
thgreasi, line 39: Should we also remove this `return null;`?

thgreasi (approved):

PR #1213, opened 2022-12-27
CI: Fix action config typos

joshbwlng:
Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #1214, opened 2022-12-29
Tests: fix composition typings/usage

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1222, opened 2023-01-03
Use authenticatedApiKey middleware on some endpoints that require one

Page-:
Change-type: patch

### DIFF
--- a/src/infra/auth/middleware.ts
+++ b/src/infra/auth/middleware.ts
@@ -95,6 +95,23 @@ export const authenticatedApiKey: RequestHandler = async (req, res, next) => {
 };
 
 /**
+ * This ensures that valid credentials have been passed, , returning 401 if they have not
+ */
+export const authenticated: RequestHandler = async (req, res, next) => {
+	try {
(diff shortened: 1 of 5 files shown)


Page-:
@balena-ci I self-certify!

src/infra/auth/middleware.ts
```diff
@@ -94,6 +94,23 @@ export const authenticatedApiKey: RequestHandler = async (req, res, next) => {
 	}
 };
 
+/**
+ * This ensures that valid credentials have been passed, , returning 401 if they have not
+ */
+export const authenticated: RequestHandler = async (req, res, next) => {
+	try {
+		await getUser(req, undefined);
```
thgreasi, line 102: @Page- Isn't this going to slow down the device state GET requests when originating from devices?
Afaict this will fire run `getUserQuery()` for each request, which in the majority of the cases will be failing b/c it will be a request from a device, and then it would continue in the `catch` part exactly how it was doing before.

Is this going to be a benefit just for user originating requests that will now be failing sooner than later?

PR #1223, opened 2023-01-03
Include id in "image"."is a build of-service" index

Page-:
Change-type: patch


thgreasi (commented):
LGTM :+1: @Page- 
But I'm now curious what made you consider this addition :thinking: 
I can see it will benefit this v query, and expect will give more choices to the planner, but was there anything more specific?
See: https://github.com/balena-io/open-balena-api/blob/79169abd67562004d56f6767a4fca3c95df4a762/src/features/ci-cd/hooks/service-installs.ts#L20-L51

Page-:
@balena-ci I self-certify!

PR #1230, opened 2023-01-17
Use `authenticatedApiKey` middleware for device logs endpoints

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1234, opened 2023-01-25
Update dependencies

thgreasi:
Update @balena/pinejs from 14.54.7 to 14.57.2
Update @balena/abstract-sql-compiler from 7.24.0 to 7.25.3

Change-type: patch


thgreasi:
I self-certify!

PR #1238, opened 2023-01-26
Send registry image pulls metrics to prometheus

otaviojacobi:
Change-type: patch

### DIFF
--- a/src/features/registry/metrics.ts
+++ b/src/features/registry/metrics.ts
@@ -1,0 +1,13 @@
+import { metrics } from '@balena/node-metrics-gatherer';
+
+metrics.describe.counter(
+	'api_registry_image_pull_total',
+	'Number of image pulls getting the token via API',
+);
+
+export function incrementRegistryImagePulls(
+	value: number = 1,
+	payload: any = {},
+) {
+	metrics.counter('api_registry_image_pull_total', value, payload);
+}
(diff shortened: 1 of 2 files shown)


src/features/registry/metrics.ts
```diff
@@ -0,0 +1,13 @@
+import { metrics } from '@balena/node-metrics-gatherer';
+
+metrics.describe.counter(
+	'api_registry_image_pull_total',
+	'Number of image pulls getting the token via API',
+);
+
+export function incrementRegistryImagePulls(
+	value: number = 1,
+	payload: any = {},
```
Page-, line 10: Can you type this better than `any` please

src/features/registry/metrics.ts
```diff
@@ -0,0 +1,13 @@
+import { metrics } from '@balena/node-metrics-gatherer';
+
+metrics.describe.counter(
+	'api_registry_image_pull_total',
+	'Number of image pulls getting the token via API',
+);
+
+export function incrementRegistryImagePulls(
+	value: number = 1,
```
Page-, line 9: When you specify a default value then the type defaults to the type of that default value:
```suggestion
	value = 1,
```
There's nothing wrong with explicitly specifying `number` but it's redundant and makes sense imo to reserve for only the cases where you want a type that is different to that of the default value

src/features/registry/registry.ts
```diff
@@ -591,6 +592,17 @@ export const token: RequestHandler = async (req, res) => {
 					authorizeRequest(req, scopes, tx),
 				]),
 		);
+
+		access.forEach((ac) =>
+			incrementRegistryImagePulls(1, {
```
Page-, line 597: I would name the metric differently to show that it is images that have had a token validated for access rather than image pulls (you can use the image to pull 0 or 2+ times as well so it's not 1-1)

PR #1243, opened 2023-02-02
Reduce "as Writable" casts with "satisfies"

thgreasi:
Change-type: patch


thgreasi:
I self-certify!

src/infra/auth/auth.ts
```diff
@@ -149,15 +149,15 @@ export const userFields = [
 	'email',
 	'created_at',
 	'jwt_secret',
-] as const;
```
thgreasi, line 152: This `as const` was also making this a ReadonlyArray, which required consumers to do some extra casts in order to get TS to allow them to push more fields that they might want to be populated in the JWT.

thgreasi:
I self-certify!

PR #1245, opened 2023-02-02
Listen for the correct message type when using sharded pub/sub

Page-:
Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/backends/redis.ts
+++ b/src/features/device-logs/lib/backends/redis.ts
@@ -28,6 +28,7 @@ const UNSUBSCRIBECMD = REDIS_LOGS_SHARDED_PUBSUB
 	? 'sunsubscribe'
 	: 'unsubscribe';
 const PUBLISHCMD = REDIS_LOGS_SHARDED_PUBSUB ? 'spublish' : 'publish';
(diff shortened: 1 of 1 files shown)


Page-:
I self-certify!

src/features/device-logs/lib/backends/redis.ts
```diff
@@ -28,6 +28,7 @@ const UNSUBSCRIBECMD = REDIS_LOGS_SHARDED_PUBSUB
 	? 'sunsubscribe'
 	: 'unsubscribe';
 const PUBLISHCMD = REDIS_LOGS_SHARDED_PUBSUB ? 'spublish' : 'publish';
+const MESSAGECMD = REDIS_LOGS_SHARDED_PUBSUB ? 'smessage' : 'message';
```
thgreasi, line 31: Where did you find that :sparkles: ?
The only place I could find is mentioned but w/o any docs was v
See: https://redis.io/commands/ssubscribe/

Page-: I verified it there but I found it based upon the ioredis PR that added support (it added support for emitting an `smessage` command which raised the question)

PR #1251, opened 2023-02-06
Fix env-var schema for backwards compatibility

pipex:
PR #1199 relaxed the allowed character set for configuration variables, but it also inadvertently relaxed the `/config/vars` JSON schema exported by the API. This would cause that the UI allows the extended character set for env vars, but the API would ultimately reject it, creating an inconsistency. This PR adds a `configVarInvalidRegex` variable to the schema that should be adopted by the UI to render configuration variables.

Change-type: patch

### DIFF
--- a/src/features/vars-schema/schema.ts
+++ b/src/features/vars-schema/schema.ts
@@ -4,7 +4,8 @@ import type { JSONSchema6 } from 'json-schema';
 import {
 	BLOCKED_NAMES,
 	DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES,
-	INVALID_CHARACTER_REGEX,
+	INVALID_ENV_VAR_REGEX,
+	INVALID_CONFIG_VAR_REGEX,
 	RESERVED_NAMES,
 	RESERVED_NAMESPACES,
 	SUPERVISOR_CONFIG_VAR_PROPERTIES,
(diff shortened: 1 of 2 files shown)


src/features/vars-schema/schema.ts
```diff
@@ -31,11 +32,12 @@ export const schema: RequestHandler = (req, res) => {
 	const varsConfig = {
 		reservedNames: RESERVED_NAMES,
 		reservedNamespaces: RESERVED_NAMESPACES,
-		invalidRegex: INVALID_CHARACTER_REGEX.toString(),
+		invalidRegex: INVALID_ENV_VAR_REGEX.toString(),
 		whiteListedNames: ALLOWED_NAMES,
 		whiteListedNamespaces: ALLOWED_NAMESPACES,
 		blackListedNames: BLOCKED_NAMES,
 		configVarSchema,
+		configVarInvalidRegex: INVALID_CONFIG_VAR_REGEX.toString(),
```
Page-, line 40: I'd keep them next to each other as it makes sense to group them imo:
```suggestion
		invalidRegex: INVALID_ENV_VAR_REGEX.toString(),
		configVarInvalidRegex: INVALID_CONFIG_VAR_REGEX.toString(),
		whiteListedNames: ALLOWED_NAMES,
		whiteListedNamespaces: ALLOWED_NAMESPACES,
		blackListedNames: BLOCKED_NAMES,
		configVarSchema,
```

thgreasi (approved):

PR #1253, opened 2023-02-09
Fixes fleet default state releaseUuid query parameter

fisehara:
Change-type: patch


fisehara:
I self-certify

thgreasi (commented):
:+1:

PR #1254, opened 2023-02-10
Allow public applications be of a legacy application type

thgreasi:
Moving the rule to bC to make grandfathering
application types easier.

Change-type: patch


thgreasi:
I self-certify!

PR #1256, opened 2023-02-15
Fix cpu_id reported either in lower / upper case

fisehara:
Force cpu_id to lower case

Change-type: patch

### DIFF
--- a/src/features/device-state/routes/state-patch-v2.ts
+++ b/src/features/device-state/routes/state-patch-v2.ts
@@ -275,6 +275,9 @@ export const statePatchV2: RequestHandler = async (req, res) => {
 						deviceBody = truncateShortTextFields(deviceBody);
 						// If we're updating anyway then ensure the metrics data is included
 						deviceBody = { ...deviceBody, ...metricsBody };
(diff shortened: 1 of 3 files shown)


src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -275,6 +275,7 @@ export const statePatchV2: RequestHandler = async (req, res) => {
 						deviceBody = truncateShortTextFields(deviceBody);
 						// If we're updating anyway then ensure the metrics data is included
 						deviceBody = { ...deviceBody, ...metricsBody };
+						deviceBody.cpu_id = deviceBody.cpu_id?.toLowerCase();
```
Page-, line 278: I think you should only do the assignment if the property exists, otherwise you'll be assigning undefined and that could cause funky behavior:
```suggestion
						if (deviceBody.cpu_id != null) {
							deviceBody.cpu_id = deviceBody.cpu_id.toLowerCase();
						}
```
(and the same for v3)

fisehara:
LGTM

PR #1259, opened 2023-02-16
Allow device API Keys to GET their logs

thgreasi:
Change-type: minor
See: https://balena.zulipchat.com/#narrow/stream/345882-_help/topic/resources.20accessible.20via.20the.20deviceApiKey


thgreasi:
Closing since the device should anyway be able to access the logs directly from the container, so there is no reason to read them back from the cloud.

PR #1272, opened 2023-03-03
Block deleting releases that are tracked by devices or apps

myarmolinsky:
Block deleting releases that are tracked by devices or apps

Change-type: minor

### DIFF
--- a/src/features/releases/hooks/restrict-release-deletion.ts
+++ b/src/features/releases/hooks/restrict-release-deletion.ts
@@ -1,0 +1,68 @@
+import { sbvrUtils, hooks, errors } from '@balena/pinejs';
+import { Application, Device, PickExpanded } from '../../../balena-model';
+const { BadRequestError } = errors;
+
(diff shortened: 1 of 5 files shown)


thgreasi:
Maybe better prhase it like
> Block deleting releases that are tracked by devices or apps

src/features/releases/hooks/restrict-release-deletion.ts
```diff
@@ -0,0 +1,65 @@
...
+			throw new BadRequestError(
+				`Unable to delete release ${applicationPinnedToRelease.should_be_running__release[0].raw_version} because it is the ${applicationPinnedToRelease.is_of__class}'s target release.`,
+			);
+		}
+
+		const [devicesPinnedToRelease] = await api.get({
+			resource: 'device',
+			options: {
+				$top: 1,
+				$select: 'uuid',
```
Page-, line 43: If you only care for existence then it's likely better to select only the primary key, ie
```suggestion
				$select: 'id',
```

PR #1273, opened 2023-03-06
Block device api key permissions of frozen devices

thgreasi:
Change-type: minor
See: https://github.com/balena-io/balena-api/issues/3775
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/Block.20permissions.20of.20frozen.20devices

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -6,6 +6,7 @@ import { sbvrUtils } from '@balena/pinejs';
 
 import {
 	API_VPN_SERVICE_API_KEY,
+	IGNORE_FROZEN_DEVICE_PERMISSIONS,
 	VPN_GUEST_API_KEY,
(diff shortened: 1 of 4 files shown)


src/lib/auth.ts
```diff
@@ -20,7 +21,18 @@ const writePerms = (
 
 const matchesActor = 'actor eq @__ACTOR_ID';
 const matchesUser = `user/any(u:u/${matchesActor})`;
-const ownsDevice = `owns__device/any(d:d/${matchesActor})`;
+const matchesNonFrozenDeviceActor = (alias: string) => {
```
Page-, line 24: You could default to no alias, eg
```suggestion
const matchesNonFrozenDeviceActor = (alias = '') => {
```

Page- (approved):

PR #1280, opened 2023-03-08
Cache the entirety of the device logs write context

Page-:
This will have more impact than caching only the permissions checking part.

Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/store.ts
+++ b/src/features/device-logs/lib/store.ts
@@ -39,7 +39,6 @@ import {
 } from '../../../infra/cache';
 
 const {
-	NotFoundError,
 	UnauthorizedError,
 	ServiceUnavailableError,
 	UnsupportedMediaTypeError,
@@ -48,30 +47,7 @@ const { api } = sbvrUtils;
 
 const supervisor = new Supervisor();
 
-const getWriteContext = async (req: Request): Promise<LogContext> => {
-	const { uuid } = req.params;
-	return await sbvrUtils.db.readTransaction(async (tx) => {
-		const resinApi = api.resin.clone({ passthrough: { req, tx } });
-		const device = (await resinApi.get({
-			resource: 'device',
-			id: { uuid },
-			options: {
-				$select: ['id', 'belongs_to__application'],
-			},
-		})) as PickDeferred<Device, 'id' | 'belongs_to__application'> | undefined;
-		if (!device) {
-			throw new NotFoundError('No device with uuid ' + uuid);
-		}
-		await checkDeviceLogsWritePermissions(device, req, tx);
-		return addRetentionLimit<LogContext>({
-			id: device.id,
-			belongs_to__application: device.belongs_to__application!.__id,
-			uuid,
-		});
-	});
-};
-
-const checkDeviceLogsWritePermissions = (() => {
+const getWriteContext = (() => {
 	const authQuery = _.once(() =>
 		api.resin.prepare<{ id: number }>({
 			method: 'POST',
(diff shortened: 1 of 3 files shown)


src/features/device-logs/lib/store.ts
```diff
@@ -48,68 +47,72 @@ const { api } = sbvrUtils;
...
+				return addRetentionLimit({
+					id: device.id,
+					belongs_to__application: device.belongs_to__application!.__id,
+					uuid,
+				});
+			});
 		},
 		{
-			cacheKey: 'checkDeviceLogsWritePermissions',
+			cacheKey: 'getDeviceLogsWriteContext',
```
Page-, line 100: The cache key needs to change because the contents/semantics have changed, but it makes sense regardless imo

src/features/device-logs/lib/store.ts
```diff
@@ -48,68 +47,72 @@ const { api } = sbvrUtils;
...
+						$select: ['id', 'belongs_to__application'],
+					},
+				})) as
+					| PickDeferred<Device, 'id' | 'belongs_to__application'>
+					| undefined;
+				if (!device) {
+					return false;
+				}
+				if (!(await hasDeviceLogsWritePermissions(device, req, tx))) {
+					return false;
```
Page-, line 90: We return false for these and throw the error outside the cache because thrown errors are not cached and we want these failures to be cached (as they're expected errors/behavior rather than exceptions)

thgreasi (approved):
:sparkles:

src/features/device-logs/lib/store.ts
fisehara: This makes sense, although I think merging missing Permissions and device not found into `UnauthorizedError()` will make debugging later more tricky.

src/features/device-logs/lib/store.ts
```diff
@@ -48,68 +47,72 @@ const { api } = sbvrUtils;
...
+					resource: 'device',
+					id: { uuid },
+					options: {
+						$select: ['id', 'belongs_to__application'],
+					},
+				})) as
+					| PickDeferred<Device, 'id' | 'belongs_to__application'>
+					| undefined;
+				if (!device) {
+					return false;
```
fisehara, line 87:
```suggestion
					return WriteContextStatus.deviceNotFound;
```

src/features/device-logs/lib/store.ts
fisehara, line 90:
```suggestion
					return WriteContextStatus.unauthorized;
```

src/features/device-logs/lib/store.ts
```diff
@@ -48,68 +47,72 @@ const { api } = sbvrUtils;
...
-	return async (
-		ctx: { id: number },
-		req: permissions.PermissionReq,
-		tx: Tx,
-	) => {
-		if (!(await hasWritePermissions(ctx.id, req, tx))) {
+	return async (req: Request): Promise<LogContext> => {
+		const { uuid } = req.params;
+		const maybeLogContext = await $getWriteContext(uuid, req);
+		if (maybeLogContext === false) {
```
fisehara, line 112:
```suggestion
		if (maybeLogContext === WriteContextStatus.unauthorized;) {
```

fisehara: And the other thrown error would be:
```
		if (maybeLogContext === WriteContextStatus.DeviceNotFound) {	
			throw new NotFoundError('No device with uuid ' + uuid);	
		}
```

PR #1283, opened 2023-03-09
Remove duplicate is_frozen check from service install insert permission

thgreasi:
Change-type: patch


thgreasi:
I self-certify!

PR #1288, opened 2023-03-09
Add optional LOGS_HOST env var for log-stream subdomain

cywang117:
This PR makes it so that newly generated config.jsons include `logsEndpoint`, if the LOGS_HOST env var is present. It doesn't need to wait until LOGS_HOST is deployed in balena's cloud environments, since if LOGS_HOST isn't present, a config.json without `logsEndpoint` will be generated.

This PR doesn't add support for migration of existing devices to use the logs subdomain; that is planned to be added in a future feature.

Change-type: patch


docker-compose.test-custom.yml
```diff
@@ -61,6 +61,7 @@ services:
       IMAGE_STORAGE_SECRET_KEY: SECRET_KEY
       JSON_WEB_TOKEN_EXPIRY_MINUTES: 10080
       JSON_WEB_TOKEN_SECRET: purple
+      LOGS_HOST: logs_host.com
```
Page-, line 64: It'd be better to use
```suggestion
      LOGS_HOST: logs.balenadev.io
```
as it's a domain we control and less likely to cause accidental weirdness if we reach out to a random domain for testing (alternatively I think `*.example.com` is also safe?)

Page- (changes requested):
`LOGS_HOST` isn't actually used in the code anywhere? I assume it needs to be added to the device-config generation?

docker-compose.test-custom.yml
cywang117: Changed to `logs.balenadev.io` in https://github.com/balena-io/open-balena-api/commit/6aa95b8696218eedaa5b1576948dafbd8887c9d9 👍🏼

cywang117:
This is ready for re-review! This only adds `logsEndpoint` to newly generated config.jsons if LOGS_HOST env var exists; migration of existing devices would be a separate feature. After this PR is merged, I'll be adding the LOGS_HOST env var to staging and making sure everything is good, before adding it to production.

src/lib/config.ts
```diff
@@ -130,6 +130,7 @@ export const JSON_WEB_TOKEN_EXPIRY_MINUTES = intVar(
 	'JSON_WEB_TOKEN_EXPIRY_MINUTES',
 );
 export const JSON_WEB_TOKEN_SECRET = requiredVar('JSON_WEB_TOKEN_SECRET');
+export const LOGS_HOST = optionalVar('LOGS_HOST', '');
```
joshbwlng, line 133:
```suggestion
export const LOGS_HOST = optionalVar('LOGS_HOST');
```
Nit - probably don't need to set an empty default value.

src/features/device-config/device-config.ts
```diff
@@ -120,6 +121,8 @@ export const generateConfig = async (
 				delta: `https://${DELTA_HOST}`,
 				registry: REGISTRY2_HOST,
 				vpn: VPN_HOST,
+				// If undefined, balena-device-config won't generate a logs endpoint
+				logs: LOGS_HOST ? `https://${LOGS_HOST}` : undefined,
```
thgreasi, line 125: NIT: I guess it could even be v so that we don't set empty properties at all (even though express might be removing them during serialization anyway):
```suggestion
				...(LOGS_HOST != null && { logs: `https://${LOGS_HOST}` },
```

thgreasi (commented):
Other than that & Josh's comment, LGTM :+1:

src/features/device-config/device-config.ts
Page-: `undefined` cannot be represented in JSON and so would be removed altogether

Page-, line 125: If you want to check for nullish then I prefer to be explicit about it (it can avoid confusion later):
```suggestion
				logs: LOGS_HOST != null ? `https://${LOGS_HOST}` : undefined,
```

cywang117: Updated in https://github.com/balena-io/open-balena-api/commit/006d7e85a29fab30f5e44589e0f6448c6e4a6a82 👍🏼

cywang117:
I tested this on BoB -- was not able to get the HAProxy set up as I'm not sure how to translate the kubernetes config for logs from our staging/production environments into an HAProxy conf, but I did add `LOGS_HOST` as an env var in balena-cloud/BoB's `docker-compose.yml`, and was able to see `logsEndpoint` generated as a config using `balena config generate`. When I removed `LOGS_HOST` from docker-compose.yml with this patch, I no longer saw `logsEndpoint` get generated, therefore this change is backwards compatible. @joshbwlng

cywang117:
If no further comments, I'll self-certify 24h from time of writing this message 👀

cywang117:
@balena-ci I self-certify!

PR #1293, opened 2023-03-12
Adds webresource typing

otaviojacobi:


src/balena-model.ts
```diff
@@ -11,6 +11,13 @@ export type Deferred<T> = Exclude<T, any[]>;
 export type PickDeferred<T, K extends keyof T> = {
 	[P in K]: Deferred<T[P]>;
 };
+export interface WebResource {
```
thgreasi, line 14: Is this type autogenerated like the rest in this file?

src/balena.sbvr
```diff
@@ -609,6 +612,11 @@ Fact type: device type is of cpu architecture
 Fact type: device type has logo
 	Necessity: each device type has at most one logo
 
+Fact type: device type has logo image
+	Necessity: each device type has at most one logo image
+	Necessity: each logo image is of exactly one device type
```
thgreasi, line 617: I this necessary?

thgreasi (commented):
I'm not sure about the field name that we should use, but other than that it looks nice so far @**otaviojacobi**  :+1: 
(maybe we could even do a major and make the current `logo` a webresource and rename the current base64 field :thinking: )
The only missing part (and probably the one requiring the most effort) is updating the import func in contracts/index.ts to populate the new fields.

src/balena-model.ts
otaviojacobi: Yes :)

otaviojacobi:
LGTM

otaviojacobi:
LGTM

PR #1294, opened 2023-03-13
Remove `check-device-releases-before-delete` hook

myarmolinsky:
Remove `check-device-releases-before-delete` hook

Depends-on: https://github.com/balena-io/open-balena-api/pull/1290
Changelog-entry: Allow deleting applications when a device is running its release in another application
Change-type: minor


test/14_release-pinning.ts
```diff
@@ -39,6 +40,24 @@ describe(`Tracking latest release`, () => {
 			passthrough: { user: admin },
 		});
 
+		const org = await fx.organizations.admin;
+		const deviceType: DeviceType = (
+			await pineUser.get({
+				resource: 'device_type',
+				options: { $select: 'id', $filter: { slug: 'intel-nuc' } },
+			})
+		).body[0];
```
thgreasi, line 49:
```suggestion
		const { body: deviceType } = await pineUser.get<DeviceType>({
			resource: 'device_type',
			id: { slug: 'intel-nuc' },
			options: { $select: 'id' },
		}).expect(200);
```

test/14_release-pinning.ts
```diff
@@ -39,6 +40,24 @@ describe(`Tracking latest release`, () => {
...
+		applicationToDelete = await fixtures.createResource({
+			resource: 'application',
+			method: 'POST',
+			body: {
+				app_name: 'application_to_delete',
+				organization: org.id,
+				is_for__device_type: deviceType.id,
+			},
+			user: { token: admin.token! },
+		});
```
thgreasi, line 59: Isn't this v working?
```suggestion
		({ body: applicationToDelete } = await pineUser.post({
			resource: 'application',
			body: {
				app_name: 'application_to_delete',
				organization: org.id,
				is_for__device_type: deviceType.id,
			}
		}).expect(200));
```

test/14_release-pinning.ts
```diff
@@ -343,7 +363,7 @@ describe(`Tracking latest release`, () => {
 			});
 			app4ReleaseId = app4Release.id;
 			const app5Release = await addReleaseToApp(admin, {
-				belongs_to__application: application3Id,
+				belongs_to__application: applicationToDelete.id,
```
thgreasi, line 366: I would prefer it if we didn't touch pre-existing releases, since this could influence whether existing test cases actually test what they should any more.
Let's just create new releases.

test/14_release-pinning.ts
```diff
@@ -310,6 +329,7 @@ describe(`Tracking latest release`, () => {
 		let app3ReleaseId: number;
 		let app4ReleaseId: number;
 		let app5ReleaseId: number;
+		let app6ReleaseId: number;
```
thgreasi, line 332: v ?
```suggestion
		let appToDeleteRelease1Id: number;
```

test/14_release-pinning.ts
```diff
@@ -546,5 +585,25 @@ describe(`Tracking latest release`, () => {
...
+					},
+				})
+				.expect(200);
+			await pineUser.patch({
+				resource: 'device',
+				id: device4.id,
+				body: { belongs_to__application: application3Id },
+			});
+			await pineUser
+				.delete({ resource: 'application', id: applicationToDelete.id })
```
thgreasi, line 605: `device4` was not in `applicationToDelete` though before.
How about provisioning a new one right here in the correct app?
```
			const device = await fakeDevice.provisionDevice(admin, applicationToDelete);
			await pineUser
				.patch({
					resource: 'device',
					id: device4.id,
					body: {
						is_running__release: app6ReleaseId,
					},
				})
				.expect(200);
			// ...
```

test/14_release-pinning.ts
myarmolinsky: Not exactly typing-wise. For this to work, I need to destructure into a const and then assign the value to `applicationToDelete`

thgreasi (approved):

thgreasi:
Btw @myarmolinsky  can you add a `Depends-on` and link this PR to the `informative` PR that you did so that this includes more context?

PR #1297, opened 2023-03-21
Fix unintentionally marking image installs as deleted

Page-:
Change-type: patch


Page-:
This isn't actually an issue, I mistook `imageIds` as `imgInstallIds`

PR #1301, opened 2023-03-29
Track scheduled job runs in the model

thgreasi:
Change-type: minor
See: https://github.com/balena-io/balena-api/pull/4348#discussion_r1151560771


thgreasi:
I self-certify!

PR #1302, opened 2023-03-30
Enforce that for a release to be marked "successful" all its images must have been successfully built

dfunckt:
This rule should exist since multi-container but apparently doesn't.

```sql
-- It is necessary that each image that is part of a release that has a status that equals "success", has a status that equals "success".
SELECT (
	SELECT COUNT(*)
	FROM "image" AS "image.0",
		"release" AS "release.1",
		"image-is part of-release" AS "image.0-is part of-release.1"
	WHERE $1 = "release.1"."status"
	AND "release.1"."status" IS NOT NULL
	AND "image.0-is part of-release.1"."image" = "image.0"."id"
	AND "image.0-is part of-release.1"."is part of-release" = "release.1"."id"
	AND NOT (
		$2 = "image.0"."status"
		AND "image.0"."status" IS NOT NULL
	)
	AND ($3 = '{}'
	OR "image.0"."id" = ANY(CAST($3 AS INTEGER[])))
) = 0 AS "result";
```

See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/.22successful.22.20release.20images

### DIFF
--- a/src/migrations/00082-ensure-successful-releases-have-successful-images.sql
+++ b/src/migrations/00082-ensure-successful-releases-have-successful-images.sql
@@ -1,0 +1,47 @@
+UPDATE "release" SET "status" = 'failed' WHERE "id" IN (
+  SELECT "release.1"."id"
+  FROM "image" AS "image.0",
+    "release" AS "release.1",
+    "image-is part of-release" AS "image.0-is part of-release.1"
+  WHERE 'success' = "release.1"."status"
+  AND "release.1"."status" IS NOT NULL
+  AND "image.0-is part of-release.1"."image" = "image.0"."id"
+  AND "image.0-is part of-release.1"."is part of-release" = "release.1"."id"
+  AND NOT (
+    'success' = "image.0"."status"
+    AND "image.0"."status" IS NOT NULL
+  )
+  -- As a precaution, exclude releases that `should be-running` on any device,
+  -- as it's invalid for non-successful releases to be running on devices.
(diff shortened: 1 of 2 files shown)


src/migrations/00082-ensure-successful-releases-have-successful-images.sql
```diff
@@ -0,0 +1,21 @@
...
+      "image-is part of-release" AS "image.0-is part of-release.1"
+    WHERE 'success' = "release.1"."status"
+    AND "release.1"."status" IS NOT NULL
+    AND "image.0-is part of-release.1"."image" = "image.0"."id"
+    AND "image.0-is part of-release.1"."is part of-release" = "release.1"."id"
+    AND NOT (
+      'success' = "image.0"."status"
+      AND "image.0"."status" IS NOT NULL
+    )
+  ) > 0
```
thgreasi, line 17: Heads-up that the query part is using spaces while the rest of the file is using tabs.

thgreasi (approved):
Not sure if you want to also add an update to fix the violating rows here or only in bC but other than that LGTM :+1:

src/migrations/00082-ensure-successful-releases-have-successful-images.sql
thgreasi: Nevermind I just did it and force pushed XD

dfunckt:
I’m thinking it’s best to add the migration here too @thgreasi, due to open-balena, for the very unlikely case some instance has invalid data.

PR #1303, opened 2023-04-06
Revert #1302

dfunckt:
This reverts commit 4cdddd2d65c682e405d958f532826c74af182ae5.

Just after #1302 was merged we found that the scenario about device having "should be running release" pointing to invalid releases is also possible for applications. Reverting so we can discuss how to handle that case.


dfunckt:
I self-certify!

PR #1308, opened 2023-04-11
device-logs: Reduce the memory operations on initial stream GET response

thgreasi:
Accidentally found while investigating a "missing logs" report.
The `.shift()` is an `O(N)` operation and runs N times, while `.splice()` is `O(N)` and we can run it just once.
I initially thought that we might preferred `.shift()` for concurrency reasons, but that's not the case since it's a sync piece of code that runs it.

As an alternative, in case we still prefer a more "live buffer" processing approach (atm not necessary) then we could use a `while (i < buffer.length)` along with `splice()`.

Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/read.ts
+++ b/src/features/device-logs/lib/read.ts
@@ -138,7 +138,7 @@ async function handleStreamingRead(
 	// Subscribe in parallel so we don't miss logs in between
 	getBackend().subscribe(ctx, onLog);
 	try {
-		const logs = await getHistory(ctx, req, LOGS_DEFAULT_SUBSCRIPTION_COUNT);
+		let logs = await getHistory(ctx, req, LOGS_DEFAULT_SUBSCRIPTION_COUNT);
 
 		// We need this cast as typescript narrows to `StreamState.Buffering`
 		// because it ignores that during the `await` break it can be changed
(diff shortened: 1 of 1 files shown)


src/features/device-logs/lib/read.ts
```diff
@@ -144,16 +144,16 @@ async function handleStreamingRead(
 
 		const afterDate = logs.length && logs[logs.length - 1].createdAt;
 		// Append the subscription logs to the history queue
-		while (buffer.length) {
-			const log = buffer.shift();
+		for (const log of buffer) {
 			if (log && log.createdAt > afterDate) {
```
Page-, line 148: Presumably these need to be in date order, in which case can we find the index of the first valid log and then add all of those to the `logs` array in one go instead of continuing to iterate?

thgreasi: @Page- you means with something like v ?
```js
		const firstAfterDateIndex = buffer.findIndex(log => log.createdAt > afterDate);
		if (firstAfterDateIndex > 0) {
			buffer.splice(0, firstAfterDateIndex);
		}
		if (firstAfterDateIndex !== -1 && buffer.length > 0) {
			logs = logs.concat(buffer);
		}
```
Using `,concat` seems to be both faster than `.push(...buffer)` and safer than `.splice(0, 0, ...buffer)` (which for length > 200_000 starts throwing an `RangeError: Maximum call stack size exceeded`).
I didn't account for the GC that `.concat` will cause later though.

Page-: Yes that is what I meant (via any of the methods you mentioned)

fisehara (commented):
It looks good to me, but I'd like to take the chance and introduce some device-log stream test cases for checking any regression.

thgreasi:
@fisehara there you go
See: https://github.com/balena-io/open-balena-api/pull/1329

fisehara:
@thgreasi Thanks for adding the tests!

PR #1314, opened 2023-04-17
Switch from uuids to device ids for the device-heartbeat mechanism

Page-:
This allows for faster queries when applying the device heartbeats

Change-type: major


Page-:
LGTM

PR #1317, opened 2023-04-18
Avoid device updates when Redis indicates the device is heartbeating

thgreasi:
Change-type: patch


src/features/device-heartbeat/index.ts
```diff
@@ -324,29 +330,36 @@ export class DeviceOnlineStateManager extends EventEmitter<{
...
+			}
+			return JSON.parse(value) as DeviceOnlineStateManagerMessage;
+		} catch {
+			// Ignore
+		}
+	}
+
 	private async scheduleChangeOfStateForDevice(
 		deviceId: number,
+		currentManagerState: DeviceOnlineStateManagerMessage | undefined,
```
Page-, line 349: I think it'd be a bit clearer as:
```suggestion
		previousManagerState: DeviceOnlineStateManagerMessage | undefined,
```

src/features/device-heartbeat/index.ts
```diff
@@ -406,12 +419,20 @@ export class DeviceOnlineStateManager extends EventEmitter<{
 			return;
 		}
 
-		// update the device model...
-		await this.updateDeviceModel(deviceId, DeviceOnlineStates.Online);
+		const currentDeviceOnlineState = await this.getDeviceOnlineState(deviceId);
```
Page-, line 422: I'd also rename this as previous/last since it is the last one we set and and online will be the new one and during a transition state "current" is a bit confusing as it flips very quickly from pointing at the previous to the new state and so right here you're correct that it's "current", but a few lines later it is no longer current but is still named as such
```suggestion
		const previousDeviceOnlineState = await this.getDeviceOnlineState(deviceId);
```

PR #1318, opened 2023-04-18
heartbeat: Reuse the poll interval from the State GET result

thgreasi:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/Device.20heartbeat

### DIFF
--- a/src/features/device-state/index.ts
+++ b/src/features/device-state/index.ts
@@ -56,7 +56,10 @@ export const setup = (app: Application) => {
 };
 
 export interface Events {
-	'get-state': (deviceId: number, req: Pick<Request, 'apiKey'>) => void;
+	'get-state': (
+		deviceId: number,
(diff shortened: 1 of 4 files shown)


Page- (approved):

src/features/device-state/index.ts
```diff
@@ -56,7 +56,10 @@ export const setup = (app: Application) => {
 };
 
 export interface Events {
-	'get-state': (deviceId: number, req: Pick<Request, 'apiKey'>) => void;
+	'get-state': (
+		deviceId: number,
+		req: Pick<Request, 'apiKey'> & { config?: Dictionary<string> },
+	) => void;
```
Page-, line 62: You can also add extra args, you don't have to bundle it into one. It doesn't matter either way though

PR #1324, opened 2023-04-20
Reuse & export the helper that generates the config for device state GET

thgreasi:
Change-type: minor
See: https://github.com/balena-io/balena-api/pull/4387#discussion_r1172552515


thgreasi:
I self-certify!

PR #1336, opened 2023-05-03
Fix device heartbeat being stuck to offline when the DB update fails

thgreasi:
If setting the DB value to Online failed then we need to rethrow at
See: https://github.com/balena-io/open-balena-api/blob/9a29bc63e70973d72738c2e0fdbf478d84705b7a/src/features/device-heartbeat/index.ts#L442
so that the Redis state is not set to Online, otherwise the device's DB record  will be stuck on its previous heartbeat state forever.

Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/345882-_help/topic/device.20state.20stuck.20at.20VPN.20only

### DIFF
--- a/src/features/device-heartbeat/index.ts
+++ b/src/features/device-heartbeat/index.ts
@@ -236,7 +236,7 @@ export class DeviceOnlineStateManager extends EventEmitter<{
 	private async updateDeviceModel(
 		deviceId: number,
 		newState: DeviceOnlineStates,
-	): Promise<boolean> {
+	): Promise<void> {
 		const startAt = Date.now();
 		let err;
 
@@ -256,14 +256,13 @@ export class DeviceOnlineStateManager extends EventEmitter<{
 				},
 				body,
 			});
-			return true;
 		} catch ($err) {
 			err = $err;
 			captureException(
 				$err,
-				'DeviceStateManager: Error updating the API with the device new state.',
+				`DeviceStateManager: Error updating the API state of device ${deviceId} to ${newState}.`,
 			);
-			return false;
+			throw err;
 		} finally {
 			this.emit('change', {
 				deviceId,
@@ -306,14 +305,21 @@ export class DeviceOnlineStateManager extends EventEmitter<{
 									DeviceOnlineStates.Offline,
 									API_HEARTBEAT_STATE_TIMEOUT_SECONDS, // put the device into a timeout state if it misses it's scheduled heartbeat window... then mark as offline
 								),
-								this.updateDeviceModel(deviceId, DeviceOnlineStates.Timeout),
+								this.updateDeviceModel(
+									deviceId,
+									DeviceOnlineStates.Timeout,
+								).catch(() => {
+									// Ignore errors
+								}),
 							]);
 							break;
 						case DeviceOnlineStates.Offline:
 							await this.updateDeviceModel(
 								deviceId,
 								DeviceOnlineStates.Offline,
-							);
+							).catch(() => {
+								// Ignore errors
+							});
 							break;
 						default:
 							throw new Error(


src/features/device-heartbeat/index.ts
```diff
@@ -306,13 +309,18 @@ export class DeviceOnlineStateManager extends EventEmitter<{
...
+									DeviceOnlineStates.Timeout,
+									'catch',
+								),
 							]);
 							break;
 						case DeviceOnlineStates.Offline:
 							await this.updateDeviceModel(
 								deviceId,
 								DeviceOnlineStates.Offline,
+								'catch',
```
thgreasi, line 323: TBH I'm not sure why those two calls have always being silencing those errors :thinking:
The only thing I can imagine is so that `deleteMessageAsync()` gets called so that the queue gets cleaned-up rather than keep indefinitely grow, but I'm not convinced about that either.

Rsmq though does mention that received but not deleted messages will be re-received after the RSMQ_READ_TIMEOUT timeout, so I'm not sure whether that's the real reason :thinking: , so I just did the least intrusive change for this PR.

src/features/device-heartbeat/index.ts
```diff
@@ -256,14 +256,13 @@ export class DeviceOnlineStateManager extends EventEmitter<{
...
-			return true;
 		} catch ($err) {
 			err = $err;
 			captureException(
 				$err,
-				'DeviceStateManager: Error updating the API with the device new state.',
+				`DeviceStateManager: Error updating the API state of device ${deviceId} to ${newState}.`,
 			);
-			return false;
+			throw err;
```
thgreasi, line 265: If setting the DB value to Online failed then we need to rethrow at
See: https://github.com/balena-io/open-balena-api/blob/9a29bc63e70973d72738c2e0fdbf478d84705b7a/src/features/device-heartbeat/index.ts#L442
so that the Redis state is not set to Online, otherwise the device's DB record  will be stuck on its previous heartbeat state forever.

src/features/device-heartbeat/index.ts
```diff
@@ -306,14 +305,21 @@ export class DeviceOnlineStateManager extends EventEmitter<{
...
+								}),
 							]);
 							break;
 						case DeviceOnlineStates.Offline:
 							await this.updateDeviceModel(
 								deviceId,
 								DeviceOnlineStates.Offline,
-							);
+							).catch(() => {
+								// Ignore errors
```
thgreasi, line 321: TBH I'm not sure why those two calls have always being silencing those errors :thinking:
The only thing I can imagine is so that `deleteMessageAsync()` gets called so that the queue gets cleaned-up rather than keep indefinitely grow, but I'm not convinced about that either.

Rsmq though does mention that received but not deleted messages will be re-received after the RSMQ_READ_TIMEOUT timeout, so I'm not sure whether that's the real reason :thinking: , so I just did the least intrusive change for this PR.

thgreasi: Could it be so that we don't keep re-receiving the message and re-running the PATCH forever, in case that the PATCHes fail b/c of DB load, resulting even more load to the DB?

fisehara (commented):
@thodoris 
Looks reasonable, but I'm wondering as the other methods called in `captureEventFor` catch exceptions to ignore them, if it should follow the pattern.
Then we need to evaluate the return value , rather than throwing an except.
I don't know if this is some kind of protection against spamming the exception capturer.

thgreasi:
@fisehara the best reason I can think of why we were silencing the errors is v:
https://github.com/balena-io/open-balena-api/pull/1336#discussion_r1183402153

> Looks reasonable, but I'm wondering as the other methods called in captureEventFor catch exceptions to ignore them, if it should follow the pattern.

It shouldn't follow this pattern of silencing errors. In fact, stopping the execution of `captureEventFor` was the primary reason I did this change. I originally went the "check the return value approach" to result a smaller diff, but then I realized that it made more sense to follow the standard throw/catch approach.

fisehara (approved):
After discussion LGTM

PR #1338, opened 2023-05-03
Add cache busting config to detect out-of-sync heartbeats & upsert them

thgreasi:
Allows us to fix the case that we have
detected that Online heartbeats keys in REDIS
where not written in the DB, by making the
API upsert the state in the DB once.

Change-type: minor
See: https://balena.zulipchat.com/#narrow/stream/345882-_help/topic/device.20state.20stuck.20at.20VPN.20only


src/features/device-heartbeat/index.ts
```diff
@@ -150,6 +151,8 @@ interface MetricEventArgs {
 interface DeviceOnlineStateManagerMessage {
 	id: string;
 	currentState: DeviceOnlineStates;
+	/** Cache Busting value */
```
fisehara, line 154: I know it stands next to it, but cb is very commonly used for callback.

src/features/device-heartbeat/index.ts
```diff
@@ -436,7 +442,10 @@ export class DeviceOnlineStateManager extends EventEmitter<{
 		// If redis still has a valid message about the device being online we can avoid reaching to the DB...
 		if (
 			previousDeviceOnlineState == null ||
-			previousDeviceOnlineState.currentState !== DeviceOnlineStates.Online
+			previousDeviceOnlineState.currentState !== DeviceOnlineStates.Online ||
```
fisehara, line 445: Shouldn't we make it a continuous  cache busting mechanism? I see that the redis cache is updated with new expiration, should we maybe rethink this continuous update, or should we allow it to get stale after the expiration has run out, access the DB and write the redis with DB synced data?

fisehara (commented):
I wonder if making it a one time busting is helping on the long run.

src/features/device-heartbeat/index.ts
thgreasi: Can you expand on how a continuous cache busting mechanism would work?
The primary point of this "cache" is to avoid reaching continuously to the DB

Theoretically with #1336 we should no longer get into this inconsistent state, since we will only be updating Redis only after we are sure that the DB write was persisted.
The fact that we only got such reports on support after the PG upgrade sounds like going out of sync is already not that common.
In case though that we are missing more code paths that can lead to an inconsistent state, we can still as a hotfix change the cache busting key to fix it for customers while trying to find the code path that allowed going out of sync at the first place.
An other approach I was thinking was for example to have a counter and persist the write to the DB on every Nth heartbeat or something. That would conceal though possible paths that can lead to inconsistent states, so I preferred to leave that as a last resort.

src/features/device-heartbeat/index.ts
thgreasi: TBH I could't figure out what the ideal short prop name would be, so I asked an AI bot what is the most common name used for this use case and it sounded convincing enough to me. Better ideas are welcome of course.
Also, callbacks shouldn't be that common nowadays, right?

src/features/device-heartbeat/index.ts
thgreasi: @fisehara here is an alternative w/ a TTL approach #1339

thgreasi:
Closing in favor of #1340

PR #1341, opened 2023-05-05
Change device heartbeat's redis prefix to device-heartbeat-state

Page-:
Change-type: patch

### DIFF
--- a/src/features/device-heartbeat/index.ts
+++ b/src/features/device-heartbeat/index.ts
@@ -165,7 +165,7 @@ export class DeviceOnlineStateManager extends EventEmitter<{
 		},
 	) => void;
 }> {
-	private static readonly REDIS_NAMESPACE = 'device-online-state';
+	private static readonly REDIS_NAMESPACE = 'device-heartbeat-state';
 	private static readonly EXPIRED_QUEUE = 'expired';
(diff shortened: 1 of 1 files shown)


src/features/device-heartbeat/index.ts
```diff
@@ -305,21 +305,14 @@ export class DeviceOnlineStateManager extends EventEmitter<{
 									DeviceOnlineStates.Offline,
 									API_HEARTBEAT_STATE_TIMEOUT_SECONDS, // put the device into a timeout state if it misses it's scheduled heartbeat window... then mark as offline
 								),
-								this.updateDeviceModel(
-									deviceId,
-									DeviceOnlineStates.Timeout,
-								).catch(() => {
-									// Ignore errors
-								}),
+								this.updateDeviceModel(deviceId, DeviceOnlineStates.Timeout),
```
thgreasi, line 308: Should we now move this before the scheduleChangeOfStateForDevice and stop using a Promise.all?
Since scheduleChangeOfStateForDevice runs in parallel, it might have already scheduled a new state change before updateDeviceModel throws, in which case the `deleteMessageAsync` at the end of the function will not be called, hence ending up w/ 2 messages about the same device in the queue (the old Online->Timeout that wasn't deleted and will be re-received, and the new Timeout->Offline).

PR #1343, opened 2023-05-09
Add the device.changed_api_heartbeat_state_on__date term to the model

thgreasi:
Resolves: #644
Change-type: minor
See: https://balena.fibery.io/Work/Project/Start-recording-the-last-timestamp-the-device-heartbeat-changed-254

### DIFF
--- a/src/features/device-heartbeat/index.ts
+++ b/src/features/device-heartbeat/index.ts
@@ -250,7 +250,7 @@ export class DeviceOnlineStateManager extends EventEmitter<{
 
 		try {
 			// patch the api_heartbeat_state value to the new state...
-			const body = {
+			const baseBody = {
 				api_heartbeat_state: newState,
 			};
 			await api.resin.patch({
@@ -259,10 +259,17 @@ export class DeviceOnlineStateManager extends EventEmitter<{
 				id: deviceId,
 				options: {
 					$filter: {
-						$not: body,
+						$not: baseBody,
 					},
 				},
-				body,
+				body: {
+					...baseBody,
+					// Since the heartbeat manager is the only place that we update the heartbeat state
+					// we are updating the heartbeat's change date in here rather than a hook, so that
+					// we can avoid the extra DB request that a generic hook would require for checking
+					// whether the value actually changed or not.
+					changed_api_heartbeat_state_on__date: Date.now(),
+				},
 			});
 		} catch ($err) {
 			err = $err;
--- a/test/03_device-state.ts
+++ b/test/03_device-state.ts
@@ -11,7 +11,10 @@ import * as config from '../src/lib/config.js';
 import * as stateMock from '../src/features/device-heartbeat/index.js';
 import { assertExists, itExpectsError, waitFor } from './test-lib/common.js';
 import * as fixtures from './test-lib/fixtures.js';
-import { expectResourceToMatch } from './test-lib/api-helpers.js';
+import {
+	expectResourceToMatch,
+	thatIsDateStringAfter,
+} from './test-lib/api-helpers.js';
 import { redis, redisRO } from '../src/infra/redis/index.js';
 import { setTimeout } from 'timers/promises';
 import { MINUTES, SECONDS } from '@balena/env-parsing';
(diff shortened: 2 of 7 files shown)


src/features/device-heartbeat/index.ts
```diff
@@ -254,10 +254,13 @@ export class DeviceOnlineStateManager extends EventEmitter<{
...
 				options: {
 					$filter: {
-						$not: body,
+						$not: baseBody,
 					},
 				},
-				body,
+				body: {
+					...baseBody,
+					last_api_heartbeat_state_change_event: Date.now(),
```
thgreasi, line 262: I originally added this as a hook, but at the end I preferred moving it here
* so that we can avoid the extra `.get()` request to confirm that the value has indeed changed, or blindly setting `last_api_heartbeat_state_change_event` whenever `values.api_heartbeat_state != null`.
* since this is the only place that we should be updating the heartbeat anyway.

Let me know though if you prefer to switch it back to hooks.

src/features/device-heartbeat/index.ts
```diff
@@ -254,10 +254,13 @@ export class DeviceOnlineStateManager extends EventEmitter<{
...
 					$filter: {
-						$not: body,
+						$not: baseBody,
 					},
 				},
-				body,
+				body: {
+					...baseBody,
+					last_changed_api_heartbeat_state_on__date: Date.now(),
+				},
```
fisehara, line 263: I wonder if `last_changed_api_heartbeat_state_on__date` should only be updated when called via: `await this.updateDeviceModel(deviceId, DeviceOnlineStates.Online);`
and not when patching it to timeout or offline? As these events are internal events that do not show the device are heartbeating anymore?

test/03_device-state.ts
```diff
@@ -276,29 +281,38 @@ mockery.registerMock('../src/lib/config', configMock);
...
-									'API heartbeat state is not timeout',
+								await expectResourceToMatch(
+									getActor(),
+									'device',
+									getDevice().id,
+									{
+										api_heartbeat_state: DeviceOnlineStates.Timeout,
+										last_changed_api_heartbeat_state_on__date:
+											thatIsDateStringAfter(stateUpdatedAfter),
+									},
```
fisehara, line 309: When we use the `last_changed_api_heartbeat_state_on__date` what do we want to convey? Do we want to convey when the device was last seen with a proper heartbeat or when the heartbeat status changes from online, to timeout, to offline?
How will we use the timeout and offline timestamps, eg as indicator in the UI?

fisehara (commented):
Unfortunately I don't understand properly what this added property adds to the device state information? 
Do we want to convey the last proper heartbeat time or do we want to surface all changes of heartbeat event states? 
Eg. when it's set to online over and over again, it would be of interest when the last time the heartbeat was set to online, or am I mixing things up?

src/features/device-heartbeat/index.ts
thgreasi: The intention here is to track when the heartbeat was changes both for any reason (both from state-GETs and from heartbeat changes b/c of the heartbeat mechanism). Knowing then the device switched to Timeout/Offline is probably more useful than knowing since when the device originally started heartbeating.

fisehara (approved):

thgreasi:
Thank you @fisehara I will make this part of the next cycle, so that we confirm the name again :heart:

dfunckt:
@thgreasi I wonder if having a “state last modified date” instead would be sufficient, ie the date the device last switched between Online, Downloading, etc. states.

That would allow us to show eg. “Downloading” and “for 10 minutes”, or something similar.

It would also be useful to filter/order by this field in the device list and check if I, as a fleet operator, need to intervene—3 days in Downloading state clearly means something is off.

cc @shaunmulligan

thgreasi:
@dfunckt the `device.state` is whatever the supervisor reports & is not changing based on the vpn.
The `device.overall_status` is computed term that depends on the heartbeat state on the other hand has different definitions between v6 & v7, so in order to capture the change date of the `overall_status` we would need to be computing & storing the value for both versions (or ignore v6 for example).

That's why we ended up deciding (years back) that storing the change date of the most primitive value is probably better, since the data is still usable even if we change how `overall_status` gets computed.

PR #1347, opened 2023-05-11
Add a CPU ID readable character filter

fisehara:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/CPU.20ID.20filter

### DIFF
--- a/src/features/device-state/routes/state-patch-v2.ts
+++ b/src/features/device-state/routes/state-patch-v2.ts
@@ -157,6 +157,15 @@ export const statePatchV2: RequestHandler = async (req, res) => {
 			if (local.name != null) {
 				deviceBody.device_name = local.name;
 			}
+			if (deviceBody.cpu_id != null) {
+				if (/[^\x20-\x7E]/.test(deviceBody.cpu_id)) {
+					// if the CPU id is not in the character range of 0x20 (SPACE) to 0x7e (~) we drop the CPU ID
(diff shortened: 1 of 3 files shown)


src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -192,16 +192,25 @@ export const statePatchV2: RequestHandler = async (req, res) => {
...
 						if (deviceBody.cpu_id != null) {
-							deviceBody.cpu_id = deviceBody.cpu_id.toLowerCase();
+							if (/[^\x20-\x7E]/.test(deviceBody.cpu_id)) {
+								// if the CPU id is not in the character range of 0x20 (SPACE) to 0x7e (~) we drop the CPU ID
+								// this cpu id wouldn't be rendered anyway
+								delete deviceBody.cpu_id;
+							} else {
+								deviceBody.cpu_id = deviceBody.cpu_id.toLowerCase();
+							}
+						}
```
Page-, line 202: It would make more sense to move this up next to
```ts
			if (local.name != null) {
				deviceBody.device_name = local.name;
			}
```
so that the `Object.keys(deviceBody).length > 0` check will be handled in the standard way

fisehara: Oh I see, thanks

Page- (approved):

PR #1352, opened 2023-05-17
Combine two of the rules about supervisor releases managing devices

thgreasi:
Those two rules
* seem to be running at the same rate
* they both have row narrowing on the release, so their runtime when creating releases should be super fast
* the slowest part in the execution plans of both is the device-to-release join
* wrt the columns that trigger the rules, their difference is limited to the following:
  * device `is of-device type`, application `is for-device type`, device type, cpu architecture
  * application `is public`, `is host` & `slug`
  * Changing those values individually is way less frequent than changing the supervisor release or registering new devices or creating new applications, but in any case the combined rule performs better than the individual ones.

As a result it seems that
* the majority of the CPU time of the two rules is from when new devices are created or when updating the `device.should be managed by-release`
* combining should be beneficial,
* especially for the "same cpu arch" rule, including the application slug filter seems to be beneficial since the `slug` has higher selectivity which it allows the planner to switch from a `Seq Scan on public.application` to an `Index Scan using application_slug_public_host_idx on public.application` which is faster since it reduces the number of application rows that need to be processed.

Results:
* https://explain.dalibo.com/plan/4fe3010969dged0d 2s 767ms supervisor app check
* https://explain.dalibo.com/plan/137chd23f2f37558 1s 196ms same cpu arch check
* https://explain.dalibo.com/plan/44ec397727c54e49 1s 156ms combined

Change-type: patch


fisehara (commented):
LGTM, I don't know if you want a second review on it before merging it. Otherwise I'll approve it.

Page- (approved):

PR #1362, opened 2023-06-09
Gracefully deny device registrations with duplicate api key

thgreasi:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346855-_announcements/topic/balenaCloud.20end-to-end.20tests


src/features/device-provisioning/hooks/cache-invalidation.ts
```diff
@@ -0,0 +1,45 @@
...
+				options: {
+					$select: keyProperty,
+					$filter: {
+						id: { $in: affectedIds },
+					},
+				},
+			})) as Array<{ [key in K]: string }>;
+
+			// Invalidate the caches only when the tx is committed
+			tx.on('end', async () => {
```
Page-, line 33: Currently this `async` is unused

src/features/device-provisioning/hooks/cache-invalidation.ts
```diff
@@ -0,0 +1,45 @@
...
+				for (const affectedItem of affectedItems) {
+					// Run in the background as this is not a reason to fail the request
+					cache.delete(affectedItem[keyProperty]);
```
Page-, line 36: If the delete fails then it'll be an unhandled promise rejection and cause a bunch of warnings

src/features/device-provisioning/gracefully-deny-conflicting-registrations.ts
```diff
@@ -0,0 +1,101 @@
...
+import {
+	API_KEY_EXISTS_CACHE_TIMEOUT,
+	DEVICE_EXISTS_CACHE_TIMEOUT,
+} from '../../lib/config';
+import { createMultiLevelStore } from '../../infra/cache';
+
+const { ConflictError } = errors;
+const { api } = sbvrUtils;
+
+const checkDeviceExistsQuery = _.once(() =>
```
Page-, line 12: Don't we already have a mechanism for checking devices exist?

thgreasi: We do but the other one the exists/isFrozen one which is unnecessarily expensive if it has to run.
Moreover for this case we do not care to cache the result if the device does not exist.

src/features/device-provisioning/gracefully-deny-conflicting-registrations.ts
```diff
@@ -0,0 +1,65 @@
...
+	// we have to check for the UUID so that a registration with conflicting UUID & api key still throws
+	// the same error as if this optimization did not exist.
+	const device = await checkDeviceExistsIsFrozen(uuid);
+	if (device != null) {
+		throw new ConflictError('"uuid" must be unique.');
+	}
+
+	const existingApiKey = apiKey != null && (await checkApiKeyExists(apiKey));
+	if (existingApiKey) {
+		throw new ConflictError('"key" must be unique.');
```
fisehara, line 63: Maybe more specific like: 

```suggestion
		throw new ConflictError('"api_key" must be unique.');
```

src/features/device-provisioning/gracefully-deny-conflicting-registrations.ts
```diff
@@ -0,0 +1,65 @@
...
+export async function gracefullyDenyConflictingRegistrations(
+	uuid: string,
+	apiKey: string | undefined,
+) {
+	// Even requests with conflicting device UUIDs are cheap, since they throw before the rules run,
+	// we have to check for the UUID so that a registration with conflicting UUID & api key still throws
+	// the same error as if this optimization did not exist.
+	const device = await checkDeviceExistsIsFrozen(uuid);
+	if (device != null) {
+		throw new ConflictError('"uuid" must be unique.');
```
fisehara, line 58: Maybe call it a device uuid, when consuming the API an UUID could also be related to fleet?

```suggestion
		throw new ConflictError('"device uuid" must be unique.');
```

src/features/device-provisioning/register.ts
```diff
@@ -43,6 +42,8 @@ export const register: RequestHandler = async (req, res) => {
 			throw new BadRequestError('UUID must be specified');
 		}
 
+		await gracefullyDenyConflictingRegistrations(uuid, req.body.api_key);
```
fisehara, line 45: If the following calls fail the cache has been written but the uuid is potentially not registered? Can it be possible that the rules fail the TX but the cache has already save the uuid / api_key and will report it conflicting?
Does the cache needs a cleanup in the catch?

fisehara (commented):
Make sense, NIT + error case in register after cache is written.

src/features/device-provisioning/gracefully-deny-conflicting-registrations.ts
thgreasi: This is exactly the error that pine throws, so I kept it the same for consistency (so that optimization is transparent to users). I will add a comment.

src/features/device-provisioning/gracefully-deny-conflicting-registrations.ts
thgreasi: > This is exactly the error that pine throws, so I kept it the same for consistency (so that optimization is transparent to users). I will add a comment.

src/features/device-provisioning/hooks/cache-invalidation.ts
```diff
@@ -0,0 +1,52 @@
...
+			tx.on('end', () => {
+				for (const affectedItem of affectedItems) {
+					// Run in the background as this is not a reason to fail the request
+					(async () => {
+						try {
+							await cache.delete(affectedItem[keyProperty]);
+						} catch (err) {
+							captureException(
+								err,
+								`Error while invalidating a(n) ${model} record from the device registration grafecully deny cache: ${affectedIds}`,
```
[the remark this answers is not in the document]
thgreasi: Nice spot :+1:

src/features/device-provisioning/gracefully-deny-conflicting-registrations.ts
```diff
@@ -0,0 +1,69 @@
...
+			isCacheableValue: (value: any) => value === true,
+			ttl: API_KEY_EXISTS_CACHE_TIMEOUT,
+		},
+		local: false,
+	},
+	false,
+);
+
+const checkApiKeyExists = async (key: string) => {
+	return await checkApiKeyExistsStore.wrap(key, async () => {
```
Page-, line 38: Any reason you're manually creating the store and using `.wrap` rather than using the memoize helper for it?

thgreasi: @Page- b/c I'm using `isCacheableValue: (value: any) => value === true`, so that we do not cache keys that are not found, since in that case the cache would have to be invalidated right after. So instead of caching & deleting in the same register call, I thought we could avoid storing/caching it at the first place.

thgreasi: I could switch it over to `multiCacheMemoizee` and throw to avoid caching in the cases that it doesn't make sense though if you prefer it.

PR #1371, opened 2023-07-04
Split metrics from `device` resource

fisehara:
Remodel `device` resource and split metrics into `device metrics record`

This reduces the load on the `device` resource when updating device metrics periodically. device metrics are far less frequently needed than the `device` resource. Thus writing to the heavily used `device` resource should be offloaded to the `device metrics record` resource.

Change-type: major

### DIFF
--- a/src/balena-init.sql
+++ b/src/balena-init.sql
@@ -188,3 +188,6 @@ ON "application" ("slug" varchar_pattern_ops, "is public", "is host");
 
 CREATE INDEX IF NOT EXISTS "scheduled_job_run_start_timestamp_idx"
 ON "scheduled job run" (DATE_TRUNC('milliseconds', "start timestamp"));
+
+CREATE INDEX IF NOT EXISTS "device_metrics_record_by_device_idx"
+ON "device metrics record" ("is reported by-device");
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -581,6 +581,31 @@ Fact type: device should be managed by release
 	Necessity: each device should be managed by at most one release
 
 
+Term: device metrics record
+
+-- device metrics record
+Fact type: device metrics record is reported by device
+    Synonymous Form: device reports device metrics record
+    Necessity: each device metrics record is reported by exactly one device
+	Necessity: each device reports exactly one device metrics record
+
+Fact type: device metrics record has memory usage
+	Necessity: each device metrics record has at most one memory usage
+Fact type: device metrics record has memory total
+	Necessity: each device metrics record has at most one memory total
+Fact type: device metrics record has storage block device
+	Necessity: each device metrics record has at most one storage block device
+Fact type: device metrics record has storage usage
+	Necessity: each device metrics record has at most one storage usage
+Fact type: device metrics record has storage total
+	Necessity: each device metrics record has at most one storage total
+Fact type: device metrics record has cpu usage
+	Necessity: each device metrics record has at most one cpu usage
+Fact type: device metrics record has cpu temp
+	Necessity: each device metrics record has at most one cpu temp
+Fact type: device metrics record is undervolted
+
+
 -- application config variable
 
 Fact type: application config variable has value
--- a/src/features/device-state/index.ts
+++ b/src/features/device-state/index.ts
@@ -23,9 +23,11 @@ export {
 	serviceInstallFromImage,
 } from './state-get-utils';
 export {
-	metricsPatchFields,
+	validDeviceMetricsRecordPatchFields,
 	v2ValidPatchFields,
+	v2ValidDevicePatchFields,
 	v3ValidPatchFields,
+	v3ValidDevicePatchFields,
 } from './state-patch-utils';
 
 const gracefulGet = resolveOrDenyDevicesWithStatus(304);
--- a/src/features/device-state/routes/state-patch-v2.ts
+++ b/src/features/device-state/routes/state-patch-v2.ts
@@ -10,11 +10,12 @@ import { getIP } from '../../../lib/utils';
 import type { ImageInstall, PickDeferred } from '../../../balena-model';
 import {
 	shouldUpdateMetrics,
-	metricsPatchFields,
-	v2ValidPatchFields,
 	upsertImageInstall,
 	deleteOldImageInstalls,
 	truncateShortTextFields,
+	StatePatchDeviceMetricsRecordBody,
+	validDeviceMetricsRecordPatchFields,
+	v2ValidDevicePatchFields,
 } from '../state-patch-utils';
 import type { ResolveDeviceInfoCustomObject } from '../middleware';
 
@@ -26,7 +27,7 @@ type LocalBody = NonNullable<StatePatchV2Body['local']>;
  * These typings should be used as a guide to what should be sent, but cannot be trusted as what actually *is* sent.
  */
 export type StatePatchV2Body = {
-	local?: {
+	local?: StatePatchDeviceMetricsRecordBody & {
 		should_be_running__release?: number;
 		name?: string;
 		/**
@@ -46,15 +47,6 @@ export type StatePatchV2Body = {
 		download_progress?: number | null;
 		api_port?: number;
 		api_secret?: string;
-		memory_usage?: number;
-		memory_total?: number;
-		storage_block_device?: string;
-		storage_usage?: number;
-		storage_total?: number;
-		cpu_temp?: number;
-		cpu_usage?: number;
-		cpu_id?: string;
-		is_undervolted?: boolean;
 		is_on__commit?: string | null;
 		apps?: Array<{
 			services?: {
@@ -139,20 +131,13 @@ export const statePatchV2: RequestHandler = async (req, res) => {
 			const { apps } = local;
 
 			let deviceBody:
-				| Pick<LocalBody, (typeof v2ValidPatchFields)[number]> & {
+				| Pick<LocalBody, (typeof v2ValidDevicePatchFields)[number]> & {
 						is_running__release?: number | null;
-				  } = _.pick(local, v2ValidPatchFields);
-			let metricsBody: Pick<LocalBody, (typeof metricsPatchFields)[number]> =
-				_.pick(local, metricsPatchFields);
-			if (
-				Object.keys(metricsBody).length > 0 &&
-				(await shouldUpdateMetrics(uuid))
-			) {
-				// If we should force a metrics update then merge the two together and clear `metricsBody` so
-				// that we don't try to merge it again later
-				deviceBody = { ...deviceBody, ...metricsBody };
-				metricsBody = {};
-			}
+				  } = _.pick(local, v2ValidDevicePatchFields);
+			const metricsBody: Pick<
+				LocalBody,
+				(typeof validDeviceMetricsRecordPatchFields)[number]
+			> = _.pick(local, validDeviceMetricsRecordPatchFields);
 
 			if (local.name != null) {
 				deviceBody.device_name = local.name;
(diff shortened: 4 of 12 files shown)


src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -334,6 +322,33 @@ export const statePatchV3: RequestHandler = async (req, res) => {
...
 
+				if (
+					Object.keys(metricsBody).length > 0 &&
+					(await shouldUpdateMetrics(uuid))
+				) {
+					updateFns.push(async (resinApiTx) => {
+						const latestDeviceMetricsRecord = await resinApiTx.get({
+							resource: 'device_metrics_record',
+							id: { is_reported_by__device: device.id },
+						});
```
Page-, line 333: This can be merged into the `fetchData` request rather than adding an extra one

src/migrations/00084-split-device-metrics-from-device-schema.sql
```diff
@@ -0,0 +1,19 @@
...
+,	"storage total" INTEGER NULL
+,	"cpu usage" INTEGER NULL
+,	"cpu temp" INTEGER NULL
+,	"is undervolted" BOOLEAN DEFAULT FALSE NOT NULL
+,	FOREIGN KEY ("is reported by-device") REFERENCES "device" ("id")
+);
+
+
+CREATE INDEX IF NOT EXISTS "device_metrics_record_by_device_idx"
+ON "device metrics record" ("is reported by-device");
```
Page-, line 19: This needs to be included in `balena-init.sql`

src/balena.sbvr
```diff
@@ -528,21 +529,6 @@ Fact type: device has ip address
...
-	Necessity: each device has at most one storage block device
-Fact type: device has storage usage
-	Necessity: each device has at most one storage usage
-Fact type: device has storage total
-	Necessity: each device has at most one storage total
-Fact type: device has cpu usage
-	Necessity: each device has at most one cpu usage
-Fact type: device has cpu temp
-	Necessity: each device has at most one cpu temp
-Fact type: device is undervolted
```
Page-, line 545: I don't think this can be safely removed until the async migration relying on it is finalized

src/migrations/00084-split-device-metrics-from-device-schema.sql
fisehara: Thanks

src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -208,6 +191,35 @@ export const statePatchV2: RequestHandler = async (req, res) => {
...
+							resource: 'device_metrics_record',
+							id: { is_reported_by__device: deviceId },
+							body: {
+								...metricsBody,
+							},
+						});
+					} else {
+						console.log(
+							`patch metricsBody:${JSON.stringify(metricsBody, null, 2)}`,
+						);
```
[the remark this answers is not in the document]
fisehara: thanks

src/features/device-state/routes/state-patch-v3.ts
fisehara: Thanks, great hint also explicitly loading with readOnly transcation. 

I've added it to the device request by expanding the `reports__device_metrics_record`

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -297,7 +282,11 @@ export const statePatchV3: RequestHandler = async (req, res) => {
 				}
 			}
 
-			if (apps != null || Object.keys(deviceBody).length > 0) {
+			if (
+				apps != null ||
+				Object.keys(deviceBody).length > 0 ||
+				Object.keys(metricsBody).length > 0
```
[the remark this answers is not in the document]
fisehara: Thanks!

fisehara:
@Page- @thgreasi when this is fine I'll follow up with a PR  for finalising the async migration and dropping the columns from `device` resource.

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -334,6 +334,26 @@ export const statePatchV3: RequestHandler = async (req, res) => {
...
+								body: {
+									...metricsBody,
+									...{ is_reported_by__device: device.id },
+								},
+							});
+						} else {
+							await resinApiTx.patch({
+								resource: 'device_metrics_record',
+								id: { is_reported_by__device: device.id },
+								body: metricsBody,
```
thgreasi, line 351: How about adding v both here and in the v2 patch?
```suggestion
								options: {
									$filter: {
										$not: metricsBody
									}
								},
								body: metricsBody,
```

src/migrations/00084-split-device-metrics-from-device-schema.sql
thgreasi: @Page- @fisehara Afaict we shouldn't need to add a manual index (neither here nor in the `balena-init.sql`), since the `"is reported by-device" INTEGER NOT NULL UNIQUE` part of the `CREATE TABLE` does auto-generate the following index:
```sql
CREATE UNIQUE INDEX "device metrics record_is reported by-device_key" ON public."device metrics record" USING ("is reported by-device");
```
which makes `device_metrics_record_by_device_idx` a duplicate index.

If we still prefer to have the index in the migration just in case, then it should probably be matching the auto-generated one.

thgreasi: > Adding a unique constraint will automatically create a unique btree index on the column or group of columns used in the constraint.

See: https://www.postgresql.org/docs/current/sql-createtable.html

src/features/device-state/index.ts
```diff
@@ -23,9 +23,11 @@ export {
 	serviceInstallFromImage,
 } from './state-get-utils';
 export {
-	metricsPatchFields,
+	validDeviceMetricsRecordPatchFields,
```
thgreasi, line 26: NIT: Since this field is also exported from index.ts w/ `export * as deviceState from './features/device-state';`, how about not renaming it so that we don't introduce a breaking change?

src/balena-init.sql
```diff
@@ -188,3 +188,6 @@ ON "application" ("slug" varchar_pattern_ops, "is public", "is host");
 
 CREATE INDEX IF NOT EXISTS "scheduled_job_run_start_timestamp_idx"
 ON "scheduled job run" (DATE_TRUNC('milliseconds', "start timestamp"));
+
+CREATE INDEX IF NOT EXISTS "device_metrics_record_by_device_idx"
```
thgreasi, line 192: Reminder about the discussion on this at
See: https://github.com/balena-io/open-balena-api/pull/1371/files#r1255500040

src/migrations/00085-split-device-metrics-from-device.async.ts
```diff
@@ -0,0 +1,78 @@
...
+		FROM (
+			SELECT * FROM "device"
+			WHERE
+				"device".id NOT IN (
+					SELECT
+						"is reported by-device"
+					FROM
+						"device metrics record"
+				)
+			FOR UPDATE SKIP LOCKED
```
thgreasi, line 69:
```suggestion
			FOR UPDATE
```
The sync migration one should not leave anything pending.

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -103,11 +98,13 @@ const fetchData = async (
 					belongs_to__application: {
 						$select: 'uuid',
 					},
+					reports__device_metrics_record: {},
 				},
 			},
 		})) as Array<
 			Pick<Device, 'id' | 'uuid'> & {
 				belongs_to__application: Array<Pick<Application, 'uuid'>>;
+				reports__device_metrics_record: DeviceMetricsRecord[];
```
thgreasi, line 107: We don't really care to fetch any of the rest columns.
```suggestion
					reports__device_metrics_record: {
						$select: 'id'
					},
				},
			},
		})) as Array<
			Pick<Device, 'id' | 'uuid'> & {
				belongs_to__application: Array<Pick<Application, 'uuid'>>;
				reports__device_metrics_record: reports__device_metrics_record: [Pick<DeviceMetricsRecord, 'id'>?];
```

src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -208,6 +191,32 @@ export const statePatchV2: RequestHandler = async (req, res) => {
...
+			}
+			if (
+				Object.keys(metricsBody).length > 0 &&
+				(await shouldUpdateMetrics(uuid))
+			) {
+				updateFns.push(async (resinApiTx) => {
+					const latestDeviceMetricsRecord = await resinApiTx.get({
+						resource: 'device_metrics_record',
+						id: { is_reported_by__device: deviceId },
+					});
```
thgreasi, line 205:
```suggestion
					const latestDeviceMetricsRecord = await resinApiTx.get({
						resource: 'device_metrics_record',
						id: { is_reported_by__device: deviceId },
						options: {
							$select: 'id'
						}
					});
```

src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -208,6 +191,32 @@ export const statePatchV2: RequestHandler = async (req, res) => {
...
+					if (latestDeviceMetricsRecord == null) {
+						await resinApiTx.post({
+							resource: 'device_metrics_record',
+							id: { is_reported_by__device: deviceId },
+							body: {
+								...metricsBody,
+							},
+						});
```
thgreasi, line 213: Does that work? Shouldn't it be v ?
```suggestion
						await resinApiTx.post({
							resource: 'device_metrics_record',
							body: {
								...metricsBody,
								is_reported_by__device: deviceId,
							},
						});
```

src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -208,6 +191,32 @@ export const statePatchV2: RequestHandler = async (req, res) => {
...
+					} else {
+						await resinApiTx.patch({
+							resource: 'device_metrics_record',
+							id: { is_reported_by__device: deviceId },
+							body: metricsBody,
```
thgreasi, line 218: Given that we are in the same TX, we could also use the PK (even though it shouldn't make any difference).
```suggestion
							id: latestDeviceMetricsRecord.id,
							options: {
								$filter: {
									$not: metricsBody
								}
							},
							body: metricsBody,
```

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -334,6 +334,26 @@ export const statePatchV3: RequestHandler = async (req, res) => {
...
 				}
 
+				if (Object.keys(metricsBody).length > 0) {
+					updateFns.push(async (resinApiTx) => {
+						if (latestDeviceMetricsRecord == null) {
+							await resinApiTx.post({
+								resource: 'device_metrics_record',
+								body: {
+									...metricsBody,
+									...{ is_reported_by__device: device.id },
```
thgreasi, line 344:
```suggestion
									is_reported_by__device: device.id,
```

src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -208,6 +191,32 @@ export const statePatchV2: RequestHandler = async (req, res) => {
...
+							resource: 'device_metrics_record',
+							id: { is_reported_by__device: deviceId },
+							body: {
+								...metricsBody,
+							},
+						});
+					} else {
+						await resinApiTx.patch({
+							resource: 'device_metrics_record',
+							id: { is_reported_by__device: deviceId },
```
thgreasi, line 217:
```suggestion
							id: { is_reported_by__device: deviceId },
							options: {
								$filter: {
									$not: metricsBody
								}
							},
```

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -297,12 +285,25 @@ export const statePatchV3: RequestHandler = async (req, res) => {
 				}
 			}
 
-			if (apps != null || Object.keys(deviceBody).length > 0) {
+			if (
+				Object.keys(metricsBody).length > 0 &&
+				!(await shouldUpdateMetrics(uuid))
```
Page-, line 290: There's a special case I think where we always want to recognize changes to the `is_undervolted` property which we currently always accept but this PR has moved to being treated like any of the other "metrics" fields

src/features/device-state/routes/state-patch-v3.ts
Page-: Yeah, this avoids an unnecessary postgres write in the case there are no changes

PR #1379, opened 2023-07-18
Export the augmentReqApiKeyPermissions api key helper

thgreasi:
Change-type: minor

### DIFF
--- a/src/features/api-keys/lib.ts
+++ b/src/features/api-keys/lib.ts
@@ -335,10 +335,27 @@ export const augmentReqApiKeyPermissions = <
 	T extends permissions.PermissionReq,
 >(
 	req: T,
-	...extraPermissions: string[]
+	extraPermissions: string[],
+	/**
+	 * When mutateRequestObject is
+	 * false: A new request object with augmented permissions is returned to be used for specific tasks,
+	 *   while the rest of the code (eg: middleware & hooks) still runs with permissions the original request.
+	 * true: The permissions are augmented for the whole lifetime of the request.
+	 */
+	mutateRequestObject = false,
 ): T => {
-	const augmentedReq = _.clone(req);
-	augmentedReq.apiKey = _.cloneDeep(augmentedReq.apiKey);
+	let augmentedReq = req;
+
+	if (!mutateRequestObject) {
+		augmentedReq = _.clone(req);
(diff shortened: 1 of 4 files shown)


src/features/api-keys/lib.ts
```diff
@@ -335,9 +335,10 @@ export const augmentReqApiKeyPermissions = <
 	T extends permissions.PermissionReq,
 >(
 	req: T,
-	...extraPermissions: string[]
+	extraPermissions: string[],
+	deepCloneRequest = true,
 ): T => {
-	const augmentedReq = _.clone(req);
+	const augmentedReq = deepCloneRequest ? _.clone(req) : req;
```
Page-, line 341: `_.clone` isn't a deep clone so the variable naming is confusing, this is the difference between mutating the original request (and happening to return it) vs returning a new augmented one. If you want to mutate the original request then just doing
```ts
req.apiKey?.permissions?.push(...extraPermissions);
```
would do the same thing without additional unnecessary work

thgreasi: You have point about the naming, so I will rename the parameter to `mutateRequest = false` and add some comments on why we have it.
We still do need to deep clone the apiKey & its permissions though afaict, since otherwise doing the push directly would pollute pine's `apiKeyPermissions` cache.

src/features/api-keys/lib.ts
```diff
@@ -335,10 +335,27 @@ export const augmentReqApiKeyPermissions = <
...
+	let augmentedReq = req;
+
+	if (!mutateRequestObject) {
+		augmentedReq = _.clone(req);
+		augmentedReq.apiKey = _.cloneDeep(augmentedReq.apiKey);
+	} else if (augmentedReq.apiKey?.permissions) {
+		// When mutateRequestObject === true we still need to clone
+		// the permissions array rather than modifying it directly
+		// so that we do not pollute pine's apiKeyPermissions cache.
+		augmentedReq.apiKey.permissions = augmentedReq.apiKey.permissions.slice();
```
Page-, line 356: I'd normally clone via
```suggestion
		augmentedReq.apiKey.permissions = [...augmentedReq.apiKey.permissions];
```

Page- (commented):
Looks fine other than minor comment

thgreasi:
I self-certify!

PR #1385, opened 2023-07-26
Add /actor/v1/whoami endpoint

otaviojacobi:
Change-type: minor
See: https://balena.zulipchat.com/#narrow/stream/345746-aspect.2Fproduct/topic/Role.20Permissions.20in.20balenaCloud.20Orgs

### DIFF
--- a/src/features/auth/whoami.ts
+++ b/src/features/auth/whoami.ts
@@ -3,11 +3,57 @@ import type { RequestHandler } from 'express';
 import { sbvrUtils, permissions } from '@balena/pinejs';
 
 import { getUser } from '../../infra/auth/auth';
-import { captureException, handleHttpErrors } from '../../infra/error-handling';
+import {
+	captureException,
+	handleHttpErrors,
+	ThisShouldNeverHappenError,
+} from '../../infra/error-handling';
 
-import type { User } from '../../balena-model';
+import type { User, Application, Device } from '../../balena-model';
+import { UnauthorizedError } from '@balena/pinejs/out/sbvr-api/errors';
 
 const { api } = sbvrUtils;
+
+type ExpandedActor =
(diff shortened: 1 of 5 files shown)


src/features/auth/whoami.ts
```diff
@@ -3,12 +3,58 @@ import type { RequestHandler } from 'express';
...
+			actorTypeId: number;
+			slug: string;
+	  }
+	| {
+			id: number;
+			actorType: 'device';
+			actorTypeId: number;
+			uuid: string;
+	  };
+
```
otaviojacobi, line 57: This interface is the only thing that bothers me a bit in the sense that the ` /user/v1/whoami` endpoint also has an "id" field and this one has an "id" as well, but they will differ as one is the user id and other the actor id. Maybe change the interface to explicitly be `actorId` instead of `id`? Any opinions are welcomed

src/features/auth/whoami.ts
```diff
@@ -62,3 +108,98 @@ export const whoami: RequestHandler = async (req, res) => {
...
+			if (actorId == null) {
+				throw new UnauthorizedError(
+					'Request API Key or Token has no associated actor',
+				);
+			}
+
+			return (await api.resin.get({
+				resource: 'actor',
+				passthrough: { req: permissions.root, tx },
```
Page-, line 125: I don't think doing this with root permissions is good if it's an api key associated to a user but which would not be expected to be able to get the username/email

otaviojacobi: Thanks @Page- ! This is a nice catch. I added some logic before it similar to the user whoami endpoint to validate that when this is an user key, the user has the needed permissions before calling resource: 'actor' with root permissions. I also added tests for it

PR #1386, opened 2023-08-10
patch: Replace legacy config RESIN_OVERRIDE_LOCK

vipulgupta2048:
Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipulgupta2048@gmail.com>


vipulgupta2048:
Thinking about this again, if there is a reason we have chosen to keep RESIN_OVERRIDE_LOCK around then I can handle this one on the docs with a simple if condition. LMK

thgreasi:
@vipulgupta2048 it looks like the SUPERVISOR_OVERRIDE_LOCK being in the block list and having the bC-api & the sdk using the `RESIN_OVERRIDE_LOCK`  was intentional, so I think that we should (at least for now) stick with `RESIN_OVERRIDE_LOCK`
See: https://github.com/balena-io/open-balena-api/blob/254475e12c759d029273a76b5eebc35ac731935c/src/features/vars-schema/env-vars.ts#L48-L57
See: https://github.com/balena-io/balena-api/blob/7e6d06fe8e75d7aee5a0f0079e0f21794acc7e14/src/features/devices/routes/lock-override.ts#L6
See: https://github.com/balena-io/balena-sdk/blob/4e54ae8a0daebbf85bd2ba48c9a522a96f5803ce/src/models/device.ts#L90

PR #1390, opened 2023-08-14
CI: Use squawk to validate SQL

joshbwlng:
Change-type: patch

---

Recently binary builds `libpg-query` during `npm ci` have been failing. Use `squawk` instead of `libpg-query` to avoid building in the first place.


joshbwlng:
Hmm, probably missing a system dependency:
```
5852251822-sut-1    |      Error: Invalid SQL: /usr/src/app/node_modules/squawk-cli/js/binaries/squawk: 1: cannot open g5: No such file
5852251822-sut-1    | /usr/src/app/node_modules/squawk-cli/js/binaries/squawk: 2: ��: not found
5852251822-sut-1    | /usr/src/app/node_modules/squawk-cli/js/binaries/squawk: 3: Syntax error: word unexpected (expecting ")")
```

test/10_migrations.ts
```diff
@@ -1,10 +1,24 @@
...
+// Validate SQL files using squawk
+function validateSql(files: string[]): void {
+	try {
+		execSync(
+			`npx squawk --pg-version=15.0 --dump-ast parsed ${files.join(' ')}`,
+			{
+				stdio: ['ignore', 'ignore', 'pipe'],
+			},
+		);
+	} catch (e) {
```
fisehara, line 17: If it fails, will it highlight the name of the migration?

fisehara (commented):
LGTM
Wondering how to identify the failing migration file.

test/10_migrations.ts
joshbwlng: Responed in https://github.com/balena-io/balena-api/pull/4553#discussion_r1293174594

joshbwlng:
Closing for now as it doesn't seem to affect our CI, only machines (like my own) that have newer versions of glibc. Ubuntu 22.04 (assuming current default glibc install of 2.35) should be fine which includes our CI instances.

PR #1391, opened 2023-08-17
CI: Use pglast to validate SQL

joshbwlng:
Change-type: patch

---

Replace `libpg-query` with `pglast` for verifying SQL during CI. The main reason is due to `libpg-query` failing to build during `npm install`: https://github.com/balena-io/balena-api/actions/runs/5887828250/job/15967940338?pr=4528


joshbwlng:
Moving to https://github.com/balena-io/open-balena-api/pull/1395

PR #1394, opened 2023-08-17
Update /os/v1/config endpoint for config.json migration

cywang117:
See: https://balena.fibery.io/Work/Improvement/os-config-improving-the-interface-for-config.json-modification-901
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/os-config.20v2.20endpoint
Change-type: minor

### DIFF
--- a/src/features/os-config/index.ts
+++ b/src/features/os-config/index.ts
@@ -1,27 +1,45 @@
-import type { Application, RequestHandler } from 'express';
+import type { Application } from 'express';
 
 import {
 	DEVICE_CONFIG_OPENVPN_CA,
 	DEVICE_CONFIG_OPENVPN_CONFIG,
 	DEVICE_CONFIG_SSH_AUTHORIZED_KEYS,
+	LOGS_HOST,
 } from '../../lib/config';
 import { b64decode } from '../../lib/utils';
 
-const getOsConfiguration: RequestHandler = (_req, res) => {
-	res.json({
-		services: {
-			openvpn: {
-				config: DEVICE_CONFIG_OPENVPN_CONFIG,
-				ca: b64decode(DEVICE_CONFIG_OPENVPN_CA),
-			},
-			ssh: {
-				authorized_keys: DEVICE_CONFIG_SSH_AUTHORIZED_KEYS,
-			},
-		},
-		schema_version: '1.0.0',
-	});
+// OS service configurations
+const services = {
+	openvpn: {
+		config: DEVICE_CONFIG_OPENVPN_CONFIG,
+		ca: b64decode(DEVICE_CONFIG_OPENVPN_CA),
+	},
+	ssh: {
+		authorized_keys: DEVICE_CONFIG_SSH_AUTHORIZED_KEYS,
+	},
+};
+
+// Config.json migrations: Changes should be evaluated for security risks before applying.
+//
+// - A field may not be deleted from config.json.
+// - A field with a value of non-null tells os-config that the value will be updated
+//   to the new value if it's different.
+// - A field not found in the whitelist of the os-config schema will be ignored.
+const config = {
+	overrides: {
+		...(LOGS_HOST != null && { logsEndpoint: `https://${LOGS_HOST}` }),
+	},
 };
 
 export const setup = (app: Application) => {
-	app.get('/os/v1/config/', getOsConfiguration);
+	app.get('/os/v1/config/', (_req, res) => {
+		res.json({
+			services,
+			// Older os-configs don't know how to handle the config field, but
+			// luckily serde-rs ignores unknown fields by default.
+			config,
+			/** @deprecated schema_version is an outdated field kept for compatibility with legacy os-configs */
+			schema_version: '1.0.0',
+		});
+	});
 };
--- a/test/22_os-config.ts
+++ b/test/22_os-config.ts
@@ -1,0 +1,35 @@
+import { expect } from 'chai';
+
+import { supertest } from './test-lib/supertest';
+import { LOGS_HOST } from '../src/lib/config';
+
+describe('OS configuration endpoints', () => {
+	describe('/os/v1/config', () => {
+		it('should return a valid JSON response', async () => {
+			const { body } = await supertest().get('/os/v1/config').expect(200);
+
+			// Service configurations should be valid for their respective services
+			expect(body)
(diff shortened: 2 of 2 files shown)


src/features/os-config/index.ts
```diff
@@ -1,27 +1,45 @@
...
+	app.get('/os/v1/config/', (_req, res) => {
+		res.json({
+			services,
+			schema_version: '1.0.0',
+		});
+	});
+	app.get('/os/v2/config', (_req, res) => {
+		res.json({
+			services,
+			config,
```
joshbwlng, line 42: Is it fine to not have `schema_version` defined here as well?

src/features/os-config/index.ts
```diff
@@ -1,27 +1,45 @@
...
+	},
+	ssh: {
+		authorized_keys: DEVICE_CONFIG_SSH_AUTHORIZED_KEYS,
+	},
+};
+
+// Config.json migrations: changes should be evaluated for security risks before applying.
+const config = {
+	// config.json fields to add or change
+	update: {
```
thgreasi, line 25: How about calling this field `overrides`?

src/features/os-config/index.ts
```diff
@@ -1,27 +1,45 @@
...
+		logsEndpoint: LOGS_HOST != null ? `https://${LOGS_HOST}` : undefined,
+	},
+	// config.json fields to delete
+	delete: [],
```
thgreasi, line 29: Could such fields be formulated like `update`s or does `null` have value? Eg
```
update: {
  myFieldToRemove: null
}
```

src/features/os-config/index.ts
cywang117: I like this suggestion! I'm modifying the improvement to move to a target-state based approach for API directives to os-config, and will use this as well as `null / None` for fields to be deleted. In the meantime, and before the arch call, I'll move this back to draft status temporarily. Thank you for looking at the code @thgreasi !

src/features/os-config/index.ts
cywang117: After discussing the API changes in an architecture sync, we agreed to keep the legacy schema_version field for now and not introduce a new /os/v2/config endpoint

src/features/os-config/index.ts
cywang117: I've removed the update/delete directives in favor of `overrides`, and also removed the possibility of deleting fields from config.json.

test/22_os-config.ts
```diff
@@ -0,0 +1,24 @@
...
+
+			expect(body)
+				.to.have.property('services')
+				.that.has.all.keys('openvpn', 'ssh');
+			expect(body).to.have.property('schema_version').that.equals('1.0.0');
+			expect(body)
+				.to.have.property('config')
+				.that.deep.equals({
+					overrides: {
+						logsEndpoint: LOGS_HOST ? `https://${LOGS_HOST}` : undefined,
```
thgreasi, line 19: Does this test pass if LOGS_HOST isn't defined?
I would expect it to error w/ something like `expected { logEndpoint: undefined } but got {}`.
A way to make this test both cases could be to have two separate `describe`s and changing the LOGS_HOST value in a `before` in each of them.
See: https://github.com/balena-io/balena-api/blob/7e6d06fe8e75d7aee5a0f0079e0f21794acc7e14/test/01_user.ts#L265-L267

src/features/os-config/index.ts
```diff
@@ -1,27 +1,46 @@
...
 
 export const setup = (app: Application) => {
-	app.get('/os/v1/config/', getOsConfiguration);
+	app.get('/os/v1/config/', (_req, res) => {
+		res.json({
+			services,
+			// Older os-configs don't know how to handle the config field, but
+			// luckily serde-rs ignores unknown fields by default.
+			config,
+			// schema_version is an outdated field that is kept to be compatible
```
thgreasi, line 41: @cywang117 perhaps we could use `/** @deprecated <comment> */` but no big deal :)

test/22_os-config.ts
```diff
@@ -0,0 +1,24 @@
...
+import { LOGS_HOST } from '../src/lib/config';
+
+describe('OS configuration endpoints', () => {
+	describe('/os/v1/config', () => {
+		it('should return a valid JSON response', async () => {
+			const { body } = await supertest().get('/os/v1/config').expect(200);
+
+			expect(body)
+				.to.have.property('services')
+				.that.has.all.keys('openvpn', 'ssh');
```
thgreasi, line 13: How about moving here the respective test case of the bC-api?
See: https://github.com/balena-io/balena-api/blob/master/test/27_os.ts#L10-L11

thgreasi (commented):
LGTM just NITs (but note that I wasn't in the arch call).

test/22_os-config.ts
cywang117: This test does indeed pass even with `LOGS_HOST` undefined, with overrides being equal to `{}`. 

Do you feel strongly about adding test mock env var handling to the open-balena-api codebase? It doesn't feel entirely comfortable to me as it's a branch in the production code written specifically for tests. If you do feel strongly, I'm fine with it @thgreasi :)

src/features/os-config/index.ts
```diff
@@ -1,27 +1,46 @@
...
+
+// Config.json migrations: Changes should be evaluated for security risks before applying.
+//
+// - A field may not be deleted from config.json.
+// - A field with a value of non-null tells os-config that the value will be updated
+//   to the new value if it's different.
+// - A field not found in the whitelist of the os-config schema will be ignored.
+const config = {
+	overrides: {
+		logsEndpoint: LOGS_HOST != null ? `https://${LOGS_HOST}` : undefined,
```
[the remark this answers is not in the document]
cywang117: Done [here](https://github.com/balena-io/open-balena-api/pull/1394/files#diff-901c32d1753c4de62070d6de17728d16ad1302a806fc2050d1e7360a0bab8049R30) 👍🏼

test/22_os-config.ts
cywang117: @thgreasi friendly reminder :) if you don't feel strongly about this, I'm going to go ahead and resolve both comments in 24h and merge

cywang117:
@balena-ci I self-certify!

PR #1396, opened 2023-08-18
Update @balena/lint to 7.0.0

thgreasi:
Update @balena/lint from 6.2.2 to 7.0.0

Depends-on: https://github.com/balena-io-modules/node-balena-lint/pull/85
Depends-on: https://github.com/balena-io-modules/abstract-sql-to-typescript/pull/40
Change-type: patch

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -137,7 +137,8 @@ async function onInitHooks() {
 		},
 	});
 
-	// this will pre-fetch the device types and populate the cache...
+	// Pre-fetch the device types and populate the cache w/o blocking the API startup
+	// eslint-disable-next-line @typescript-eslint/no-floating-promises
 	getAccessibleDeviceTypes(sbvrUtils.api.resin);
 
 	await sbvrUtils.db.transaction((tx) =>
@@ -256,4 +257,5 @@ const init = async () => {
 		process.exit(1);
 	}
 };
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
 init();
--- a/src/features/device-proxy/device-proxy.ts
+++ b/src/features/device-proxy/device-proxy.ts
@@ -63,7 +63,7 @@ const validateSupervisorResponse = (
 			}
 			res.status(statusCode).json(jsonBody);
 		} else if (/^text\/(plain|html)/.test(contentType)) {
-			if (/^([A-Za-z0-9\s:'\.\?!,\/-])*$/g.test(body)) {
+			if (/^([A-Za-z0-9\s:'.?!,/-])*$/g.test(body)) {
 				res.status(statusCode).set('Content-Type', 'text/plain').send(body);
 			} else {
 				badSupervisorResponse(req, res, filter, 'Invalid TEXT data');
--- a/src/features/device-proxy/device-proxy.ts
+++ b/src/features/device-proxy/device-proxy.ts
@@ -132,10 +132,8 @@ export const proxy = async (req: Request, res: Response) => {
 		if (handleHttpErrors(req, res, err)) {
 			return;
 		}
-		if (err?.body != null) {
-			err = err.body;
-		}
-		res.status(502).send(translateError(err));
+		const errorToReturn = err?.body ?? err;
+		res.status(502).send(translateError(errorToReturn));
 	}
 };
 
--- a/src/features/device-state/routes/state-get-v2.ts
+++ b/src/features/device-state/routes/state-get-v2.ts
@@ -58,8 +58,9 @@ export type StateV2 = {
 		};
 	};
 	dependent: {
-		apps: {};
-		devices: {};
+		// Empty objects since we dropped support for dependent devices
+		apps: Record<string, never>;
+		devices: Record<string, never>;
 	};
 };
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -440,8 +440,7 @@ function setupMiddleware(app: Application) {
 	);
 	app.use(AUTH_PATH, cookieParser());
 
-	const JSON_REGEXP =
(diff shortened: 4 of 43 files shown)


tsconfig.json
```diff
@@ -6,6 +6,7 @@
 	"include": [
 		"src/**/*.ts",
 		"typings/**/*.d.ts",
+		"*.js",
```
Page-, line 9: Why is this necessary?

test/13_loki-backend.ts
```diff
@@ -111,7 +111,7 @@ describe('loki backend', () => {
 		const loki = new LokiBackend();
 		await new Bluebird(async (resolve) => {
 			let countLogs = 0;
-			loki.subscribe(ctx, () => {
+			await loki.subscribe(ctx, () => {
```
Page-, line 114: I feel like this might affect the timings of things

src/features/device-state/routes/state-get-v2.ts
```diff
@@ -58,8 +58,8 @@ export type StateV2 = {
 		};
 	};
 	dependent: {
-		apps: {};
-		devices: {};
+		apps: object;
+		devices: object;
```
Page-, line 62: Fwiw I think these should be specifically an empty object, I don't know off the top of my head how you type that?

src/features/device-proxy/device-proxy.ts
```diff
@@ -132,10 +132,8 @@ export const proxy = async (req: Request, res: Response) => {
 		if (handleHttpErrors(req, res, err)) {
 			return;
 		}
-		if (err?.body != null) {
-			err = err.body;
-		}
-		res.status(502).send(translateError(err));
+		const errorToReturn = err?.body != null ? err.body : err;
```
Page-, line 135: Am I right in thinking this can be
```suggestion
		const errorToReturn = err?.body?? err;
```
?

test/13_loki-backend.ts
thgreasi: I though that given that we already have a `setTimeout(100)` right below to "wait for a bit" it would be fine, but I'm fine keeping it as is for this PR :+1:

tsconfig.json
thgreasi: The TypeScript-ESLint plugin can only lint files that are part of the tsconfig.json (or any other tsconfig.something.json that we might be pointing it to), otherwise it errors with:
```
error  Parsing error: ESLint was configured to run on `<tsconfigRootDir>/index.js` using `parserOptions.project`: <tsconfigRootDir>/tsconfig.json
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file
```

test/13_loki-backend.ts
```diff
@@ -99,6 +99,7 @@ describe('loki backend', () => {
 		const loki = new LokiBackend();
 		const log = createLog();
 		const incomingLog = await new Bluebird(async (resolve) => {
+			// eslint-disable-next-line @typescript-eslint/no-floating-promises
 			loki.subscribe(ctx, resolve);
```
fisehara, line 103: Why woudln't await make sense here? Would it even make the `await setTimeout(100)` obsolete?

test/13_loki-backend.ts
```diff
@@ -111,6 +112,7 @@ describe('loki backend', () => {
 		const loki = new LokiBackend();
 		await new Bluebird(async (resolve) => {
 			let countLogs = 0;
+			// eslint-disable-next-line @typescript-eslint/no-floating-promises
```
fisehara, line 115: See above

init.ts
```diff
@@ -138,6 +138,7 @@ async function onInitHooks() {
 	});
 
 	// this will pre-fetch the device types and populate the cache...
+	// eslint-disable-next-line @typescript-eslint/no-floating-promises
 	getAccessibleDeviceTypes(sbvrUtils.api.resin);
```
fisehara, line 142: Wouldn't awaiting here create a guarantee that we may need?

fisehara (commented):
LGTM 
NIT two `await import` candiates.
NIT two await candidates

test/13_loki-backend.ts
thgreasi: I actually had this await originally but @Page- was concerned about it affecting the timings, so I decided to keep this a no-op.
See: https://github.com/balena-io/open-balena-api/pull/1396#discussion_r1301675512

init.ts
thgreasi: We don't need to await this, since slow or non-accessible S3 shouldn't block the API startup.
As per the pre-existing comment `// this will pre-fetch the device types and populate the cache...` this is here only to warm-up the device-type.json cache.

src/index.ts
```diff
@@ -478,6 +477,7 @@ async function runCommand(
 	cmd: string,
 	argv: string[],
 ): Promise<void> {
+	// eslint-disable-next-line @typescript-eslint/no-var-requires
 	const script = require(path.join(__dirname, 'commands', cmd));
```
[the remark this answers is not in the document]
thgreasi: It seems that we no longer use this function and we can drop it since there is only a dymmy command atm
See: https://github.com/balena-io/open-balena-api/pull/22/files

init.ts
thgreasi: Let me update the comment

fisehara (approved):

PR #1407, opened 2023-08-25
fix: let go request with `x-forwarded-proto`

jianhu:
let go request with `x-forwarded-proto` header equals to 'https', which mostly comes from lb.

without this, we will have to config `trust proxy`, which seems likes not that convenient considering changing IPs in container environment.


Page-:
You can set `trust proxy` with a number that is the number of hops you're willing to trust rather than using specific ips, you can also use ip ranges rather than specific ips if you want to trust a local range. This PR just allows anyone to set the right header in order to get through and whilst it might be convenient it seems to me at least like it could have security implications that are worth avoiding

PR #1418, opened 2023-08-30
Update @balena/pinejs dependencies

fisehara:
Update @balena/pinejs from 15.3.0 to 15.3.4

Change-type: patch


fisehara:
I self-certify!

PR #1430, opened 2023-09-14
Rate limit getting device logs

fisehara:
Opening a device logs stream or retrieving restfull device logs is not rate limited.

Rate limiter distinguishes between stream and document

Change-type: minor


src/features/device-logs/index.ts
```diff
@@ -22,6 +22,18 @@ const deviceLogsRateLimiter = createRateLimitMiddleware(
 	},
 );
 
+// Rate limit for device log creation, a maximum of 15 batches every 10 second window
+const streamableDeviceLogsRateLimiter = createRateLimitMiddleware(
+	createRateLimiter('get-device-logs', {
+		points: 10, // allow 10 device log streams / requests batches
+		blockDuration: 60, // seconds
+		duration: 60, // reset counter after 10 seconds (from the first batch of the window)
```
joshbwlng, line 30: The comment of "10 seconds" doesn't match the value of "60". Forgot to update copy pasted comment?

test/06_device-log.ts
```diff
@@ -222,4 +221,53 @@ describe('device log', () => {
 			'streamed log line 1',
 		]);
 	});
+
+	it('should rate limit stream-read device', async () => {
+		const dummyLogs = [createLog({ message: `not rate limited` })];
```
joshbwlng, line 226:
```suggestion
		const dummyLogs = [createLog({ message: 'not rate limited' })];
```
Nit

test/06_device-log.ts
```diff
@@ -222,4 +221,53 @@ describe('device log', () => {
...
+		}
+
+		const notLimited = await testRatelimitedDeviceLogsStream();
+		expect(notLimited?.['message']).to.deep.equal(dummyLogs[0].message);
+
+		while ((await testRatelimitedDeviceLogsStream()) !== 'Too Many Requests') {
+			// no empty block
+		}
+		const reateLimited = await testRatelimitedDeviceLogsStream();
+		expect(reateLimited).to.be.string('Too Many Requests');
```
joshbwlng, line 271:
```suggestion
		const rateLimited = await testRatelimitedDeviceLogsStream();
		expect(rateLimited).to.be.string('Too Many Requests');
```

test/06_device-log.ts
```diff
@@ -222,4 +221,53 @@ describe('device log', () => {
...
+				.parse(function (res, callback) {
+					res.on('data', async function (chunk) {
+						const parsedChunk = JSON.parse(Buffer.from(chunk).toString());
+						evalValue = parsedChunk;
+						// if data stream provides proper data terminate the stream (abort)
+						if (typeof parsedChunk === 'object') {
+							if (parsedChunk?.message === 'not rate limited') {
+								req.abort();
+							}
+						}
```
joshbwlng, line 249:
```suggestion
						if (typeof parsedChunk === 'object' && parsedChunk?.message === 'not rate limited') {
							req.abort();
						}
```
Nit. Also, arrays parse as `object` when tested with `typeof`, but this is just a test with known data shapes so probably don't need to worry about that too much.

test/06_device-log.ts
```diff
@@ -2,7 +2,6 @@ import _ from 'lodash';
 import { expect } from 'chai';
 import * as fixtures from './test-lib/fixtures';
 import { supertest } from './test-lib/supertest';
-
```
joshbwlng, line 5: Maybe a mistaken line deletion? I like keeping imports visually separated.

src/features/device-logs/index.ts
fisehara: Thx!

test/06_device-log.ts
fisehara: :+1:

test/06_device-log.ts
fisehara: :+1:

test/06_device-log.ts
fisehara: :+1:

test/06_device-log.ts
fisehara: :+1:

src/infra/rate-limiting/index.ts
```diff
@@ -126,14 +126,22 @@ const $createRateLimitMiddleware = (
...
-		if (typeof field === 'function') {
-			fieldFn = field;
+	if (fields != null) {
+		if (typeof fields === 'function') {
+			fieldFn = fields;
+		} else if (Array.isArray(fields)) {
+			fieldFn = (req) =>
+				fields
+					.map((field) => {
+						const path = _.toPath(field);
```
Page-, line 139: This `_.toPath` must occur only once in the setup phase as it is fairly costly when in very hot code (eg in rate limiting code that potentially runs for every request). If we didn't care about pre-calculating the path in order to do it only once then we would instead do `_.get(req, field);` and let lodash do it automatically each time

Page- (changes requested):

fisehara:
Rate limiter is in place, during tests rate limiting is intentionally relaxed. Which enforcing strict rate limiting the endpoints get limited, generously but limited.

PR #1432, opened 2023-09-19
Support extending the device fields populated in the "get-state" event

thgreasi:
Change-type: minor
See: https://balena.zulipchat.com/#narrow/stream/345882-_help/topic/Public.20IP.20not.20updating.20in.20the.20dashboard

### DIFF
--- a/src/features/device-state/index.ts
+++ b/src/features/device-state/index.ts
@@ -3,6 +3,7 @@ import type StrictEventEmitter from 'strict-event-emitter-types';
 
 import { EventEmitter } from 'events';
 import { middleware } from '../../infra/auth';
+import type { Device } from '../../balena-model';
 
 import { resolveOrDenyDevicesWithStatus } from './middleware';
 import { stateV2 } from './routes/state-get-v2';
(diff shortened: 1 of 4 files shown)


joshbwlng:
@thgreasi I think everything looks fine, but let me double-check my understanding.

We will be moving away from hardcoding device fields returned when checking a device state. Instead, we will allow consumers to extend the field list as they need. A specific example of how this will be used is in https://github.com/balena-io/balena-api/pull/4620 right?

thgreasi:
That's correct more or less @joshbwlng .
More precisely, the need we have is to enable consumers to extend the fields that we pre-fetch from the DB.
As a result:
* The custom typed `storedPublicAddress` props format for the event payload seems a not convenient any more, hence I deprecated it in favor of a prop that would hold all retrieved fields w/o changing their name.
* Consumers will be able to `getStateEventAdditionalFields.push` more device fields that they are interested at in the respective array. (eg https://github.com/balena-io/balena-api/pull/4620 as you saw)
* Atm `public_address` is fetched, but isn't really used by the oB-api, so it seems it was hardcoded b/c of bC needs, hence I moved it in the `getStateEventAdditionalFields` list to make that clear and added a TODO to remove it from there in the next major of oB (since oB doesn't need it).

joshbwlng:
@thgreasi Cool, that confirms my understanding. Is there a reason you haven't added any testing to prove the extending and fetching work as expected form the perspective of a consumer?

thgreasi:
I guess I b/c there are no tests atm in oB wrt that event, and we atm only have them in bC :(

thgreasi:
I self-certify!

PR #1434, opened 2023-09-19
New CONTRACT_ALLOWLIST option to limit device types/contracts

shaunco:
Resolves: #1433
Change-type: minor


src/lib/config.ts
```diff
@@ -82,6 +82,24 @@ export const AUTH_RESINOS_REGISTRY_CODE = optionalVar(
...
+ * null: include all device type and device contract slugs
+ * "x;y;z": include only the specified device type and contract slugs - note that you MUST list
+ *          all dependent slugs as well so for hw.device-type/asus-tinker-board-s you would need
+ *          `armv7hf;asus;tinkerboard;asus-tinker-board-s` to pick up arch.sw/armv7hf,
+ *          hw.device-manufacturer/asus, hw.device-family/tinkerboard, and hw.device-type/asus-tinker-board-s
+ * 		    For something like hw.device-type/iot-gate-imx8 you would need `aarch64;iot-gate-imx8`
+ * 			to pick up arch.sw/aarch64 and hw.device-type/iot-gate-imx8
+ *          (the order of the slugs in this variable does not matter)
+ */
+export const INCLUDED_DEVICETYPE_SLUGS = new Map(
```
Page-, line 96: Is there a reason you used a `Map` over a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)? It seems like a `Set` would be the better option for this use case where you only care if the slug exists or not

src/lib/config.ts
```diff
@@ -82,6 +82,24 @@ export const AUTH_RESINOS_REGISTRY_CODE = optionalVar(
...
+	(optionalVar('INCLUDED_DEVICETYPE_SLUGS') ?? '')
+		.split(';')
+		.filter((slug) => slug.trim().length > 0)
+		.map((slug) => [slug.trim(), true]),
```
thgreasi, line 100: NIT: Along with Page's suggestion on using a `Set`, we can also remove the duplicate `.trim()` by re-ordering the `.map` & `.filter`
```suggestion
		.map((slug) => slug.trim())
		.filter((slug) => slug.length > 0),
```

src/features/contracts/contracts-directory.ts
```diff
@@ -177,13 +178,34 @@ export const getContracts = async (type: string): Promise<Contract[]> => {
...
+	let contractFiles = await glob(
 		`${CONTRACTS_BASE_DIR}/**/contracts/${type}/**/*.json`,
 	);
 	if (!contractFiles.length) {
 		return [];
 	}
 
+	// If there are explicit includes, then everything else is excluded so we need to
+	// filter the contractFiles list to include only contracts that are in the INCLUDED_DEVICETYPE_SLUGS map
+	if (INCLUDED_DEVICETYPE_SLUGS.size > 0) {
```
thgreasi, line 190: If we want to be 100% accurate (even though atm there is no clash, so there isn't any issue), I think this should be:
```suggestion
	if (type === 'hw.device-type' && INCLUDED_DEVICETYPE_SLUGS.size > 0) {
```

src/features/contracts/contracts-directory.ts
```diff
@@ -177,13 +178,34 @@ export const getContracts = async (type: string): Promise<Contract[]> => {
...
+		const before = contractFiles.length;
+		contractFiles = contractFiles.filter((file) => {
+			// Get the contract slug from the file path
+			const match = file.match(slugRegex);
+			if (!match || match.length < 2) {
+				return false;
+			}
+
+			// Check if this slug is included in the map
+			return INCLUDED_DEVICETYPE_SLUGS.has(match[1]);
```
thgreasi, line 201:
```suggestion
			const deviceTypeSlug = file.match(slugRegex)?.[1];
			if (!deviceTypeSlug) {
				return false;
			}

			// Check if this slug is included in the map
			return INCLUDED_DEVICETYPE_SLUGS.has(deviceTypeSlug);
```

src/features/contracts/contracts-directory.ts
```diff
@@ -177,13 +178,34 @@ export const getContracts = async (type: string): Promise<Contract[]> => {
...
 		`${CONTRACTS_BASE_DIR}/**/contracts/${type}/**/*.json`,
 	);
 	if (!contractFiles.length) {
 		return [];
 	}
 
+	// If there are explicit includes, then everything else is excluded so we need to
+	// filter the contractFiles list to include only contracts that are in the INCLUDED_DEVICETYPE_SLUGS map
+	if (INCLUDED_DEVICETYPE_SLUGS.size > 0) {
+		const slugRegex = new RegExp(`/contracts/${type}/([^/]*)/`);
```
thgreasi, line 191: Shouldn't this be v ?
```suggestion
		const slugRegex = new RegExp(`/contracts/${type}/([^/]+)/`);
```

thgreasi (commented):
:heart:

src/lib/config.ts
shaunco: Old habits... will update

src/lib/config.ts
```diff
@@ -82,6 +82,24 @@ export const AUTH_RESINOS_REGISTRY_CODE = optionalVar(
...
+ * "x;y;z": include only the specified device type and contract slugs - note that you MUST list
+ *          all dependent slugs as well so for hw.device-type/asus-tinker-board-s you would need
+ *          `armv7hf;asus;tinkerboard;asus-tinker-board-s` to pick up arch.sw/armv7hf,
+ *          hw.device-manufacturer/asus, hw.device-family/tinkerboard, and hw.device-type/asus-tinker-board-s
+ * 		    For something like hw.device-type/iot-gate-imx8 you would need `aarch64;iot-gate-imx8`
+ * 			to pick up arch.sw/aarch64 and hw.device-type/iot-gate-imx8
+ *          (the order of the slugs in this variable does not matter)
+ */
+export const INCLUDED_DEVICETYPE_SLUGS = new Map(
+	(optionalVar('INCLUDED_DEVICETYPE_SLUGS') ?? '')
```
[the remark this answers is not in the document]
shaunco: thanks! I had missed that `optionalVar` had an overload that returned just a `string` (instead of `string | undefined`) when a default value was passed! Will update

src/features/contracts/contracts-directory.ts
shaunco: for my use, I was looking to filter `arch.sw`, `hw.device-manufacturer`, `hw.device-family`, and `hw.device-type` contracts to the absolute minimum set my deployment needed, which is `aarch64;iot-gate-imx8;generic-aarch64;amd64;generic-amd64`. Limiting this to just `hw.device-type` contracts would pull in 4 `arch.sw` contracts I don't need (armv7hf, i386-nlp, i386, and rpi), 2 `hw.device-manufacturer` contracts I don't need (aaeon and asus), and 2 `hw.device-family` contracts I don't need (tinkerboard and upboard).

shaunco:
Thanks @Page- and @thgreasi for the rapid review and fantastic suggestions! I committed all but one of the suggestions and resolved the associated review items.

@thgreasi - one remaining review item that I left a comment on.

src/features/contracts/contracts-directory.ts
shaunco: I had originally called the config variable `INCLUDED_CONTRACT_AND_DEVICETYPE_SLUGS`, but it felt way too long - but I can see how leaving it as just `DEVICETYPE_SLUGS` makes the filtering of `arch.sw`, `hw.device-manufacturer`, and `hw.device-family` contract filtering awkward. I tried to make up for it with a good description of behavior in config.ts.

It seemed unlikely that there would be clashes between the various contract folders, and device-type-lists.ts filtering needed just the slug without the contract-type prefix, so I went with bare slugs instead of needing both with something like: `arch.sw/amd64;hw.device-type/generic-arm64;generic-amd64`

src/features/contracts/contracts-directory.ts
```diff
@@ -177,13 +178,34 @@ export const getContracts = async (type: string): Promise<Contract[]> => {
...
 		`${CONTRACTS_BASE_DIR}/**/contracts/${type}/**/*.json`,
 	);
 	if (!contractFiles.length) {
 		return [];
 	}
 
+	// If there are explicit includes, then everything else is excluded so we need to
+	// filter the contractFiles list to include only contracts that are in the INCLUDED_DEVICETYPE_SLUGS map
+	if (INCLUDED_DEVICETYPE_SLUGS.size > 0) {
+		const slugRegex = new RegExp(`/contracts/${type}/([^/]+)/`);
```
Page-, line 191: It'd probably be good to escape `type` to avoid any potential matching issues
```suggestion
		const slugRegex = new RegExp(`/contracts/${_.escapeRegExp(type)}/([^/]+)/`);
```
I know that in practice it probably won't make a difference but I'd rather play it safe

Page- (commented):
For my part I am happy but would like @thgreasi to review his comments as it's still a valid/open thing

src/features/contracts/contracts-directory.ts
shaunco: @thgreasi ?

Page-:
I've pinged @thgreasi again so he should take a look soon

src/features/contracts/contracts-directory.ts
thgreasi: Hi @shaunco and I'm sorry for my late response here.
Just renaming the env var to `INCLUDED_CONTRACT_SLUGS` to make it clearer what it refers to should be enough as you mention. eg:
```
INCLUDED_CONTRACT_SLUGS="amd64;armv7hf;generic-amd64;fincm3"
```
I do realize that having clashes sounds like a hard thing to occur, but to be honest I prefer your seconf suggestion about defining the fully qualified slugs of all the contracts that a user cares to be included, which doesn't leave any room for ambiguities. eg:
```
INCLUDED_CONTRACTS="arch.sw/amd64;arch.sw/armv7hf;hw.device-type/generic-arm64;hw.device-type/fincm3"
```

thgreasi: Or maybe `CONTRACTS_ALLOWLIST`, which also gets alphabetically sorted next to all other contract related env vars?

shaunco: Thanks @thgreasi !

I'll go with `CONTRACT_ALLOWLIST`.

For the fully qualified slugs, would you prefer `device-type-lists.ts` to simply strip the `hw.device-type` off of `hw.device-type/generic-amd64` to get the `generic-amd64` slug that it needs, or would you prefer that the env variable must have BOTH `hw.device-type/generic-amd64` AND `generic-amd64` (that is: `CONTRACT_ALLOWLIST="hw.device-type/generic-amd64;generic-amd64"`)?

thgreasi: @shaunco In my view it's enough to have just the fully qualified strings and not any device type slugs alone. ie:
```
INCLUDED_CONTRACTS="arch.sw/amd64;hw.device-type/generic-arm64"
```
Then `device-type-lists.ts` would do something along the lines of:
```ts
slugs = slugs.filter((slug) => CONTRACT_ALLOWLIST.has(`hw.device-type/${slug}`));
```
since the string concatenation should be faster than first finding the contract slugs that start with `hw.device-type/`, then slicing that prefix on a second step, and I guess constructing a new Set with all these on a third step to keep the overall iteration of the listed folders an O(N) operation.

shaunco:
@thgreasi - I resolved your last outstanding suggestion. Thanks for all the feedback!

thgreasi (approved):
:+1: :clap:

PR #1439, opened 2023-09-28
Check user access to the device type provided to /config/vars

thgreasi:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/345889-balena-io.2Fos/topic/balenaOS.20for.20RPi.205.20.5BUnder.20NDA.5D/near/393642691


fisehara (approved):
Thx

PR #1454, opened 2023-10-17
Fix the request.creds typings

thgreasi:
Change-type: patch

### DIFF
--- a/typings/express-extension.d.ts
+++ b/typings/express-extension.d.ts
@@ -1,5 +1,8 @@
 // Augment express.js with balena-specific attributes via declaration merging.
 declare namespace Express {
+	import type { Creds } from '../src/infra/auth/jwt-passport';
+	// For some reason TS doesn't like v so we had to use `import()`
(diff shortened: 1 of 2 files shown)


typings/express-extension.d.ts
```diff
@@ -9,7 +9,7 @@ declare namespace Express {
 	export interface Request {
 		prefetchApiKey?: Resolvable<ApiKey>;
 
-		creds?: Creds;
```
thgreasi, line 12: This was unresolved before and was effectively used as `any`.

typings/express-extension.d.ts
```diff
@@ -9,7 +9,7 @@ declare namespace Express {
...
+		creds?: import('../src').Creds;
```
joshbwlng, line 12:
```suggestion
		creds?: import('../src/infra/auth/jwt-passport').Creds;
```
Just a nit to match the `ApiUser` import above. Personally I'd prefer imports at the top of the file instead of inline to make things more standard, but not necessary I guess.

thgreasi:
I self-certify!

PR #1457, opened 2023-10-18
multi-level-memoizee: Support local & global specific extra Cache options

thgreasi:
Change-type: minor

### DIFF
--- a/src/infra/cache/multi-level-memoizee.ts
+++ b/src/infra/cache/multi-level-memoizee.ts
@@ -16,42 +16,130 @@ type MultiCacheMemoizeeOpts<T extends (...args: any[]) => any> = {
 	max?: MemoizeeOptions<T>['max'];
 } & Pick<MemoizeeOptions<T>, 'preFetch' | 'normalizer'>;
 
-const memoizeeSharedOptionsKeys = ['max', 'maxAge', 'preFetch'] as const;
+const memoizeeExtraOptionsKeys = ['max', 'maxAge', 'preFetch'] as const;
 
-type MultiCacheMemoizeeSharedOpts<T extends (...args: any[]) => any> = Pick<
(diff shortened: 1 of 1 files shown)


fisehara (commented):
LGTM do you need another pair of :eyes: ?

src/infra/cache/multi-level-memoizee.ts
```diff
@@ -16,42 +16,106 @@ type MultiCacheMemoizeeOpts<T extends (...args: any[]) => any> = {
...
+// TODO: Move these to common-types.ts once we make them part of the build output.
+// This makes TS emit a union of `Record<K, T[K]>` for each `keyof T`.
+type ToSinglePropUnions<T> = { [K in keyof T]: Record<K, T[K]> }[keyof T];
+// Requires at least one of the properties of T to be defined aka NonEmptyPartial.
+type AtLeastOneProp<T> = Partial<T> & ToSinglePropUnions<T>;
+
+type ExtraCacheOpts<T extends AnyFunction> =
+	// Deprecated notation for backwards compatibility
+	// TODO: Drop me in the next major and switch AtLeastOneProp to a plain Partial
+	| SharedMultiCacheMemoizeeExtraOpts<T>
```
Page-, line 47: Can see if marking this as deprecated so works/shows up in ide hints and such, ie
```suggestion
	/**
	 * @deprecated
	 */
	| SharedMultiCacheMemoizeeExtraOpts<T>
```
I know it works for top level bits but I'm not 100% if it works on a subset of a larger type or not

thgreasi: I wasn't able to get this deprecation notice to show up, but in any case it looks better in the code rather than a plan "// Deprecated" comment.
In order for the deprecation warning to be surfaced I had to add 2 more overload signatures to `multiCacheMemoizee()`

thgreasi:
I self-certify!

PR #1465, opened 2023-10-23
Update @balena/pinejs dependency

otaviojacobi:
Update @balena/pinejs from 15.3.10 to 15.3.13

Change-type: patch


otaviojacobi:
LGTM

PR #1466, opened 2023-10-26
Draft: Replace `request` with `axios`

fisehara:
Drop in replacement for `request` with `got` is not possible, as `got` no proxy support has. `Axios` comes with proxy support and similar size of package as `got`.

Change-type: major


Page- (commented):
Have you looked at https://github.com/sindresorhus/got/blob/main/documentation/tips.md#proxying ?

fisehara:
@Page- Currently we won't be able to natively use 'got' https://github.com/sindresorhus/got/issues/1789 as `open-balena-api` is using CommonJs.

joshbwlng:
Would it be better to wait until we can use `got` before switching to `axios`? I know there's been a lot of work to move us away from CommonJs.

Issue #1484, opened 2023-11-24
Fix code scanning alert - Type confusion through parameter tampering


fisehara:
Tracking issue for:
- [ ] https://github.com/balena-io/open-balena-api/security/code-scanning/3

PR #1485, opened 2023-11-24
Fix code scanning alert - Type confusion through parameter tampering

fisehara:
Fixes: Fix code scanning alert - Type confusion through parameter tampering #1484

Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/supervisor.ts
+++ b/src/features/device-logs/lib/supervisor.ts
@@ -9,7 +9,7 @@ const MAX_LOGS_PER_BATCH = 10;
 
 export class Supervisor {
 	public convertLogs(logs: SupervisorLog[]): DeviceLog[] {
-		if (logs.length > MAX_LOGS_PER_BATCH) {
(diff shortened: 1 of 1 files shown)


src/features/device-logs/lib/supervisor.ts
```diff
@@ -9,7 +9,7 @@ const MAX_LOGS_PER_BATCH = 10;
 
 export class Supervisor {
 	public convertLogs(logs: SupervisorLog[]): DeviceLog[] {
-		if (logs.length > MAX_LOGS_PER_BATCH) {
+		if (Array.isArray(logs) && logs.length > MAX_LOGS_PER_BATCH) {
```
Page-, line 12: I would prefer this check to be higher up when grabbing `req.body` rather than nested a couple of levels deep where we would ideally expect the typings to be respected

Page-:
Also I'd prefer a better commit message/changelog entry that mentions what the aim is now, eg that we're improving validation of logs sent to the store endpoint

PR #1488, opened 2023-11-30
Update @balena/pinejs to 15.3.28

otaviojacobi:
Update @balena/pinejs from 15.3.17 to 15.3.28

Change-type: patch


otaviojacobi:
LGTM

PR #1491, opened 2023-12-06
Fix timing of log publishing

dcaputo-harmoni:
Logs are frequently not published to redis / loki because the line:

```
buffer.length = 0;
```

in `src/features/device-logs/lib/store.ts` comes before the line:

```
await publishPromise
```

So the publishPromise must complete *very* quickly before the buffer is wiped out otherwise no logs are published.  This can be fixed by simply spreading buffer into a new array, copying the values, and avoiding the timing issue.

Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/store.ts
+++ b/src/features/device-logs/lib/store.ts
@@ -243,7 +243,7 @@ function handleStreamingWrite(
 					});
 				}
 				// Even if the connection was closed, still flush the buffer
-				const publishPromise = publishBackend(backend, ctx, buffer);
+				const publishPromise = publishBackend(backend, ctx, [...buffer]);
 				// Clear the buffer
 				buffer.length = 0;
 				// Resume in case it was paused due to buffering


Page-:
@dcaputo-harmoni have you seen this on redis at all? Because from my understanding all the logs will be mapped through `schema.toBuffer(log)` which will create a copy of the array and mean that it should be fine, loki however had an async step before processing which would cause issues (fix in https://github.com/balena-io/open-balena-api/pull/1494).

Aside from that the faster solution would probably be to change `buffer.length = 0` to `buffer = []` and avoid the array clone; it used to be like that but I previously found `buffer.length = 0` to be a more performant way of clearing the buffer considering we are creating a copy anyway as part of logs processing

Page-:
Yeah I'm fairly confident this should now be solved @dcaputo-harmoni - if you do see the issue again then please raise it but I am pretty confident it's fixed as of 14.15.6

PR #1494, opened 2023-12-12
Avoid any async operations for loki publish before processing logs

Page-:
This means the logs are processed before being cleared as is expected by the interface

Change-type: patch


Page-:
LGTM

PR #1498, opened 2023-12-14
Improve the supervisor app state GET v3 tests

thgreasi:
Change-type: patch

### DIFF
--- a/test/19_apps.ts
+++ b/test/19_apps.ts
@@ -4,14 +4,17 @@ import * as fixtures from './test-lib/fixtures';
 import * as fakeDevice from './test-lib/fake-device';
 
 import * as configMock from '../src/lib/config';
-import { UserObjectParam } from './test-lib/supertest';
+import type { UserObjectParam } from './test-lib/supertest';
+import type { Application, Release } from '../src/balena-model';
+import { expectResourceToMatch } from './test-lib/api-helpers';
 
 describe('Apps', function () {
 	describe('Supervisor app', () => {
 		let fx: fixtures.Fixtures;
 		let admin: UserObjectParam;
-		let applicationId: number;
-		let supervisorAppUuid: string;
+		let userApp: Application;
+		let supervisorApp: Application;
+		let supervisorRelease2: Release;
 		let deviceWithSupervisor: fakeDevice.Device;
 		let deviceWithoutSupervisor: fakeDevice.Device;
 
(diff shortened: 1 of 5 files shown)


test/19_apps.ts
```diff
@@ -20,23 +23,31 @@ describe('Apps', function () {
...
+			supervisorApp = fx.applications.supervisorApp;
+			supervisorRelease2 = fx.releases.supervisorRelease2;
 
 			deviceWithSupervisor = await fakeDevice.provisionDevice(
 				admin,
-				applicationId,
+				userApp.id,
 				'balenaOS 2.3.0+rev1',
 				'1.0.1',
-				fx.releases.supervisorRelease1.id,
```
thgreasi, line 32: Ends up that all that this was not respected and the API was overwriting it w/ the sv release that it found based on `1.0.1`. On top of this `provisionDevice` should probably be changed to be using `register/device` rather than orchestrating OData requests, so that it's more realistic, and in that case providing the target sv release will not be a case anyway.

test/test-lib/api-helpers.ts
```diff
@@ -81,33 +82,45 @@ export const addImageToRelease = async (
...
+
+	const pineClientRequest = pineUser.get({
+		resource,
+		id,
+		options: {
+			$select: Object.keys(expectations),
+		},
+	});
+	const result = (
+		'expect' in pineClientRequest
```
fisehara, line 110: Uncommon, but do I understand correctly that the pineClientRequest has a key `expect` ?

test/test-lib/api-helpers.ts
```diff
@@ -81,33 +82,45 @@ export const addImageToRelease = async (
...
+			? (await (pineClientRequest as ReturnType<PineTest['get']>).expect(200))
+					.body
+			: await pineClientRequest
+	) as T | undefined;
+	expect(result).to.be.an('object');
+	Object.entries(expectations).forEach(([key, valueOrAssetion]) => {
```
fisehara, line 116: Why is that changed from the `for ([] of ) {}` ?

test/test-lib/api-helpers.ts
```diff
@@ -81,33 +82,45 @@ export const addImageToRelease = async (
...
+		if (typeof valueOrAssetion === 'function') {
+			valueOrAssetion(expect(result).to.have.property(key));
+		} else if (typeof valueOrAssetion === 'object' && valueOrAssetion != null) {
+			expect(result).to.have.property(key).to.deep.equal(valueOrAssetion);
 		} else {
-			expect(result).to.have.property(key, valueOrAssertion);
+			expect(result).to.have.property(key, valueOrAssetion);
 		}
```
fisehara, line 123:
```suggestion
		if (typeof valueOrAssertion === 'function') {
			valueOrAssertion(expect(result).to.have.property(key));
		} else if (typeof valueOrAssertion === 'object' && valueOrAssertion != null) {
			expect(result).to.have.property(key).to.deep.equal(valueOrAssertion);
		} else {
			expect(result).to.have.property(key, valueOrAssertion);
		}
```

test/test-lib/api-helpers.ts
thgreasi: So at this point pineUser is either a pinejs-client-supertest or a pinejs-client-request (the one the pinejs exposes) instance, and the badly named `pineClientRequest` variable here is a request object (promise). So if it's from `pinejs-client-supertest`, then it will also have a `.expect()` (since it's a supertest + pinejs-client hybrid). So here we are doing a feature check (rather than instance check) and if `.expect` is available, we use it. (let me rename the var...)

test/test-lib/api-helpers.ts
thgreasi: copied over from bC-api, will make it a for-of again :+1:

fisehara (approved):

test/test-lib/api-helpers.ts
```diff
@@ -1,8 +1,9 @@
+import type { PinejsClient } from '@balena/pinejs/out/sbvr-api/sbvr-utils';
```
Page-, line 1: @thgreasi can you remove the deep imports like this please

thgreasi: OOOoppps, sorry for that.
See: https://github.com/balena-io/open-balena-api/pull/1500

PR #1523, opened 2024-01-19
Add onTokenRefresh callback

otaviojacobi:
Change-type: minor

### DIFF
--- a/src/features/auth/refresh-token.ts
+++ b/src/features/auth/refresh-token.ts
@@ -4,21 +4,33 @@ import { updateUserXHR } from '../../infra/auth/jwt';
 import { captureException } from '../../infra/error-handling';
 import type { RequestHandler } from 'express';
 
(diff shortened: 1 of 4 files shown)


src/features/auth/refresh-token.ts
```diff
@@ -1,25 +1,42 @@
...
 import { captureException } from '../../infra/error-handling';
 import type { RequestHandler } from 'express';
 
-const { BadRequestError } = errors;
+import type { SetupOptions } from '../..';
+
+const { BadRequestError, NotFoundError } = errors;
+
+let onTokenRefresh: SetupOptions['onTokenRefresh'];
+export const setOnTokenRefresh = (
```
otaviojacobi, line 12: This is done as a separate function because following the same pattern as "onLogin" would require a breaking change on `refreshToken` function which is exported and used for `/whoami`

PR #1531, opened 2024-01-25
Fix some typing issues that are normally hidden by `skipLibCheck`

Page-:
Change-type: patch

### DIFF
--- a/typings/balena-device-config.d.ts
+++ b/typings/balena-device-config.d.ts
@@ -27,7 +27,6 @@ declare module 'balena-device-config' {
 
 	export interface GenerateParams {
 		appUpdatePollInterval?: number;
-		network?: any[];
 		ip?: string;
 		gateway?: string;
 		netmask?: string;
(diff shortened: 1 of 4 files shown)


Page-:
Ideally we could disable `skipLibCheck` altogether but there are typings in node_modules that would break the build if we did.. I am however working on ensuring all the typings we control will still be checked and will not be affected by the `skipLibCheck` directive and this is a precursor to that (since they need to be fully valid for that)

typings/balena-device-config.d.ts
```diff
@@ -27,15 +27,14 @@ declare module 'balena-device-config' {
...
-		network?: any[];
 		ip?: string;
 		gateway?: string;
 		netmask?: string;
 		wifiSsid?: string;
 		wifiKey?: string;
 		connectivity?: string;
 		network?:
-			| wifi
+			| 'wifi'
```
thgreasi, line 37: :facepalm:

thgreasi (approved):

PR #1537, opened 2024-02-05
Switch to ES modules

Page-:
Change-type: major


index.js
```diff
@@ -1,10 +1,18 @@
...
+const start = async () => {
+	// Set the desired es version for downstream modules that support it
+	(await import('@balena/es-version')).set('es2022');
+
+	await import('./init.js');
+};
+
+const numWorkers =
+	parseInt(process.env.NUM_WORKERS ?? '0', 10) ||
+	(await import('os')).cpus().length;
```
thgreasi, line 10: Q: I guess it wouldn't be that bad if we did a normal import for 'os' right, since it's a node thing and I guess it would be fast?

Page-: I would expect so but also wanted to go for minimal changes

thgreasi:
`ReferenceError: __dirname is not defined in ES module scope at file:///usr/src/app/src/balena.ts:7:31` :facepalm:

thgreasi (approved):

src/balena.ts
```diff
@@ -1,19 +1,26 @@
...
 
-import * as userHasDirectAccessToApplication from './features/applications/models/user__has_direct_access_to__application';
-import * as deviceAdditions from './features/devices/models/device-additions';
-import * as releaseAdditions from './features/ci-cd/models/release-additions';
+import * as userHasDirectAccessToApplication from './features/applications/models/user__has_direct_access_to__application.js';
+import * as deviceAdditions from './features/devices/models/device-additions.js';
+import * as releaseAdditions from './features/ci-cd/models/release-additions.js';
 
-export const apiRoot = 'resin';
+export const apiRoot = 'resin.js';
```
thgreasi, line 13: Isn't this going to change the version in the OData `/resin/<resource>` endpoints as well?

src/lib/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { pseudoRandomBytes } from 'crypto';
```
Page-, line 1: `pseudoRandomBytes` has been deprecated for a long time since it was made to match `randomBytes` and is not available when importing from ESM

src/infra/auth/jwt-passport.ts
```diff
@@ -8,14 +8,14 @@ import { TypedError } from 'typed-error';
...
-import type { ScopedAccessToken, ScopedToken } from './jwt';
-import type { PickDeferred, User as DbUser } from '../../balena-model';
-import { getGuestActorId } from './permissions';
+import { captureException } from '../error-handling/index.js';
+import type { ScopedAccessToken, ScopedToken } from './jwt.js';
+import type { PickDeferred, User as DbUser } from '../../balena-model.js';
+import { getGuestActorId } from './permissions.js';
 
-export { SignOptions } from 'jsonwebtoken';
+export type { SignOptions } from 'jsonwebtoken';
```
Page-, line 18: This is necessary so that it is omitted from the generated js

PR #1538, opened 2024-02-05
Return empty 404 instead of expressjs default html

fisehara:
Change-type: patch

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -484,6 +484,11 @@ async function startServer(
 	port: string | number,
 ): Promise<Server> {
 	let server: Server;
+	// empty 404 error for undefined paths to avoid
+	// express creates default 404 error with html body
+	app.use((_request, response) => {
+		response.status(404).end();
+	});
 	await Bluebird.fromCallback((cb) => {
 		server = app.listen(port, cb as (...args: any[]) => void);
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -484,6 +484,11 @@ async function startServer(
 	port: string | number,
 ): Promise<Server> {
 	let server: Server;
+	// empty 404 error for undefined paths
+	app.use((_request, response) => {
+		response.status(404);
+		response.end();
```
thgreasi, line 490: NIT: We can chain them for consistency :)
```suggestion
		res.status(404).end();
```

src/index.ts
```diff
@@ -484,6 +484,11 @@ async function startServer(
 	port: string | number,
 ): Promise<Server> {
 	let server: Server;
+	// empty 404 error for undefined paths
+	app.use((_request, response) => {
```
thgreasi, line 488: Just asking: Given that we already support tons of options in the setup call (eg onInit, onInitRoutes, ...) and ppl should be able to already register everything using those already, would it make more sense if we did this as he last step of `setup()`, rather than in the `startServer()`?

thgreasi (approved):

PR #1540, opened 2024-02-07
package.json: Add cache & config as exports/package entry points

thgreasi:
That's to allow consumers to access them
before the pine gets imported/initialized.

Change-type: minor
See: https://nodejs.org/api/packages.html#package-entry-points


package.json
```diff
@@ -9,6 +9,12 @@
   },
   "type": "commonjs",
   "main": "dist/index",
+  "exports": {
+    ".": "./dist/index.js",
+    "./cache": "./dist/infra/cache/index.js",
+    "./config": "./dist/lib/config.js",
+    "./dist/*": "./dist/*"
```
thgreasi, line 16: That's necessary so that deep imports from dist still work like before (eg for importing balena.sbvr or balena-init.sql or mocking the contracts in tests).

PR #1541, opened 2024-02-08
Switch test files to exporting an init function for ESM compatibility

Page-:
Change-type: patch

### DIFF
--- a/test/00_init.ts
+++ b/test/00_init.ts
@@ -25,7 +25,7 @@ const prefixes: Dictionary<true> = {};
 
 Bluebird.resolve(fs.promises.readdir(__dirname))
 	.call('sort')
-	.each((fileName) => {
+	.each(async (fileName) => {
 		const ext = path.extname(fileName);
 		if (ext !== '.ts') {
 			return;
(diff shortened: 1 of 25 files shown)


thgreasi:
> Cannot find module '/usr/src/app/test/00_init' imported from /usr/src/app/test/00_init.ts

:(

test/00_init.ts
```diff
@@ -42,27 +42,29 @@ Bluebird.resolve(fs.promises.readdir(__dirname))
 			throw new Error(`Prefix ${prefix} has already been used`);
 		}
 		prefixes[prefix] = true;
```
thgreasi, line 44: Do we need this v like we did elsewhere?
```suggestion
		prefixes[prefix] = true;
		if (prefix === '00') {
			// Don't double load this file
			return;
		}
```

test/00_init.ts
```diff
@@ -25,7 +25,7 @@ const prefixes: Dictionary<true> = {};
 
 Bluebird.resolve(fs.promises.readdir(__dirname))
 	.call('sort')
-	.each((fileName) => {
+	.each(async (fileName) => {
 		const ext = path.extname(fileName);
 		if (ext !== '.ts') {
```
thgreasi, line 30: Should this now become v ?
```suggestion
		if (ext !== '.ts' && ext !== '.js') {
```

test/00_init.ts
Page-: Yeah, just because it's not expected to be a test file really so doesn't export (or need to) an init function

test/00_init.ts
Page-: There's no need for it, it'd only be if we started writing tests in js directly or if we pre-compile the test files. Either way neither of those are part of this PR

thgreasi (approved):

PR #1561, opened 2024-02-27
Add test cases with race conditions on the heartbeat state management

thgreasi:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/Statuspage.20Alert.20-.20lj8x31dm9wh9

### DIFF
--- a/test/03_device-state.ts
+++ b/test/03_device-state.ts
@@ -9,7 +9,7 @@ import { supertest } from './test-lib/supertest.js';
 import * as versions from './test-lib/versions.js';
 import * as config from '../src/lib/config.js';
 import * as stateMock from '../src/features/device-heartbeat/index.js';
-import { waitFor } from './test-lib/common.js';
+import { itExpectsError, waitFor } from './test-lib/common.js';
 import * as fixtures from './test-lib/fixtures.js';
 import { expectResourceToMatch } from './test-lib/api-helpers.js';
 import { redis, redisRO } from '../src/infra/redis/index.js';
@@ -35,8 +35,29 @@ class StateTracker {
 	};
 }
 
+const getHeartbeatWriteCacheKey = (deviceId: number) =>
+	`device-online-state:${deviceId}`;
+
+const getHeartbeatWriteCacheState = async (deviceId: number) => {
+	const value = await redisRO.get(getHeartbeatWriteCacheKey(deviceId));
+	if (typeof value !== 'string') {
+		return null;
+	}
+	return JSON.parse(value);
+};
+
 config.TEST_MOCK_ONLY.DEFAULT_SUPERVISOR_POLL_INTERVAL = POLL_MSEC;
 config.TEST_MOCK_ONLY.API_HEARTBEAT_STATE_TIMEOUT_SECONDS = TIMEOUT_SEC;
+
+const devicePollInterval =
+	Math.ceil((POLL_MSEC * stateMock.POLL_JITTER_FACTOR) / 1000) * 1000;
+
+/**
+ * The 'get-state' event has to be consumed in a short period of time.
+ * If we wait longer, and that's comparable to the POLL_MSEC or TIMEOUT_SEC,
+ * we then could end up capturing different heartbeat state changes
+ */
+const maxGetStateEventConsumptionTimeout = 100;
 
 export default () => {
 	versions.test((version, pineTest) => {
@@ -51,6 +72,34 @@ export default () => {
 				/** Tracks updateDeviceModel() calls */
 				const tracker = new StateTracker();
 				const updateDeviceModel = stateMock.getInstance()['updateDeviceModel'];
+
+				const expectDeviceHeartbeat = async (
+					deviceId: number,
+					params:
+						| string
+						| {
+								db?: string;
+								cache?: string | null;
+						  },
+				) => {
+					const db = typeof params === 'string' ? params : params.db;
+					const cache = typeof params === 'string' ? params : params.cache;
+
+					if (db != null) {
+						await expectResourceToMatch(pineUser, 'device', deviceId, {
+							api_heartbeat_state: db,
+						});
+					}
+					if (cache !== undefined) {
+						const actualCacheValue =
+							await getHeartbeatWriteCacheState(deviceId);
+						if (cache == null) {
+							expect(actualCacheValue).to.be.null;
+						} else {
+							expect(actualCacheValue).to.have.property('currentState', cache);
+						}
+					}
+				};
 
 				before(async () => {
 					fx = await fixtures.load('03-device-state');
@@ -170,10 +219,6 @@ export default () => {
 					});
 
 					describe('Event Tracking', () => {
-						const devicePollInterval =
-							Math.ceil((POLL_MSEC * stateMock.POLL_JITTER_FACTOR) / 1000) *
-							1000;
-
 						let deviceUserRequestedState: fakeDevice.Device;
 
 						const stateChangeEventSpy = sinon.spy();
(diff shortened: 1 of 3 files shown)


test/test-lib/common.ts
```diff
@@ -36,3 +37,35 @@ export async function waitFor({
 
 	throw new TimedOutError();
 }
+
+export const itExpectsError = (
+	title: string,
+	fn: Mocha.AsyncFunc,
+	isExpectedError?: string | RegExp | ((err: Error) => boolean),
```
joshbwlng, line 44:
```suggestion
	expectedError?: string | RegExp | ((err: Error) => boolean),
```
`isExpectedError` makes sense for the function, but not for the others. I'd call is `expectedError` and update the function case to return the expected error string to make it better match the other options.

test/03_device-state.ts
```diff
@@ -510,6 +553,314 @@ export default () => {
...
+									device3ChangeEventSpy.resetHistory();
+									await expectDeviceHeartbeat(device3.id, {
+										db: DeviceOnlineStates.Online,
+										cache: DeviceOnlineStates.Online,
+									});
+
+									await setTimeout(heartbeatTimeoutChangeInterval);
+									await waitFor({
+										checkFn: () => device3ChangeEventSpy.called,
+									});
```
joshbwlng, line 588: Do we need the `await setTimeout()` if we're executing `await waitFor()` directly after?

test/03_device-state.ts
```diff
@@ -510,6 +553,314 @@ export default () => {
...
+									'should turn Online and not consume the stale heartbeat redis queue message when a newer state GET has arrived',
+									async function () {
+										await fakeDevice.getState(
+											device3,
+											device3.uuid,
+											stateVersion,
+										);
+										await setTimeout(100);
+										// The write cache has expired so the DB's heartbeat gets updated
+										expect(device3ChangeEventSpy.called).to.be.true;
```
joshbwlng, line 615:
```suggestion
										// The write cache has expired so the DB's heartbeat gets updated
										await waitFor({
											checkFn: () => device3ChangeEventSpy.called,
										});
```

test/test-lib/common.ts
thgreasi: If I get this right you suggest to do `(err: Error) => string`.
How will then the returned error message be used by the helper?
That parameter is used by the code as a filter on whether the correct error was thrown, and the string & regex alternatives are there just for convenience.
If anything, I would expect us to actually convert the string & regex to a matcher function (in the helper's body), so that we can have a single v check.
```
if (!isExpectedError(err)) {
	throw err;
}
```

test/03_device-state.ts
thgreasi: What you say is correct. It seems that the mock device poll interval (thus how much time is needed to switch to Timeout) is small enough that `waitFor` is enough.
If we later increase the mock poll interval to a value that above the max default wait time that `waitFor` has, we then would need the setTimeout as well.
I will go with `waitFor` & leave a comment.

test/03_device-state.ts
thgreasi: What the `await setTimeout(100);`s all over the test cases try to do, is to wait just enough time for the `'get-state'` event to be fired and processed by the API. The default max wait of `waitFor` is bigger than the mock device poll interval * jitter. As a result while in most test cases we want to wait for `it should shortly switch from timeout to online after a state get`, going for a plain waitFor would make it incorrectly accept a switch from timeout to offline as well.

I think I need to make a helper function w/ that setTimeout explains better why we need to do this...

test/test-lib/common.ts
joshbwlng: I was thinking a function that returned a string that you then compared to the actual error message string in the helper. But converting the string and regex to a matcher function like you mentioned would be fine as well I think, as long as each option follows the same general usage pattern.

test/03_device-state.ts
```diff
@@ -510,6 +555,316 @@ export default () => {
...
+										expect(
+											await getHeartbeatWriteCacheState(device3.id),
+										).to.have.property('currentState', 'online');
+									},
+									/The api_heartbeat_state should have recovered to online but it was found offline/,
+								);
+
+								// TODO: This should not be happening. Drop this v test once the previous tests gets fixed.
+								itExpectsError(
+									'[FIXME] is expected that it will stay stuck to Offline when doing regular state GETs',
```
thgreasi, line 697: This whole test case gets removed in #1560 since the device will now be switching to online on the first state GET, so if you see anything ugly, be sure that it's going away right after this gets merged.

test/03_device-state.ts
```diff
@@ -510,6 +555,316 @@ export default () => {
 									api_heartbeat_state: DeviceOnlineStates.Offline,
 								});
 							});
+
+							describe('when the write cache expires while the device is on Timeout (eg API stops serving requests)', function () {
+								let device3: fakeDevice.Device;
+								const device3ChangeEventSpy = sinon.spy();
+
+								before(async function () {
+									device3 = await fakeDevice.provisionDevice(
```
joshbwlng, line 564:
```suggestion
									// Provision a device and wait for its heartbeat to be Online
									device3 = await fakeDevice.provisionDevice(
```

test/03_device-state.ts
```diff
@@ -510,6 +555,316 @@ export default () => {
...
+											},
+										);
+									},
+									/The api_heartbeat_state should have become online but it was found offline/,
+								);
+
+								itExpectsError(
+									'should switch to Online after a subsequent state GET and not stay stuck as offline forever',
+									async function () {
+										// Confirm that we are int he case that the DB says Offline & the write Cache says Online
```
joshbwlng, line 656:
```suggestion
										// Confirm that we are in the case that the DB says Offline & the write Cache says Online
```

test/03_device-state.ts
```diff
@@ -510,6 +555,316 @@ export default () => {
...
+									/The api_heartbeat_state should have recovered to online but it was found offline/,
+								);
+							});
+
+							describe('when the write cache expires while the device is Online (eg API stops serving requests)', function () {
+								let device3: fakeDevice.Device;
+								const device3ChangeEventSpy = sinon.spy();
+
+								before(async function () {
+									device3 = await fakeDevice.provisionDevice(
```
joshbwlng, line 752: You also provision a device and wait for it to be Online in the above before. May be worth reducing code duplication and creating a shared function to do this.

joshbwlng: Actually it looks like two test cases are also quite similar. I first though we might could tidy things up by looping through `['Timeout', 'Online']` right above the describe level, but I guess the internal test logic is different enough that it wouldn't help.

test/03_device-state.ts
```diff
@@ -510,6 +555,316 @@ export default () => {
...
+
+								beforeEach(function () {
+									device3ChangeEventSpy.resetHistory();
+									delete tracker.states[device3.id];
+								});
+
+								itExpectsError(
+									'should stay Online and not consume the stale heartbeat redis queue message when a newer state GET has arrived',
+									async function () {
+										const initialWait = 0.8 * devicePollInterval;
```
joshbwlng, line 787: Such accurate wait requirements makes me worry about test flakiness. Is that an invalid concern?

test/test-lib/common.ts
```diff
@@ -6,33 +8,86 @@ type PredicateFunction = () => Resolvable<boolean>;
...
  * predicate will cease the looping. If `FALSE` then we delay by `delayMs` milliseconds.
  *
  * If we haven't seen a `TRUE` when the loop finishes, we throw a {@link TimedOutError}
- *
- * @export
- * @param {number} delayMs
- * @param {number} maxCount
- * @param {PredicateFunction} checkFn
  */
 export async function waitFor({
```
joshbwlng, line 16: I came up with a possible alternative so we can avoid changing current default values. What do you think?
```typescript
export async function waitFor({
    delayMs = 100,
    maxCount = 100,
    maxWait,
    checkFn,
}: {
    delayMs?: number;
    maxCount?: number;
    maxWait?: number;
    checkFn: PredicateFunction;
}) {
    // If maxWait is provided, override delayMs and maxCount values
    // and throw an error when maxWait is exceeded.
    let timeout: NodeJS.Timeout | null = null;
    if (maxWait != null) {
        delayMs = Math.max(maxWait / 10, 10);
        maxCount = Math.ceil(maxWait / delayMs);

        // Set timeout for an error throw in case checkFn() takes too long.
        timeout = setTimeout(() => {
            throw new TimedOutError();
        }, maxWait);
    }

    try {
        for (let i = 1; i <= maxCount; i++) {
            console.log(`⌚  Waiting (${i}/${maxCount})...`);
            await delay(delayMs);

            if (await checkFn()) {
                return;
            }
        }

        throw new TimedOutError();
    } finally {
        if (timeout != null) {
            clearTimeout(timeout);
        }
    }
}
```

test/03_device-state.ts
thgreasi: That should be clear from the describe "when the write cache expires while the device is on Timeout" but fine.

test/03_device-state.ts
thgreasi: Yea, given the extra orchestration of setting up the spies & the stateMock, I think it's better to keep them separate, so that they also closely match the title in the describe, or it's more obvious if they do not match...

test/03_device-state.ts
thgreasi: It shouldn't, except if we make the mock poll interval even smaller.
Let me leave a comment on how the number was picked.

thgreasi: I also added bigger descriptions of what the two describes are trying to test:

// Here we test how the heartbeat behaves when the write cache has expired (eg b/c of downtime) and when the API starts working again,
// the first state GET request that we receive for a device happens while its heartbeat has switched to Timeout
// (the stale redis queued message to switch to Timeout has been consumed, and one to go Offline has been scheduled).

// Here we test how the heartbeat behaves when the write cache has expired (eg b/c of downtime) and when the API starts working again,
// the first state GET request that we receive for a device happens while its heartbeat is still Online
// (the stale redis queued message to switch to Timeout has not yet be consumed).

test/test-lib/common.ts
thgreasi: The throw in that setTimeout will not propagate to the promise change (and probably crash the test runner as unhandled exception). I'm afraid that the `Promise.race` is the only way to achieve exactly this, so let me try to do that...

joshbwlng: I see what you mean. I came up with another way to write the race to make it a bit more clear (at least to me), but I'll leave it up to you:
```typescript
	// Primary wait loop
	const waitLoop = (async () => {
		for (let i = 1; i <= maxCount; i++) {
			console.log(`⌚  Waiting ${delayMs}ms (${i}/${maxCount})...`);
			await setTimeout(delayMs);

			if (await checkFn()) {
				return;
			}
		}

		throw new TimedOutError();
	})();

	// Timeout logic when maxWait is provided
	const timeoutFn = maxWait == null ? undefined : (async () => {
		await setTimeout(maxWait);
		throw new TimedOutError(`Exceeded maxWait ${maxWait}`);
	})();

	// Await the race between the primary wait loop and the timeoutFn
	await Promise.race([waitLoop, timeoutFn].filter(Boolean));
```

PR #1570, opened 2024-03-05
push optional OpenVPN UDP host to clients

ab77:
Requires:
* https://github.com/balena-io/open-balena-vpn/pull/406


src/lib/config.ts
```diff
@@ -39,7 +39,6 @@ down /etc/openvpn-misc/downscript.sh
 comp-lzo
 dev resin-vpn
 dev-type tun
-proto tcp
```
thgreasi, line 42: We don't need this at all anymore?

src/lib/config.ts
```diff
@@ -132,10 +131,20 @@ export const DB_QUERY_TIMEOUT = intVar(
...
 export const DEVICE_CONFIG_OPENVPN_CA = requiredVar('DEVICE_CONFIG_OPENVPN_CA');
+export const VPN_UDP_HOST = optionalVar('VPN_UDP_HOST');
+export const VPN_HOST = requiredVar('VPN_HOST');
+export const VPN_PORT = requiredVar('VPN_PORT');
+
+const remotes = [];
+if (VPN_UDP_HOST) {
+	remotes.push(`remote ${VPN_UDP_HOST} ${VPN_PORT} udp`);
+}
+remotes.push(`remote ${VPN_HOST} ${VPN_PORT} tcp`);
```
thgreasi, line 142: I saw in the docs at some point that
> If you want OpenVPN to listen on both a UDP and TCP port, you must run two separate OpenVPN instances

See: https://openvpn.net/community-resources/how-to/#:~:text=If%20you%20want%20OpenVPN%20to%20listen%20on%20both%20a%20UDP%20and%20TCP%20port%2C%20you%20must%20run%20two%20separate%20OpenVPN%20instances

So should this be an `if-else`?

src/lib/config.ts
ab77: @thgreasi it's handled by the default `remote` directive [here](https://github.com/balena-io/open-balena-api/pull/1570/files#diff-0b4f5977a52cef50ed20e7cdf9761795df9ee1cb3b5e5ea45edeb9394e8a47ffR142) now..

config/confd/conf.d/env.toml
```diff
@@ -50,18 +49,20 @@ keys = [
...
-	"WEBRESOURCES_S3_REGION",
-	"WEBRESOURCES_S3_HOST",
 	"WEBRESOURCES_S3_BUCKET",
+	"WEBRESOURCES_S3_HOST",
 	"WEBRESOURCES_S3_MAX_FILESIZE",
-	"WEBRESOURCES_CLOUDFRONT_PUBLICKEY",
-	"WEBRESOURCES_CLOUDFRONT_HOST",
-	"WEBRESOURCES_CLOUDFRONT_PRIVATEKEY_PATH"
+	"WEBRESOURCES_S3_REGION",
+	"WEBRESOURCES_S3_SECRET_KEY",
```
klutchell, line 67: Can you do this reordering in a separate commit? It's nice to have it done, but it makes it much harder to review.

Page-: They should actually just be removed altogether as they have no purpose when using an env var backend

ab77: Done, deleted.

src/lib/config.ts
ab77: @Page- @thgreasi need a decision here folks. Do we add extra code and not modify the config unless a new feature switch is set or go as-is.

thgreasi: @ab77 I would be happy with something like:
````
const DEVICE_CONFIG_OPENVPN_CONFIG = `
client
${
  VPN_UDP_HOST == null
    ? `remote ${VPN_HOST} ${VPN_PORT}`
    : `remote ${VPN_UDP_HOST} ${VPN_PORT} udp
remote ${VPN_HOST} ${VPN_PORT} tcp`
}
remote {{VPN_DETAILS}}
resolv-retry infinite
remote-cert-tls server
tls-version-min 1.2
ca /etc/openvpn/ca.crt
auth-user-pass /var/volatile/vpn-auth
auth-retry none
script-security 2
up /etc/openvpn-misc/upscript.sh
up-restart
down /etc/openvpn-misc/downscript.sh
comp-lzo
dev resin-vpn
dev-type tun
${VPN_UDP_HOST == null ? 'proto tcp' : ''}
nobind

persist-key
persist-tun
verb 3
user openvpn
group openvpn

reneg-bytes 0
reneg-pkts 0
reneg-sec 0
`;
```
That should give exactly the same config and we can (I can help) also add a test case for that.
PS: Not sure whether the extra space before the `nobind` would cause any issues when VPN_UDP_HOST is defined, but even in that case we could just add a `.replace('\n\nnobind', '\nnobind')`.

src/lib/config.ts
```diff
@@ -137,10 +137,28 @@ export const DB_QUERY_TIMEOUT = intVar(
...
+		.split('\n')
+		.filter((el) => el !== 'proto tcp')
+		.join('\n');
+} else {
+	tmpConf = openVpnConfig.replace(
+		'{{VPN_DETAILS}}',
+		`remote ${VPN_HOST} ${VPN_PORT}`,
+	);
+}
+export const DEVICE_CONFIG_OPENVPN_CONFIG = tmpConf;
```
thgreasi, line 160: Tiny-NIT: This should be equivalent:
```suggestion
export const DEVICE_CONFIG_OPENVPN_CONFIG = (() => {
	if (VPN_UDP_HOST) {
		const remotes = [
			`remote ${VPN_UDP_HOST} ${VPN_PORT} udp`,
			`remote ${VPN_HOST} ${VPN_PORT} tcp`,
		].join('\n');
		return openVpnConfig
			.replace('{{VPN_DETAILS}}', remotes)
			.replace('\nproto tcp\n', '\n')

	} else {
		return openVpnConfig.replace(
			'{{VPN_DETAILS}}',
			`remote ${VPN_HOST} ${VPN_PORT}`,
		);
	}
})();
```

config/confd/templates/env.tmpl
```diff
@@ -75,6 +75,7 @@ TOKEN_AUTH_JWT_ALGO='{{getenv "TOKEN_AUTH_JWT_ALGO"}}'
 TRANSACTION_TIMEOUT_MS=30000
 UV_THREADPOOL_SIZE=128
 VPN_HOST={{getenv "VPN_HOST"}}
+{{if getenv "VPN_UDP_HOST"}}VPN_UDP_HOST="{{getenv "VPN_UDP_HOST"}"}"{{end}}
```
thgreasi, line 78: I think there is a typo in `"}"}"` and I don't think we need the quotes around the `"{{getenv ... }}".
```suggestion
{{if getenv "VPN_UDP_HOST"}}VPN_UDP_HOST={{getenv "VPN_UDP_HOST"}}{{end}}
```

Moreover b/c of how `optionalVar` already handles empty strings as nulls, I think this could even (optionally) become:
```suggestion
VPN_UDP_HOST={{getenv "VPN_UDP_HOST"}}
```

thgreasi (commented):
Other than those two NITs lgtm :+1:

src/lib/config.ts
ab77: Please commit your suggestion @thgreasi, or leave my version.

thgreasi (approved):

PR #1573, opened 2024-03-08
Use the release semver fields along with release_tags in hostApp operations

thgreasi:
Change-type: minor
See: https://balena.fibery.io/Work/Project/Remove-hostApp-version-release-tag-dependencies-226
See: https://balena.fibery.io/Work/Improvement/Start-using-the-release-semver-fields-for-hostapp-validation-in-the-API-along-with-release_tags-1447
See: https://balena.fibery.io/Work/Task/Use-the-release-semver-fields-along-with-release_tags-in-hostApp-operations-1840

### DIFF
--- a/src/features/hostapp/hooks/target-hostapp.ts
+++ b/src/features/hostapp/hooks/target-hostapp.ts
@@ -5,12 +5,15 @@ import {
 	errors as pinejsErrors,
 } from '@balena/pinejs';
 import * as semver from 'balena-semver';
+import type { SemVer } from 'semver';
 import type {
 	Device,
 	ReleaseTag,
 	PickDeferred,
+	Release,
 } from '../../../balena-model.js';
 import { groupByMap } from '../../../lib/utils.js';
+import { ThisShouldNeverHappenError } from '../../../infra/error-handling/index.js';
 const { BadRequestError } = pinejsErrors;
 
 hooks.addPureHook('PATCH', 'resin', 'device', {
@@ -90,7 +93,7 @@ hooks.addPureHook('POST', 'resin', 'device', {
 			request.values.os_variant != null &&
 			request.values.is_of__device_type != null
 		) {
-			const [hostappRelease] = await getOSReleaseResource(
+			const hostappRelease = await getOSReleaseResource(
 				api,
 				request.values.os_version,
 				request.values.os_variant,
@@ -150,7 +153,7 @@ async function setOSReleaseResource(
 			async ([deviceTypeId, affectedDevices]) => {
 				const affectedDeviceIds = affectedDevices.map((d) => d.id);
 
-				const [osRelease] = await getOSReleaseResource(
+				const osRelease = await getOSReleaseResource(
 					api,
 					osVersion,
 					osVariant,
@@ -183,13 +186,19 @@ async function getOSReleaseResource(
 	osVariant: string,
 	deviceTypeId: number,
 ) {
-	return await api.get({
+	const parsedOsVersion = semver.parse(osVersion);
+	// balena-semver is able to parse all OS versions that we support,
+	// so if it can't parse the given version string, then we can be sure
+	// that there can't be a hostApp release matching it.
+	if (parsedOsVersion == null) {
+		return;
+	}
+	const releases = (await api.get({
 		resource: 'release',
 		options: {
-			$select: ['id', 'belongs_to__application'],
+			$top: 2,
+			$select: 'id',
 			$filter: {
-				// TODO: better to (eventually) use release version, once that's fully supported
-				// https://www.flowdock.com/app/rulemotion/resin-tech/threads/Ao4qzbDh8Z4Pgq_xF99Ld5fME35
 				status: 'success',
 				belongs_to__application: {
 					$any: {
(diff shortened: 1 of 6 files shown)


src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -202,58 +211,112 @@ async function getOSReleaseResource(
...
-								$expr: {
-									rt: {
-										value: normalizeOsVersion(osVersion),
-										tag_key: 'version',
-									},
-								},
-							},
-						},
+						// This effectively normalizes the versions string and
+						// compares it with the `semver`` computed term, but we are
```
joshbwlng, line 217:
```suggestion
						// compares it with the `semver` computed term, but we are
```

src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -202,58 +211,112 @@ async function getOSReleaseResource(
...
+								],
 							},
 						],
 					},
 				],
 			},
+			$orderby: {
+				// We order the resuts by semver_major DESC so that we always prefer rows
+				// that were matched via the semver fields rather than tags in case of a conflict,
+				// (since versions using only tags would have a 0.0.0 semver_.
```
joshbwlng, line 291:
```suggestion
				// (since versions using only tags would have a 0.0.0 semver).
```

src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -315,51 +402,64 @@ async function checkHostappReleaseUpgrades(
...
+			release_tag: [Pick<ReleaseTag, 'value'>?];
+		}
+	>;
 
-	for (const release of releases) {
-		const oldVersion = release.value;
-		if (semver.lt(newHostappVersion, oldVersion)) {
+	for (const oldOsRelease of oldOsReleases) {
+		const oldVersion = getBaseVersionFromReleaseSemverOrTag(oldOsRelease);
+		// Let the device upgrade if it is operatred by a release w/ 0.0.0 semver,
```
joshbwlng, line 460:
```suggestion
		// Let the device upgrade if it is operated by a release w/ 0.0.0 semver,
```

src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -202,58 +211,112 @@ async function getOSReleaseResource(
...
+	}
+
+	const [release] = releases;
+	return release;
 }
 
-function normalizeVariant(variant: string) {
+function getRevisionFromSemver(parsedOsVersion: SemVer) {
+	for (const buildPart of parsedOsVersion.build) {
+		const match = buildPart.match(/^rev(\d+)$/)?.[1];
```
joshbwlng, line 311:
```suggestion
	const pattern = /^rev(\d+)$/;
	for (const buildPart of parsedOsVersion.build) {
		const match = buildPart.match(pattern)?.[1];
```
Define the RegExp outside of the loop for better performance. May even be worth defining it outside of the function itself to completely reduce the number of times it ends up being compiled.

src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -202,58 +211,112 @@ async function getOSReleaseResource(
...
+				// that were matched via the semver fields rather than tags in case of a conflict,
+				// (since versions using only tags would have a 0.0.0 semver_.
+				semver_major: 'desc',
+			},
 		},
-	});
+	})) as [
+		PickDeferred<Release, 'id' | 'belongs_to__application'>?,
+		PickDeferred<Release, 'id' | 'belongs_to__application'>?,
+	];
```
joshbwlng, line 298: What's the benefit of doing this vs an `Array<...>` instead? Just denoting that it should ever only be 2 entries as most?

thgreasi: My primary goal was to have the typings properly reflect that the first item of the array might be undefined.
I tried `Array<PickDeferred<Release, 'id' | 'belongs_to__application'> | undefined>` but it looked weird. Tbh this could also just `[PickDeferred<Release, 'id' | 'belongs_to__application'>?]` and we would just be checking for the length in exactly the same way (which will work until TS starts inferring the length as well in the future).
I guess I can do the last one, but probably the correct answer would be using TS's `noUncheckedIndexedAccess` option, but that requires more changes...

thgreasi: I think I will keep this as is, since doing `[PickDeferred<Release, 'id' | 'belongs_to__application'>?]` feels wrong given that we do `$top: 2`.

test/15_target-hostapp.ts
```diff
@@ -433,47 +449,85 @@ export default () => {
...
-				expect(resp.body.d[0]['supervisor_version']).to.equal(
-					supervisorVersion,
-				);
+					const supervisorVersion = 'v12.3.5';
+					const newOsVersion = 'balenaOS 2.88.5+rev1';
+					await invalidatedReleaseDevice.patchStateV2({
+						local: {
+							supervisor_version: supervisorVersion,
+							os_version: newOsVersion,
+							os_variant: 'production',
```
joshbwlng, line 497: Doubt it matters, but we don't actually have any "os_variant=production" device records in our production db. Maybe you meant "prod" here?

src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -202,58 +211,110 @@ async function getOSReleaseResource(
...
+			`Found more than one hostApp release matching version ${osVersion} ${osVariant} for device type ${deviceTypeId}.`,
+		);
+	}
+
+	const [release] = releases;
+	return release;
 }
 
-function normalizeVariant(variant: string) {
+function getRevisionFromSemver(parsedOsVersion: SemVer) {
```
joshbwlng, line 306: Personal opinion: I strongly prefer typing function returns instead of letting the internals implicitly type the return value. This helps to reduce accidentally breaking things, and also makes the typing more explicit and intentional.

src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -202,58 +211,110 @@ async function getOSReleaseResource(
...
-						},
+						// This effectively normalizes the versions string and
+						// compares it with the `semver` computed term, but we are
+						// using the individual fields so that DB indexes can also be used.
+						semver_major: parsedOsVersion.major,
+						semver_minor: parsedOsVersion.minor,
+						semver_patch: parsedOsVersion.patch,
+						semver_prerelease: parsedOsVersion.prerelease.join('.'),
+						semver_build: parsedOsVersion.build.join('.'),
+						revision: getRevisionFromSemver(parsedOsVersion),
```
joshbwlng, line 224: `getRevisionFromSemver()` defaults to the number `0` while the `revision` field in the db defaults to `null`. Is this not a problem?

src/features/hostapp/hooks/target-hostapp.ts
```diff
@@ -315,51 +400,64 @@ async function checkHostappReleaseUpgrades(
...
-	// TODO: this validation should eventually use release_version or raw_version, not tags
-	const newHostappVersion = newHostappRelease.release_tag[0].value;
+	const newHostappVersion =
+		getBaseVersionFromReleaseSemverOrTag(newHostappRelease);
 
-	const releases = (await api.get({
-		resource: 'release_tag',
+	if (newHostappVersion == null) {
+		throw new BadRequestError(
+			`Could not find the version of hostapp release with this ID ${newHostappReleaseId}`,
```
joshbwlng, line 420:
```suggestion
			`Could not find the version for the hostapp release with this ID ${newHostappReleaseId}`,
```

test/15_target-hostapp.ts
```diff
@@ -276,7 +282,9 @@ export default () => {
 														$alias: 'is',
 														$expr: {
 															is: {
-																application: nucHostApp.id,
+																application: !isEsr
+																	? nucHostApp.id
+																	: nucEsrHostApp.id,
```
joshbwlng, line 287:
```suggestion
																application: isEsr
																	? nucEsrHostApp.id
																	: nucHostApp.id,
```

More readable

test/15_target-hostapp.ts
```diff
@@ -289,7 +297,9 @@ export default () => {
 								const [service] = serviceInstalls[0].installs__service;
 								expect(service).to.have.property(
 									'id',
-									fx.services['intel-nuc_service1'].id,
+									fx.services[
+										!isEsr ? 'intel-nuc_service1' : 'intel-nuc-esr_service1'
```
joshbwlng, line 301:
```suggestion
										isEsr ? 'intel-nuc-esr_service1' : 'intel-nuc_service1'
```
Same as above, easier to read/parse this way

test/15_target-hostapp.ts
```diff
@@ -340,73 +342,87 @@ export default () => {
...
-					.send({ should_be_operated_by__release: esrHostappReleaseId })
+					.patch(`/${version}/device(${device1.id})`)
+					.send({ should_be_operated_by__release: esrTagOnlyHostAppReleaseId })
 					.expect(200);
 				const { body } = await supertest(admin).get(
-					`/${version}/device(${device.id})?$select=should_be_operated_by__release`,
+					`/${version}/device(${device1.id})?$select=should_be_operated_by__release`,
 				);
 				expect(body.d[0]).to.not.be.undefined;
 				expect(body.d[0]['should_be_operated_by__release'].__id).to.equal(
```
joshbwlng, line 424: Can you also use `expectResourceToMatch()` here?

test/15_target-hostapp.ts
```diff
@@ -340,73 +342,87 @@ export default () => {
...
-				const { body } = await supertest(admin)
-					.get(
-						`/${version}/device(${preprovisionedDevice.id})?$select=should_be_operated_by__release`,
-					)
-					.expect(200);
-				expect(body.d[0]).to.have.property(
-					'should_be_operated_by__release',
-					null,
-				);
+				it(`should remove the target preprovisioned hostapp and link with the reported upgraded ${titlePart} release, if it is an implied downgrade`, async () => {
```
joshbwlng, line 383: Maybe a title something like this is easier to understand: `should move device from older preprovisioned release to upgraded ${titlePart} release reported in state patch`. What do you think?

src/features/hostapp/hooks/target-hostapp.ts
thgreasi: The TS team advises to let inference do its work in most cases, but I do see your point. I've typed this and moved the defaulting to `0` outside of the function since on second thought, it isn't this helper's responsibility.

src/features/hostapp/hooks/target-hostapp.ts
thgreasi: Not it isn't, since atm at least we only support provisioning w/ finalized OS releases.

thgreasi: That was interesting actually... let me check what would mean to make the API support provisioning w/ raft OS releases right away :smiling_imp:

test/15_target-hostapp.ts
thgreasi: Yes :)
I was just trying to keep the diff as short as possible to be kind w/ the reviewers, but since you asked ..... :) :tada:

src/features/hostapp/hooks/target-hostapp.ts
thgreasi: Done :)

joshbwlng (approved):

PR #1589, opened 2024-03-30
Update os-config to support custom delta server

dcaputo-harmoni:
Changes needed to support custom delta servers such as [this](https://github.com/dcaputo-harmoni/open-balena-delta).  Needed because supervisor defaults to `delta.balena-cloud.com` otherwise.

### DIFF
--- a/src/features/os-config/index.ts
+++ b/src/features/os-config/index.ts
@@ -4,6 +4,7 @@ import {
 	DEVICE_CONFIG_OPENVPN_CA,
 	DEVICE_CONFIG_OPENVPN_CONFIG,
 	DEVICE_CONFIG_SSH_AUTHORIZED_KEYS,
+	DELTA_HOST,
 	LOGS_HOST,
 } from '../../lib/config.js';
 import { b64decode } from '../../lib/utils.js';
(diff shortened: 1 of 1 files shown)


thgreasi:
Hi,
Even though the supervisor does have a default for `https://delta.balena-cloud.com`
https://github.com/balena-os/balena-supervisor/blob/4596149d0e1e7922c94ea8066725b7659976613d/src/config/schema-type.ts#L25C13-L25C43
the supervisor does try to first fetch the deltaEndpoint from `config.json`:
https://github.com/balena-os/balena-supervisor/blob/4596149d0e1e7922c94ea8066725b7659976613d/src/config/schema.ts#L17

As a result, once you have properly configured DELTA_HOST on your open-balena-api, newly provisioned devices should already start up with the correct, custom provided deltaEndpoint. Can you confirm that?

It would be great if we could avoid hardcoding `balena-cloud.com` in the open-balena-api's sources, and on top of that avoid the additional payload in the os-config endpoint for all non `balena-cloud.com` that this change would add.

dcaputo-harmoni:
@thgreasi thanks for the quick reply!  Can you point me to the code where newly provisioned devices are assigned the deltaEndpoint variable in config.json?  We are not seeing this happen and I was looking for somewhere where this was done automatically - but could not find it. In the end we decided to build the newer version (1.3.0) of os-config into our base OS to allow this to happen.

thgreasi:
Hi @dcaputo-harmoni and I'm sorry for the long delay on replying to this...

As long as the `DELTA_HOST` env var gets populated on the API container, then the device confg generation will start populating the `deltaEndpoint` field on the config.json files that it generates:
https://github.com/balena-io/open-balena-api/blob/v20.2.16/src/features/device-config/device-config.ts#L116
https://github.com/balena-io-modules/balena-device-config/blob/v6.3.0/lib/config.coffee#L89

PR #1606, opened 2024-04-17
Tests: Use model typings for device type helper typings

Page-:
Change-type: patch

### DIFF
--- a/test/test-lib/device-type.ts
+++ b/test/test-lib/device-type.ts
@@ -7,32 +7,35 @@ const { api } = sbvrUtils;
 export const loadDefaultFixtures = () =>
 	setDefaultFixtures(
 		'deviceTypes',
-		new Proxy({} as Dictionary<Pick<DeviceType, 'id' | 'slug' | 'name'>>, {
-			get: async (obj, slug) => {
-				if (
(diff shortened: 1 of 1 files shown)


test/test-lib/device-type.ts
```diff
@@ -33,11 +28,11 @@ export const loadDefaultFixtures = () =>
...
 						},
-					})) as DeviceType;
+					})) as Pick<DeviceType, 'id' | 'slug' | 'name'>;
 					obj[slug] = deviceType;
 				}
 
 				return obj[slug as any];
 			},
-		}) as any as Dictionary<Promise<DeviceType>>,
+		}) as any as Dictionary<Promise<Pick<DeviceType, 'id' | 'slug' | 'name'>>>,
```
otaviojacobi, line 37: Can't we go with
```
new Proxy({} as Dictionary<Promise<Pick<DeviceType, 'id' | 'slug' | 'name'>>, {
```

And then 
```
obj[slug] = Promise.resolve(deviceType);
```

So that all this `as any as ....` mess goes away? (worked for me locally)

otaviojacobi (approved):

thgreasi (approved):

PR #1607, opened 2024-04-18
Drop the /device-types/v1/:deviceType/images route in favor of hostApps

thgreasi:
Change-type: major
See: https://balena.fibery.io/Work/Improvement/Stop-relying-on-device-types-v1-device-type.json-for-unrelated-things-257


src/features/device-types/device-types.ts
```diff
@@ -113,11 +110,11 @@ export const getImageSize = async (
...
-	const normalizedSlug = deviceType.slug;
+	const deviceTypeJson = deviceTypeInfo.latest;
+	const normalizedSlug = deviceTypeJson.slug;
 
 	if (buildId === 'latest') {
-		buildId = deviceType.buildId;
+		buildId = deviceTypeJson.buildId;
 	}
 
 	if (!deviceTypeInfo.versions.includes(buildId)) {
```
otaviojacobi: Would love if we could addtionality drop the entire `versions` of this interface too :)

thgreasi: Yea I can imagine... probably better to do it in the same PR

src/features/device-types/device-types.ts
```diff
@@ -113,16 +114,107 @@ export const getImageSize = async (
...
+			$orderby: [
+				// prefer finalized releases over draft
+				{ is_final: 'desc' },
+				// prefer the highest semver wise release
+				{ semver_major: 'desc' },
+				{ semver_minor: 'desc' },
+				{ semver_patch: 'desc' },
+				{ revision: 'desc' },
+				{ semver_prerelease: 'desc' },
+				// prefet prod over dev
```
otaviojacobi: // prefer

src/features/device-types/device-types.ts
```diff
@@ -94,36 +80,133 @@ export const getAccessibleDeviceTypeJsons = async (
...
+									},
+								},
+							},
+						},
+					},
+				},
+				status: 'success',
+				...(parsedOsVersion === 'latest' && {
+					is_invalidated: false,
+					// Avoid any ESR-loooking release that might have been accidentally
```
otaviojacobi: loooking

src/features/device-types/device-types.ts
```diff
@@ -1,12 +1,16 @@
...
 import { errors } from '@balena/pinejs';
 
-import { captureException } from '../../infra/error-handling/index.js';
+import {
+	captureException,
+	ThisShouldNeverHappenError,
+} from '../../infra/error-handling/index.js';
 
 import { getCompressedSize, getDeviceTypeJson } from './build-info-facade.js';
-import type { DeviceTypeInfo } from './device-types-list.js';
```
otaviojacobi, line 9: I think you still need this?

src/features/device-types/device-types.ts
```diff
@@ -111,19 +114,128 @@ export const getImageSize = async (
...
-	const normalizedSlug = deviceType.slug;
+	const deviceTypeJson = deviceTypeInfo.latest;
+	const normalizedSlug = deviceTypeJson.slug;
+
+	const parsedOsVersion =
+		buildId === 'latest' ? 'latest' : semver.parse(buildId);
+	if (parsedOsVersion == null) {
+		throw new UnknownVersionError(slug, buildId);
+	}
+	if (parsedOsVersion !== 'latest' && parsedOsVersion.major >= 2000) {
```
joshbwlng: I don't really understand why major >= 2000 means ESR, but I see we reference this magic number again below - maybe better to put it in a variable to make it easier to understand?

test/02_device-types.ts
```diff
@@ -278,121 +288,80 @@ export default () => {
...
-						'2.0.2+rev2',
-						'2.0.2+rev2.dev',
-						'2.0.2+rev1',
-						'2.0.2+rev1.dev',
-					],
-					latest: '2.0.2+rev2',
-				});
-			});
-
-			it('should return a proper result for an alias', async () => {
```
joshbwlng, line 297: Does the new endpoint still support aliases?

thgreasi: You mean the OData replacement of `/device-types/v1/:deviceType/images`?
If so then yes the model supports aliases
https://docs.balena.io/reference/api/resources/balenaos_versions
https://github.com/balena-io/balena-sdk/blob/v23.2.0/src/models/device-type.ts#L174-L180

If you mean the /device-types/v1/:deviceType/images/:version/download-size then yes it does support aliases, which get resolved near the top with:
```
export const getImageSize = async (
	resinApi: typeof sbvrUtils.api.resin,
	slug: string,
	buildId: string,
): Promise<number> => {
	const deviceTypeJson = await getDeviceTypeJsonBySlug(resinApi, slug);
	const normalizedSlug = deviceTypeJson.slug;
```
which is a good reminder to replace w/ a call to the model in a follow-up actually

src/features/device-types/device-types.ts
```diff
@@ -105,25 +108,138 @@ export const getDeviceTypeJsonBySlug = async (
...
-	if (!deviceTypeInfo.versions.includes(buildId)) {
+	const releases = await resinApi.get({
+		resource: 'release',
+		options: {
+			$top: parsedOsVersion === 'latest' ? 1 : 2,
+			$select: ['semver', 'variant'],
+			$filter: {
+				belongs_to__application: {
+					$any: {
+						$alias: 'a',
```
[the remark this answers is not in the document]
thgreasi: that's the standard pattern we follow in pine, b/c otherwise it would also increase the query size on top of requiring extra typing.

test/02_device-types.ts
joshbwlng: I just noticed one of the tests being dropped mentioned ensuring support for aliases, so just double-checking. :+1:

thgreasi: https://github.com/balena-io/open-balena-api/pull/2300

otaviojacobi (approved):

PR #1616, opened 2024-04-24
Add revokeUserRole to auth

otaviojacobi:
Change-type: minor

### DIFF
--- a/src/infra/auth/permissions.ts
+++ b/src/infra/auth/permissions.ts
@@ -62,6 +62,21 @@ export const assignUserPermission = (
 	tx: Tx,
 ) => getOrInsertId('user__has__permission', { user, permission }, tx);
 
+export const revokeUserRole = (user: number, role: number, tx: Tx) =>
+	api.Auth.delete({
+		resource: 'user__has__role',
+		options: {
+			$filter: {
+				user,
+				role,
+			},
+		},
+		passthrough: {
+			tx,
+			req: permissions.root,
+		},
+	});
+
 // api key helpers
 
 const getOrInsertApiKey = async (
--- a/src/index.ts
+++ b/src/index.ts
@@ -91,6 +91,7 @@ import {
 	getOrInsertRoleId,
 	assignUserPermission,
 	assignUserRole,
+	revokeUserRole,
 } from './infra/auth/permissions.js';
 import { createScopedAccessToken, createJwt } from './infra/auth/jwt.js';
 import { middleware as authMiddleware } from './infra/auth/index.js';
@@ -198,6 +199,7 @@ export const auth = {
 	getOrInsertRoleId,
 	assignUserPermission,
 	assignUserRole,
+	revokeUserRole,
 	getUserIDFromCreds,
 	registryAuth,
 	normalizeHandle,


joshbwlng:
@otaviojacobi I think this looks fine, but can you provide some context as to why its needed and/or where it will be used?

otaviojacobi:
> @otaviojacobi I think this looks fine, but can you provide some context as to why its needed and/or where it will be used?

Yes, this is my bad :) 
Basically, with SAML we need now to be able to remove the traditional 'default-user' from a user and add the new 'saml-user' - which is basically the same but without permissions to `renew token & generate api keys`.

See: https://github.com/balena-io/balena-api/pull/4980/files#diff-0b549fe83bafc3ff9c5350c19b595f402f9f8490120067f44eca71e2e8246c0cR216-R222

I thought it was weird to have this function on SAML flow and that it actually belongs as a helper from open-balena-api similar to its pairs (e.g. assignUserRole)

joshbwlng (approved):

src/infra/auth/permissions.ts
```diff
@@ -62,6 +62,21 @@ export const assignUserPermission = (
 	tx: Tx,
 ) => getOrInsertId('user__has__permission', { user, permission }, tx);
 
+export const revokeUserRole = (user: number, role: number, tx: Tx) =>
```
thgreasi, line 65: I'm fine with it, but is there any reason that you prefer having a helper rather than doing this inline like we already do in other places like v ?
See: https://github.com/balena-io/balena-api/blob/ed13684d7f4e55f49e4ac2f66de059478983bbce/src/lib/accounts.ts#L15


I mean pine itself is supposed to be our API to the DB, so such a simple wrapper seems to not be adding much.

thgreasi (approved):

src/infra/auth/permissions.ts
otaviojacobi: @thgreasi Consistency really. We do have `assignUserPermission` as a function provided by open-balena-api.
One example is this piece of code: https://github.com/balena-io/balena-api/pull/4980/files#diff-0b549fe83bafc3ff9c5350c19b595f402f9f8490120067f44eca71e2e8246c0cR157-R162

IMO it reads much better than if we were inlining as:
```ts
const samlUserRoleId = await getRoleId('saml-user', tx);
const defaultUserRoleId = await getRoleId('default-user', tx);

api.Auth.delete({
		resource: 'user__has__role',
		options: {
			$filter: {
				user: existingUser.id,
				role:  defaultUserRoleId,
			},
		},
		passthrough: {
			tx,
			req: permissions.root,
		},
	});
await auth.assignUserRole(existingUser.id, samlUserRoleId, tx);
```

I would also be ok with not having any helper at all - but then we would be repeating mandatory things (e.g. permissions.root) all over the place too, which isn't as nice as with the helpers

otaviojacobi: And also at the end of the day even if we can just do pine queries they can become slightly too long to say too litle so we end up creating these functions on balena api anyway, see: https://github.com/balena-io/balena-api/pull/4980/files#diff-279973c749874c2e77d071ea426b890e8ba40ba2661d3208e6dfc2902cb4d1b7R35-R55

Adding it on open-balena-api at least makes it belong in a proper place to be shared

Issue #1612, opened 2024-04-19
v19.1.19 Error: Cannot find module '/usr/src/app/init.js' imported from /usr/src/app/index.js


jenish1231:
![image](https://github.com/balena-io/open-balena-api/assets/19236975/5f876b47-44e9-4e81-8158-3dd4a172e7af)

thgreasi:
Hi,
Did you have chance to narrow this down to which version this first started happening?

Kind regards,
Thodoris

thgreasi:
Can you also clarify whether you are trying to run the Docker image or trying to import the open-balena-api as a node module?

jenish1231:
@thgreasi I tried to use it as a Docker image. Setting `PRODUCTION_MODE: true` solved the problem. Thank you for your quick response.

thgreasi:
Hi @jenish1231 PR #1618 should allow you to run open-balena-api even without the `PRODUCTION_MODE: true` variable.

PR #1618, opened 2024-04-25
correct startup when run as a standalone container

ab77:
tested: https://github.com/balena-io/open-balena/actions/runs/8929931824/job/24529013349?pr=141


otaviojacobi (dismissed):

otaviojacobi (approved):

thgreasi:
This might also be properly fixing #1612

PR #1629, opened 2024-05-09
Update overall status computed field

JSReds:
Change-type: major


test/24_device_additions.ts
```diff
@@ -27,6 +27,17 @@ export default () => {
 		describe('Device additions', function () {
 			let pineUser: PineTest;
 			const ctx: AnyObject = {};
+
+			const itShouldHaveOverallStatus = (
+				testCase: string,
+				deviceName: string,
+				overallStatus: string,
+			) => {
+				it(testCase.replace('${OVERALL_STATUS}', overallStatus), async () => {
```
thgreasi, line 36: NIT:
```suggestion

			const itShouldHaveOverallStatus = (
				overallStatus: string,
				restTitle: string,
				deviceName: string,
			) => {
				it(`should have a(n) ${overallStatus} ${restTitle}, async () => {
```
This way the test case would naturally read v :)
```
					itShouldHaveOverallStatus(
						versions.gt(version, 'v6') ? 'reduced-functionality' : 'idle',
						'when the heartbeat is Offline',
						'deviceVpnOnHeartbeatOffline',
					);
```

test/24_device_additions.ts
```diff
@@ -89,65 +110,81 @@ export default () => {
...
 
 				describe('Given a device downloading a multicontainer release', () => {
-					it('should properly calculate the overall_status', async () => {
-						await testOverallStatus(
-							ctx.user,
-							ctx.deviceUpdating.id,
-							'updating',
-						);
-					});
+					itShouldHaveOverallStatus(
```
thgreasi, line 183: Let's also add 2 more test cases for an updating container while vpn is off and while heartbeat is off & test overall_progress for them as well.

src/features/devices/models/device-additions.ts
```diff
@@ -113,39 +204,55 @@ export const addToModel = (abstractSql: AbstractSqlModel) => {
...
+			[
+				'When',
+				hasPartialConnectivity,
+				['EmbeddedText', 'reduced-functionality'],
+			],
+			['Else', ['EmbeddedText', 'operational']],
 		],
 	});
 
+	console.log(abstractSql.tables['device']);
```
thgreasi, line 254:
```suggestion
```

src/features/devices/models/device-additions.ts
```diff
@@ -4,8 +4,139 @@ import type {
...
+	],
+	[
+		'And',
+		['Equals', ['ReferencedField', 'device', 'is online'], ['Boolean', false]],
+		[
+			'In',
+			['ReferencedField', 'device', 'api heartbeat state'],
+			['EmbeddedText', 'online'],
+			['EmbeddedText', 'timeout'],
+		],
```
thgreasi, line 78: Shouldn't we only check for online heartbeat, so that this matches w/ how we check the heartbeat in the vpn false case?
```suggestion
						'Equals',
						['ReferencedField', 'device', 'api heartbeat state'],
						['EmbeddedText', 'online'],
```
I mean, for a device w/ the vpn disabled, ie a device that we can infer its state only from its heartbeat state, timeout is probably the only `reduced-functionality` case. My reasoning:
* vpn disabled & heartbeat online => operational
* vpn disabled & heartbeat offline/unknown => disconnected
* vpn disabled & heartbeat timeout => reduced-functionality (even this doesn't feel concrete enough, but that's the only that's left)

src/translations/v6/v6.ts
```diff
@@ -92,6 +96,101 @@ export const getV6Translations = (abstractSqlModel = v6AbstractSqlModel) =>
...
+							'And',
+							['Not', ['ReferencedField', 'device', 'is online']],
+							[
+								'Equals',
+								['ReferencedField', 'device', 'status'],
+								['EmbeddedText', 'Shipped'],
+							],
+						],
+					],
+					['ToLower', ['ReferencedField', 'device', 'status']],
```
thgreasi, line 124: Should we also drop these non-used & already dropped HoD states, rather re-adding them?

src/features/devices/models/device-additions.ts
```diff
@@ -4,8 +4,139 @@ import type {
...
+	],
+];
+
+export const isPostProvisioning: EqualsNode = [
+	'Equals',
+	['ReferencedField', 'device', 'provisioning state'],
+	['EmbeddedText', 'Post-Provisioning'],
+];
+// This check does not double check heartbeat state as it is already checked from isOverallOffline which runs before
+export const hasPartialConnectivity: OrNode = [
```
thgreasi, line 59: How about v so that vpn disabled&off & heartbeat timeout gives reduced-functionality?
```
-- In an ideal world:
partial = CASE
	WHEN vpn disabled
	THEN heartbeat = timeout
	ELSE (vpn on & heartbeat != online) || (vpn off & heartbeat = online,timeout)

=>

partial = CASE
	WHEN heartbeat = timeout THEN True
	ELSE (vpn enabled && vpn on & heartbeat IN offline)
		|| (vpn off & heartbeat = online && vpn enabled)

approximately =>

partial = CASE
	WHEN heartbeat = timeout THEN True
	ELSE (vpn on & heartbeat IN offline)
		|| (vpn off & heartbeat = online && vpn enabled)

=>

partial =
	|| heartbeat = timeout
	|| (vpn on & heartbeat IN offline/unknown)
	|| (vpn off & heartbeat = online && vpn enabled)
```

src/features/devices/models/device-additions.ts
```diff
@@ -55,12 +56,69 @@ export const isPostProvisioning: EqualsNode = [
 	['ReferencedField', 'device', 'provisioning state'],
 	['EmbeddedText', 'Post-Provisioning'],
 ];
+
+const VPNConfigEnabled: NotEqualsNode = [
```
thgreasi, line 60:
```suggestion
const isVpnEnabled: NotEqualsNode = [
```

src/features/devices/models/device-additions.ts
```diff
@@ -113,36 +204,50 @@ export const addToModel = (abstractSql: AbstractSqlModel) => {
...
-					'Exists',
+					'And',
 					[
-						'SelectQuery',
-						['Select', []],
-						['From', ['Table', 'image install']],
+						'In',
+						['ReferencedField', 'device', 'api heartbeat state'],
+						['EmbeddedText', 'online'],
+						['EmbeddedText', 'timeout'],
```
thgreasi, line 212: Should we add the same check in the `overall_progress` (and add a v6 translation)?
I kind of can see the use case of showing a progress bar in a reduced-functionality device... so just checking whether this was discussed. 

We would effectively just need to change
```
			[
				'When',
				[
						'NotIn',
						['ReferencedField', 'device', 'api heartbeat state'],
						['EmbeddedText', 'online'],
						['EmbeddedText', 'timeout'],
				],
				// Otherwise if the device is offline but has previously been online we return no info
				['Null'],
			],
```
to
```
			[
				'When',
				isOverallOffline,
				// Otherwise if the device is offline but has previously been online we return no info
				['Null'],
			],
```

Not a blocker though, since we can change it later.

test/24_device_additions.ts
```diff
@@ -24,15 +24,59 @@ export default () => {
...
 		describe('Device additions', function () {
 			let pineUser: PineTest;
 			const ctx: AnyObject = {};
+
+			const itShouldHaveOverallStatus = (
+				overallStatus: string,
+				restTitle: string,
+				deviceName: string,
+			) => {
+				it(`should have "${overallStatus}" ${restTitle}`, async () => {
```
thgreasi, line 52:
```suggestion
				it(`should have overall_status = "${overallStatus}" ${restTitle}`, async () => {
```
or
```suggestion
				it(`should have overall_status = "${overallStatus}" ${restTitle}`, async () => {
```

thgreasi (commented):
"Update the overall_status computed field in the /resin model"

thgreasi (approved):

PR #1642, opened 2024-05-21
Expire refreshed tokens within the same time as initial token.

fisehara:
Change-type: major


src/infra/auth/jwt.ts
```diff
@@ -181,7 +181,10 @@ export const createJwt = (
 	payload: AnyObject,
 	jwtOptions: jsonwebtoken.SignOptions = {},
 ): string => {
-	_.defaults(jwtOptions, { expiresIn: EXPIRY_SECONDS });
+	const expiresIn = payload.exp
+		? Math.floor(payload.exp - Date.now() / 1000)
+		: EXPIRY_SECONDS;
+	_.defaults(jwtOptions, { expiresIn });
 	delete payload.iat;
 	delete payload.exp;
```
thgreasi, line 189: Let's do:
```suggestion
	// to exactly match what jsonwebtoken does
	if (typeof payload.exp === 'undefined') {
		_.defaults(jwtOptions, { expiresIn: EXPIRY_SECONDS }); // feel free to drop lodash from here
	} else {
		// by not setting `expiresIn`, jsonwebtoken.sign will preserve the original expiry found in the tokem
		delete jwtOptions.expiresIn;
	}
	delete payload.iat;
```
to exactly match what jsonwebtoken does
See: https://github.com/auth0/node-jsonwebtoken/blob/bc28861f1fa981ed9c009e29c044a19760a0b128/sign.js#L162-L207C7

fisehara: @thgreasi I also wonder I we want to guardrail the `payload.exp` with something like payload.exp = payload.exp > (Date.now () / 1000) + EXPIRY_SECONDS? EXPIRE_SECONDS : payload.exp ?

test/04_session.ts
```diff
@@ -124,10 +126,43 @@ export default () => {
 				});
 
 				it('should be refreshable with /user/v1/refresh-token', async function () {
+					// wait 2 seconds to make sure the token is already starting to expire
+					await new Promise((resolve) => setTimeout(resolve, 2000));
```
Page-, line 130: Use
```ts
import { setTimeout } from 'timers/promises';
```
and
```suggestion
					await setTimeout(2000);
```

thgreasi (approved):

PR #1643, opened 2024-05-21
better handling of exceptions from AWS/S3

ab77:
.. required (ideally) by https://github.com/balena-io/open-balena/pull/141

We have two contracts repos, one is [private](https://github.com/balena-io/private-contracts/tree/master/contracts/hw.device-type) and the other [public](https://github.com/balena-io/contracts/tree/master/contracts/hw.device-type).

So open-balena-api can't access the private repo, to check which device types are private, but it can enumerate the entire images bucket, which has with public `List` ACL enabled. Since it can see private devices in the enumerated list, the API will try to access various assets, but will get back a 403 from S3 API, since the objects themselves don't have public-read enabled. This behaviour produces a lot of exceptions in the API logs and this is what this PR attempts to make better.

A more complete solution would probably require rethinking private/public DT handling in open-balena-api vs. balena-api.

.. in any case, events now reflected in the logs as follows:

```
...
May 21 19:30:31 45abc1b164c4 api[22260]: Forbidden (403): images/coral-dev/2.101.1+rev1/IGNORE belongs to a private device type or has incorrect permissions
May 21 19:30:31 45abc1b164c4 api[22260]: Forbidden (403): images/asus-tinker-edge-t/2.101.1+rev1/IGNORE belongs to a private device type or has incorrect permissions
...
May 21 22:58:39 45abc1b164c4 api[46805]: AccessDenied (403): images/imx8mmebcrs08a2/2.108.1+rev3/device-type.json belongs to a private device type or has incorrect permissions
May 21 22:58:39 45abc1b164c4 api[46805]: AccessDenied (403): images/bigmak-imx8m-plus/2.114.6/device-type.json belongs to a private device type or has incorrect permissions
```

* [StartLimitIntervalSec](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#StartLimitIntervalSec=interval)

### DIFF
--- a/src/features/device-types/storage/s3.ts
+++ b/src/features/device-types/storage/s3.ts
@@ -57,6 +57,21 @@ function createS3Client() {
 
 const s3Client = createS3Client();
 
+function logUnauthenticatedWarning(
+	clientS3: UnauthenticatedS3Facade | AWSWrapper.AWS.S3,
+	pathS3: string,
+	err: any,
+): void {
+	if (
+		clientS3 instanceof UnauthenticatedS3Facade &&
+		[401, 403].includes(err.statusCode)
+	) {
+		console.warn(
+			`${err.code} (${err.statusCode}): ${pathS3} belongs to a private device type or has incorrect permissions`,
+		);
+	}
+}
+
 async function getFileInfo(s3Path: string) {
 	const req = s3Client.headObject({
 		Bucket: S3_BUCKET,
@@ -66,11 +81,16 @@ async function getFileInfo(s3Path: string) {
 }
 
 export async function getFile(s3Path: string) {
-	const req = s3Client.getObject({
-		Bucket: S3_BUCKET,
-		Key: s3Path,
-	});
-	return await req.promise();
+	try {
+		const req = s3Client.getObject({
+			Bucket: S3_BUCKET,
+			Key: s3Path,
+		});
+		return await req.promise();
+	} catch (err) {
+		// catch errors for private device types when running unauthenticated
+		logUnauthenticatedWarning(s3Client, s3Path, err);
+	}
 }
 
 export async function getFolderSize(
@@ -128,6 +148,8 @@ export async function fileExists(s3Path: string): Promise<boolean> {
 		await getFileInfo(s3Path);
 		return true;
 	} catch (err) {
+		// catch errors for private device types when running unauthenticated
+		logUnauthenticatedWarning(s3Client, s3Path, err);
 		if (err.statusCode === 404) {
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


src/features/device-types/storage/s3.ts
```diff
@@ -128,6 +139,15 @@ export async function fileExists(s3Path: string): Promise<boolean> {
...
 	} catch (err) {
+		// catch errors for private device types when running unauthenticated
+		if (s3Client instanceof UnauthenticatedS3Facade) {
+			if ([401, 403].includes(err.statusCode)) {
+				console.warn(
+					`${err.code} (${err.statusCode}): ${s3Path} belongs to a private device type or has incorrect permissions`,
+				);
+				return false;
+			}
+		}
```
joshbwlng, line 150: This is mostly duplicate code with the other log further up in the file, maybe we could consolidate this into a shared function that only logs the warning. Mayne something like this:
```typescript
function logUnauthenticatedWarning(s3Client, s3Path, err): void {
    if (s3Client instanceof UnauthenticatedS3Facade && [401, 403].includes(err.statusCode)) {
        console.warn(
            `${err.code} (${err.statusCode}): ${s3Path} belongs to a private device type or has incorrect permissions`,
        );
    }
}
```

Then we could simplify the `return false` logic by only doing that in the `if` block below with:
```typescript
if ([401, 403, 404].includes(err.statusCode)) {
    return false;
}
```

What do you think?

src/features/device-types/storage/s3.ts
```diff
@@ -57,6 +57,17 @@ function createS3Client() {
 
 const s3Client = createS3Client();
 
+function logUnauthenticatedWarning(clientS3, pathS3, err): void {
```
joshbwlng, line 60:
```suggestion
function logUnauthenticatedWarning(clientS3: UnauthenticatedS3Facade | AWSWrapper.AWS.S3, pathS3: string, err: any): void {
```

Missing argument types

joshbwlng (approved):

src/features/device-types/storage/s3.ts
```diff
@@ -66,11 +81,16 @@ async function getFileInfo(s3Path: string) {
...
-	return await req.promise();
+	try {
+		const req = s3Client.getObject({
+			Bucket: S3_BUCKET,
+			Key: s3Path,
+		});
+		return await req.promise();
+	} catch (err) {
+		// catch errors for private device types when running unauthenticated
+		logUnauthenticatedWarning(s3Client, s3Path, err);
```
thgreasi, line 92: @ab77 This atm silences ALL errors. Shouldn't we re-throw the error when this is a different type of error that logUnauthenticatedWarning didn't handle?

joshbwlng: Fair point, we should make sure other errors aren't ignored

PR #1648, opened 2024-05-28
Switch from using import assertions to import attributes

Page-:
Change-type: patch


thgreasi (approved):

thgreasi:
Oh I see that I only did that for the top level imports... :facepalm:

PR #1652, opened 2024-06-04
Switch from `Object.prototype.hasOwnProperty.call` to `Object.hasOwn`

Page-:
Change-type: patch


JSReds (approved):
Approval based on that NOTE: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

PR #1654, opened 2024-06-04
Enforces login with auth permission

otaviojacobi:
We are moving towards more granular login permissions based on roles. The main advantage is that this allow for more control over different permissions, for example, users added by SAML won't be able to do `auth.credentials_login` (and `auth.social_service_account_login` still to be done in balena-api). More over, this also allow for more control, for example, in the future I could see orgs configuring their users to only allow specific login methods (altough this is a stretch, the work here creates the foundation for it to be possible).

### DIFF
--- a/src/features/auth/login.ts
+++ b/src/features/auth/login.ts
@@ -3,6 +3,7 @@ import { errors, sbvrUtils } from '@balena/pinejs';
 import { comparePassword, findUser } from '../../infra/auth/auth.js';
 import { loginUserXHR } from '../../infra/auth/jwt.js';
 import { captureException } from '../../infra/error-handling/index.js';
+import { permissions } from '@balena/pinejs';
 
 import type { SetupOptions } from '../../index.js';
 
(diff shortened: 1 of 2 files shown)


joshbwlng:
@otaviojacobi Could you give some background on why we're adding `auth.credentials_login`, and also some tests should probably be added.

src/features/auth/login.ts
```diff
@@ -23,6 +24,11 @@ export const login =
 				if (!user) {
 					throw new NotFoundError('User not found.');
 				}
+				const userPermissions = await permissions.getUserPermissions(user.id);
```
Page-, line 27: I would put this after the password check, that way if that fails we avoid a db request. Although also this check could potentially be in the `onLogin` if it's unused in open-balena?

otaviojacobi: @Page- It is not directly used on open-balena-api right now but I could see it being used in the future, more over, credentials_login is something of the default-user that exists already on open-balena so in my mind it makes sense to be here instead of on the api callback

Page- (approved):

src/features/auth/login.ts
```diff
@@ -28,6 +29,12 @@ export const login =
 				if (!matches) {
 					throw new BadRequestError('Current password incorrect.');
 				}
+
+				const userPermissions = await permissions.getUserPermissions(user.id);
```
thgreasi, line 33: I think that you should be able to replace this w/ a `middleware.permissionRequired('auth.credentials_login', { statusCode: 400, message: 'User not allowed to login.' })`, but
1. we can't do it that way if we want to only make this check after validating the user's password.
2. it's not a big deal
3. we atm do not support that kind of a second argument

PR #1657, opened 2024-06-06
Do not rewrite the field names when aliasing relationships

Page-:
This can cause problems because looking up a rewritten field will fail because it does not actually exist, and instead we expect the one field to point to two different places via aliasing

Change-type: patch


Page-:
Superseded by #1659

PR #1662, opened 2024-06-10
Add unique email constraint on user model

otaviojacobi:
This enforces database level uniqueness on the email field. I opted for keeping the current hooks alongside this (see https://github.com/balena-io/open-balena-api/blob/d3b61faad9d3aeb6fa22f129e70c2a75ce2f3070/src/features/auth/hooks/validate-username-email.ts#L45) in order to keep the error messages clean. In the case of a PATCH for a user with an email that is duplicated, without this hook api would respond with 200 and just don't do anything, so I decided to keep it. As the case where the hook fail is a very rare race condition I could not figure it out how to test the constraint.

Change-type: minor

### DIFF
--- a/src/balena-init.sql
+++ b/src/balena-init.sql
@@ -201,3 +201,5 @@ ADD CONSTRAINT "user$M+9koFfMHn7kQFDNBaQZbS7gAvNMB1QkrTtsaVZoETw=" CHECK (NOT (
 		AND "email" IS NOT NULL
 	)
 ));
+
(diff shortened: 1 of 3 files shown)


joshbwlng:
I'd guess we'd want to reduce hook overhead and let the database error on duplicates at some point in the future, but I get the point of not wanting to do this now. If I understand correctly that would probably result in a `major` and not just a `minor` (response code changes, error message changes). Dropping the hook check would allow for testing the database-level constraint later on.

Page- (approved):

PR #1667, opened 2024-06-13
Fix refresh-token including additional jwt properties

thgreasi:
Change-type: patch

### DIFF
--- a/src/infra/auth/jwt.ts
+++ b/src/infra/auth/jwt.ts
@@ -47,6 +47,14 @@ export interface TokenUserPayload
 	authTime?: number;
 }
 
+const jwtValidFields = [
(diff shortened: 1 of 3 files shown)


src/infra/auth/jwt.ts
```diff
@@ -47,6 +47,14 @@ export interface TokenUserPayload
 	authTime?: number;
 }
 
+const jwtValidFields = [
```
thgreasi, line 50: We might later (during the bC cleanup) want to export this (so that consumers can augment it) and move the `_.pick` in the `createSessionToken()`.

fisehara (approved):
:tada: Good catch, thanks for bringing the pick down to open-balena-api

PR #1668, opened 2024-06-13
is_pinned_on__release: Add migration to sync 'is pinned on-release' with 'should be running-release' in device table

JSReds:
Depends-on: Async migration finalisation

Projects: https://balena.fibery.io/Work/Project/Server-side-pagination-for-devices-Cycle-3-539

Change-type: minor


src/migrations/00088-sync-is-pinned-on-release-to-should-be-running-release-on-device.sql
thgreasi, line 1: I think this should be 00090 now

thgreasi (approved):

PR #1670, opened 2024-06-14
Update @balena/pinejs and make use of model based typings

Page-:
Change-type: patch


src/features/device-provisioning/register.ts
```diff
@@ -79,6 +80,7 @@ export const register: RequestHandler = async (req, res) => {
 			const device = await resinApiTx.post({
 				resource: 'device',
 				body: {
+					// @ts-expect-error This is for backwards compatibility :(
```
fisehara, line 83: In what sense?

src/infra/auth/auth.ts
```diff
@@ -183,7 +183,8 @@ const getUserQuery = _.once(
 			},
 		}) as PreparedFn<
 			{ key: string },
-			Promise<Array<PickDeferred<DbUser['Read'], (typeof userFields)[number]>>>
+			Promise<Array<PickDeferred<DbUser['Read'], (typeof userFields)[number]>>>,
+			DbUser
```
fisehara, line 187: In this module there is no naming issue with 'User' vs 'DbUser' should we better us 'User' instead of DbUser?

init.ts
```diff
@@ -50,6 +50,7 @@ async function onInitHooks() {
 	setSyncSettings({
 		'hw.device-type': {
 			resource: 'device_type',
+			// @ts-expect-error These typings need fixing :(
```
fisehara, line 53: But hwat is the problem?

Page-: Basically these typings are hideously generic to the point it almost needs the same level of work as the pinejs-client did to make these also work. Since `resource` can be any `keyof Model` it means that `uniqueKey` needs to be a field of `Model[keyof Model]` and the only field that exists for all of them is `id` which means `slug` is rejected. So it needs to know the specific resource it's pointing at to look at the fields for that resource, and this has to work along with inference as well, it's not impossible to type but it's a real headache (with the `refersTo` version being even worse I think). I'll take another bash at these

src/infra/auth/auth.ts
Page-: It looks like the clashing `User` has been removed so the code could change all the `DbUser` references not but it doesn't make sense to be part of this PR imo

src/features/device-provisioning/register.ts
Page-: It's not generically typed but I've added a single commit so you can see the somewhat hardcoded version that works for what we have

src/features/ci-cd/hooks/track-latest-release.ts
```diff
@@ -25,10 +25,10 @@ const updateLatestRelease = async (
 			$select: 'id',
 			$expand: {
 				owns__release: {
-					$select: 'id',
+					$select: 'id' as const,
 					$top: 1,
 					// the most recently started build should be the latest, to ignore variable build times
-					$orderby: { start_timestamp: 'desc' },
+					$orderby: { start_timestamp: 'desc' as const },
```
otaviojacobi, line 31: In the places where we need two `as const` casts on the same query,  can't' we just do it once in the whole dict to reduce a bit the `as const` everywhere? 
```suggestion
	const appsToUpdate = await api.get({
		resource: 'application',
		options: {
			$select: 'id',
			$expand: {
				owns__release: {
					$select: 'id',
					$top: 1,
					// the most recently started build should be the latest, to ignore variable build times
					$orderby: { start_timestamp: 'desc' },
					$filter: trackableReleaseFilter,
				},
			},
			$filter: {
				should_track_latest_release: true,
				owns__release: {
					$any: {
						$alias: 'r',
						$expr: {
							r: {
								id: { $in: releaseIds },
							},
						},
					},
				},
			},
		} as const,
	});
```

Page-: Yeah we can, this was primarily a way of showing specifically what is causing the need for the `as const`

src/features/contracts/index.ts
```diff
@@ -17,6 +17,12 @@ import { MINUTES } from '@balena/env-parsing';
 import { captureException } from '../../infra/error-handling/index.js';
 import { scheduleJob } from '../../infra/scheduler/index.js';
 
+type ValidContractResources =
+	| 'device_type'
+	| 'cpu_architecture'
+	| 'device_family'
+	| 'device_manufacturer'; // | 'device_type_alias'
```
joshbwlng, line 24: Unintentional comment?

Page-: Yeah it's unintentional, I'll remove it

src/infra/auth/auth.ts
Page-: #1682

Page-:
I can put the second commit as a separate PR if people want but I thought it made sense to live here to show the additional things that model based typings unlock

src/features/applications/hooks/slug.ts
```diff
@@ -59,7 +59,7 @@ hooks.addPureHook('PATCH', 'resin', 'application', {
 						id: { $in: ids },
 					},
 				},
-			});
+			} as const);
```
otaviojacobi, line 62: I am not sure what is the best pattern here - 
1 - add const on the whole object as we know we are not changing it
2 - add const only where it is necessary (in this case, just the `organization: { ... } as const`)

No blockers on this, just wondering which is best

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -92,7 +92,7 @@ const createReleaseServiceInstalls = async (
 						}),
 			},
 		},
-	})) as Array<
+	} as const)) as Array<
```
otaviojacobi, line 95: All this``as` cast can be gone

src/features/device-config/download.ts
```diff
@@ -29,7 +29,7 @@ const getApp = async (appId: number, req: Request) => {
 				},
 			},
 		},
-	})) as
+	} as const)) as
```
otaviojacobi, line 32: Any reason for keeping this cast?

src/features/host-os-access/access.ts
```diff
@@ -33,15 +32,14 @@ export async function hostOSAccess(req: Request, res: Response): Promise<void> {
 
 	try {
 		const allowedDevices = (await api.resin.post({
-			resource: 'device',
+			url: `device(${device.id})/canAccess`,
 			id: device.id,
```
otaviojacobi, line 36: can we also remove the id here?

otaviojacobi:
This all looks very nice, I would just like to have the typings removed on the places we added the "as const" cast (or the const cast removed) - The reason being, the ones we haven't touched we know they just need improving in following PRs, but seeing one with const cast (meaning - we know this is using the new typings) and with the `as` casting at the end makes me think it is there for a reason rather than just we forgot to remove

src/features/device-config/download.ts
Page-: I think the reason for all these casts being kept is because they rely on nested $expand support, which didn't exist at the point I initially wrote the PR

otaviojacobi (approved):

Issue #1095, opened 2022-08-24
Should change device's should_be_running__release to is_pinned_on__release


thgreasi:
In v7 we should have a `device.is_pinned_on__release` fact type, which is effectively what v6's `device.should_be_running__release` is.

PR #1672, opened 2024-06-19
Jsreds/add device is pined on release 2c

JSReds:
Resolves: #1095

Project: https://balena.fibery.io/Work/Project/Server-side-pagination-for-devices-Cycle-3-539


src/features/hostapp/hooks/target-hostapp.ts
thgreasi, line 1: The changed in this file need to be reverted.

thgreasi (approved):
LGTM but let's wait for an extra approval on both oB & bC

otaviojacobi (approved):

PR #1674, opened 2024-06-19
Update dependencies

Page-:
Update @balena/abstract-sql-compiler from 9.1.4 to 9.2.0
Update @balena/pinejs from 17.0.0 to 17.0.3
Update pinejs-client-supertest from 2.0.3 to 2.0.4

Change-type: patch


src/balena-model.ts
```diff
@@ -2,6 +2,8 @@
  * This file is auto-generated with `npm run generate-model-types`
  */
```
otaviojacobi, line 3: @Page- can we also remove this other one from the generation on the package.json? I don't think it is necessary considering now abstract-sql-to-ts is already putting it in

Page-: sure

otaviojacobi (approved):

PR #1683, opened 2024-06-26
Remove the my_application resource from the resin model

thgreasi:
Users should be using `user__has_direct_access_to__application` instead.

Change-type: patch


joshbwlng:
@thgreasi A couple quick questions:
- Since we're removing something from the model, is this not considered a breaking change?
- I don't see how this PR relates to the linked Fibery document, is there a TLDR somewhere?

Page- (approved):

PR #1684, opened 2024-06-26
Release the v7 model

thgreasi:
Drops:
* `device.logs_channel`
* `device.vpn_channel`
* `device.is_managed_by__device`
  * `device.manages__device`
* `application.depends_on__application`
  * `application.is_dependent_on_by__application`
* `release.release_type`(`final`/`draft`) in favor of `release.is_final` (boolean)
* the `gateway_download` resource
  * `image.is_downloaded_by__device`
  * `device.downloads__image`
* the `my_application` resource
* POSTs to the `device` and `application` resource with a text `device_type` field
* populating the response body with a status code message (eg 'OK')

Changes:
* the `user.actor`, `device.actor`, `application.actor` to return a deferred (`{ __id: number; }`) instead of a plain number when not `$expand`ed
* the `release.contract` from a string to a JSON field
* the `device.overall_status`'s
  * `offline` to `disconnected`
  * `idle` to
    * `operational` when the device is heartbeating, and is either connected to cloudlink or has the vpn configured to be off
    * `reduced-functionality` when the device has only a heartbeat or is only connected to cloudlink 

Replaces:
* `device.should_be_running__release` with `device.is_pinned_on__release`
* `release.should_be_running_on__device` with `release.is_pinned_to__device`

Adds:
* `device.should_be_managed_by__release`

Resolves: #1097
Change-type: minor

### DIFF
--- a/src/translations/v6/v6.ts
+++ b/src/translations/v6/v6.ts
@@ -9,7 +9,7 @@ import {
 import * as userHasDirectAccessToApplication from '../../features/applications/models/user__has_direct_access_to__application.js';
 import * as DeviceAdditions from '../../features/devices/models/device-additions.js';
 
-export const toVersion = 'resin';
+export const toVersion = 'v7';
 
 export const v6AbstractSqlModel = generateAbstractSqlModel(
 	new URL('v6.sbvr', import.meta.url),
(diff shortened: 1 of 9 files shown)


Page- (commented):
I haven't done a full review based on the changes listed yet (thanks so much for adding them) but I would like to see them also in the commit message so that they are persisted in the git repo itself as well as the PR description, it just makes it a bit easier to tie the changes together if you're coming from the git repo itself rather than github

thgreasi:
@Page- can you expand on your suggestion on making individual commits a bit?
Pretty much all of the changes in v7 are already in resin, so this PR was mostly just not copying the respective translations from v6. (ie we already have commits like for dropping the dependent device props on resin)
Would you prefer empty commits w/ just the respective title, or would you prefer me to copy-paste v6 as v7 and then doing a commit for each translation that gets removed? If it's the latter one, it would be a mess to also try to have the tests passing on each individual commit.

src/translations/v6/v6.ts
```diff
@@ -87,10 +87,12 @@ export const getV6Translations = (abstractSqlModel = v6AbstractSqlModel) => {
 			'depends on-application': ['Cast', ['Null'], 'Integer'],
 		},
 		'application-has-name': {
-			$toResource: `application-has-config var name$${toVersion}`,
+			// TODO: Figure out why toVersion doesn't work.
+			$toResource: `application-has-config var name$resin`,
 		},
 		'device-has-name': {
-			$toResource: `device-has-config var name$${toVersion}`,
+			// TODO: Figure out why toVersion doesn't work.
```
Page-, line 94: It would be good to give a description of what about it doesn't work, I've seen these sorts of messages enough time in the past and been completely unable to figure out what doesn't work but also unsure if it's a edge case that I'm just failing to find or if it's actually fixed

Page- (approved):

src/translations/v6/v6.ts
thgreasi: Opened a pinejs issue with all the details and referenced it in the comment.
See: https://github.com/balena-io/pinejs/issues/794

Page- (approved):

PR #1686, opened 2024-06-26
Use improved pine typings

otaviojacobi:


src/features/cascade-delete/hooks.ts
```diff
@@ -120,33 +118,33 @@ hooks.addPureHook('DELETE', 'resin', 'user', {
...
-				throw new errors.BadRequestError('Invalid user');
-			}
-			request.custom.actorId = user.actor.__id;
-			try {
-				await authApiTx.delete({
-					resource: 'api_key',
-					options: {
-						$filter: {
-							is_of__actor: user.actor.__id,
+			})
```
otaviojacobi, line 128: this is just github diff being slightly rdumb

src/infra/pinejs-client-helpers/index.ts
```diff
@@ -12,13 +12,13 @@ const $getOrInsertId = async (
...
 	const apiTx = api.clone({ passthrough: { req: permissions.root, tx } });
-	const results = (await apiTx.get({
+	const results = await apiTx.get({
 		resource,
 		options: {
 			$select: 'id',
 			$filter: body,
 		},
-	})) as Array<{ id: number }>;
+	} as const);
```
fisehara, line 21: Is this needed here as the resource name needs to be inferred and as it's a variable it needs to be `as const` first so that the type (name) of it is known before passing in further?

otaviojacobi: @fisehara I am wrapping the whole object but this is in fact needed because of the  `$select: ['id']` (e.g. we could have done just `$select: ['id'] as const,` and it would have worked fine. I added on the whole object as this is the guideline we decided to follow [here](https://github.com/balena-io/open-balena-api/pull/1670#discussion_r1653222705)

Now, the reasoning why we need this `as const` is because of [this line](https://github.com/balena-io-modules/pinejs-client-js/blob/master/src/index.ts#L33)  as the inference requires the selected field to be Readonly (e.g. const) and by default a created object is mutable

Page- (approved):

src/infra/pinejs-client-helpers/index.ts
Page-: I expect this is needed more because `resource` is a generic string and `filter` is `AnyObject`, basically this isn't using model based typings but just the generic ones (and generic ones require `as const` to get any meaningful response typings)

PR #1692, opened 2024-06-28
Add device update status/last update status event that devices can report

Page-:


src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -58,6 +67,13 @@ export type StatePatchV3Body = {
 				release_uuid?: string;
 				releases?: {
 					[releaseUUID: string]: {
+						update_status?:
+							| 'rejected'
+							| 'downloading'
+							| 'downloaded'
+							| 'applying changes'
+							| 'aborted'
+							| 'done';
```
joshbwlng, line 76:
```suggestion
						update_status?: NonNullable<Device['Write']['update_status']>;
```

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -22,6 +22,15 @@ import type { ResolveDeviceInfoCustomObject } from '../middleware.js';
 const { BadRequestError, UnauthorizedError, InternalRequestError } = errors;
 const { api } = sbvrUtils;
 
+enum updateStatusPredence {
```
joshbwlng, line 25: What does `Predence` mean?

Page-: It was meant to be precedence, fixed

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -319,6 +330,22 @@ export const statePatchV3: RequestHandler = async (req, res) => {
 							deviceBody.is_running__release = release.id;
 						}
 					}
+					const { releases } = apps[userAppUuid];
```
thgreasi, line 333: Since `userAppUuid = device.belongs_to__application[0].uuid` should we first check whether `apps[userAppUuid]` is not null before destructuring it?
```suggestion
					const releases = apps[userAppUuid]?.releases;
```

thgreasi (commented):
Other than that LGTM :+1:

src/features/device-state/routes/state-patch-v3.ts
Page-: So interestingly this change you proposed triggers an error
```
src/features/device-state/routes/state-patch-v3.ts:321:11 - error TS7022: 'device' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.
```
@thgreasi, I'm not sure why though

thgreasi (approved):

PR #1701, opened 2024-07-04
logs: Eagerly clear the heartbeat timeout when the request finishes

thgreasi:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/API.20pods.20in.20Error.2FSatusUnknown.20state/near/449102603

### DIFF
--- a/src/features/device-logs/lib/read.ts
+++ b/src/features/device-logs/lib/read.ts
@@ -115,20 +115,26 @@ async function handleStreamingRead(
 		}
 	});
 
+	let heartbeatTimeout: ReturnType<typeof setTimeout> | null = null;
+
 	function heartbeat() {
 		if (state !== StreamState.Closed) {
 			// In order to keep the connection alive, output new lines every now and then
 			write('\n');
-			setTimeout(heartbeat, LOGS_HEARTBEAT_INTERVAL);
+			heartbeatTimeout = setTimeout(heartbeat, LOGS_HEARTBEAT_INTERVAL);
 		}
 	}
 
-	setTimeout(heartbeat, LOGS_HEARTBEAT_INTERVAL);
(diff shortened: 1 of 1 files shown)


src/features/device-logs/lib/read.ts
```diff
@@ -115,21 +115,30 @@ async function handleStreamingRead(
...
-	function heartbeat() {
-		if (state !== StreamState.Closed) {
-			// In order to keep the connection alive, output new lines every now and then
-			write('\n');
-			setTimeout(heartbeat, LOGS_HEARTBEAT_INTERVAL);
-		}
-	}
-
-	setTimeout(heartbeat, LOGS_HEARTBEAT_INTERVAL);
+	let heartbeatInterval: ReturnType<typeof setInterval> | null = setInterval(
```
Page-, line 118: If you set the empty case as `undefined` then you can just keep `clearInterval(heartbeatInterval);` as it works accepts `undefined` (it just doesn't like `null`)
```suggestion
	let heartbeatInterval: ReturnType<typeof setInterval> | undefined = setInterval(
```

src/features/device-logs/lib/read.ts
```diff
@@ -115,21 +115,24 @@ async function handleStreamingRead(
...
 	});
 
-	function heartbeat() {
-		if (state !== StreamState.Closed) {
+	let heartbeatInterval: ReturnType<typeof setInterval> | undefined =
+		setInterval(function heartbeat() {
+			if (state === StreamState.Closed) {
+				clearInterval(heartbeatInterval);
+				heartbeatInterval = undefined;
+				return;
```
Page-, line 123:
```suggestion
				close();
				return;
```

Page- (approved):

PR #1704, opened 2024-07-05
Use improved pinejs-client prepare typings

Page-:
Change-type: patch

### DIFF
--- a/src/features/api-keys/lib.ts
+++ b/src/features/api-keys/lib.ts
@@ -1,10 +1,12 @@
 import type { Request } from 'express';
 import randomstring from 'randomstring';
 import _ from 'lodash';
+import type { Params } from 'pinejs-client-core';
 
 import { sbvrUtils, permissions, errors } from '@balena/pinejs';
 import { multiCacheMemoizee } from '../../infra/cache/index.js';
 import { API_KEY_ROLE_CACHE_TIMEOUT } from '../../lib/config.js';
+import type { Role } from '../../balena-model.js';
 
 const { api } = sbvrUtils;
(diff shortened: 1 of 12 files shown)


src/features/device-heartbeat/index.ts
```diff
@@ -43,16 +48,16 @@ const getPollIntervalForDevice = _.once(() =>
...
-);
-
-const getPollIntervalForParentApplication = _.once(() =>
-	api.resin.prepare<
-		{
-			deviceId: number;
-		},
-		'application_config_variable'
-	>({
+	} as const satisfies Params<DeviceConfigVariable>;
```
Page-, line 51: The `as const` in this pattern is necessary to make sure we get typings like `resource: 'device_config_variable'` instead of `resource: string` whilst the `satisfies Params<$resource>` means that it must be a valid `Params` object for that resource (since when passed to the `prepare` generic it doesn't actually enforce that for some reason)

src/features/api-keys/lib.ts
```diff
@@ -278,8 +280,8 @@ export const createGenericApiKey = async (
 };
 
 export const isApiKeyWithRole = (() => {
-	const authQuery = _.once(() =>
-		api.Auth.prepare<{ key: string; roleName: string }, 'role'>({
+	const authQuery = _.once(() => {
+		const opts = {
```
otaviojacobi, line 284: I would prefer this to be named `params` as it directly maps to the `Params<T>` typing and `opts` might be confusing with the inner `options`

otaviojacobi (approved):

PR #1715, opened 2024-07-16
Add the device.should_be_running__release computed term to the balena model

thgreasi:
```
device.should_be_running__release := device.is_pinned_on__release ?? application.should_be_running__release
```

Depends-on: https://github.com/balena-io/open-balena-api/pull/1717
Change-type: minor

### DIFF
--- a/src/features/devices/models/device-additions.ts
+++ b/src/features/devices/models/device-additions.ts
@@ -428,4 +428,40 @@ export const addToModel = (abstractSql: AbstractSqlModel) => {
 			],
 		],
 	});
+
+	const deviceShouldBeRunningReleaseField = abstractSql.tables[
+		'device'
+	].fields.find((f) => f.fieldName === 'should be running-release');
+	if (deviceShouldBeRunningReleaseField == null) {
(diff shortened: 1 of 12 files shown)


src/features/devices/models/device-additions.ts
```diff
@@ -428,4 +429,65 @@ export const addToModel = (abstractSql: AbstractSqlModel) => {
...
+							['ReferencedField', 'application', 'id'],
+							['ReferencedField', 'device', 'belongs to-application'],
+						],
+					],
+				],
+			],
+		],
+	});
+
+	abstractSql.relationships = _.merge(abstractSql.relationships, {
```
Page-, line 470: Would it make more sense to generate these via having it in the sbvr and then converting it to computed in the abstract sql model?

thgreasi: I was also thinking about that while writing this...
Didn't try it primarily b/c we have never done that before for any other computed term on the balena model, but yes it totally makes sense and works :+1:

Page- (approved):

PR #1717, opened 2024-07-17
is_pinned_on__release: drop device "should_be_running__release" column if exist

JSReds:
is_pinned_on__release: remove device "should_be_running__release" index and column if they exist

Project: https://balena.fibery.io/Work/Project/Server-side-pagination-for-devices-Cycle-3-539#Project/Server-side-pagination-for-devices-Cycle-4-673

Change-type: patch

### DIFF
--- a/src/migrations/00091-drop-device-should-be-running-release-column.sql
+++ b/src/migrations/00091-drop-device-should-be-running-release-column.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "device"
+DROP COLUMN IF EXISTS "should be running-release";


src/migrations/00091-drop-device-should-be-running-release-column.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "device"
+DROP COLUMN IF EXISTS "should be running-release";
+DROP INDEX IF EXISTS "device_should_be_running_release_application_idx";
```
thgreasi, line 3: I think that we do not really need this, since dropping the column will drop the index as well.

thgreasi (approved):

PR #1725, opened 2024-08-06
vars-schema: Add custom kernel device-tree support for AGX Orin 64GB

acostach:
Although this is a new device-type, its' slug matches the existing SV rule for the jetson-agx-orin-devkit and needs to other changes apart from the one added in this PR.

Change-type: minor


acostach:
hey @thgreasi , although it's a small change it will need an approval since it's in balena-io

thgreasi (approved):

PR #1730, opened 2024-08-08
Update @balena/pinejs to v18

joshbwlng:
Change-type: patch

---

Main thing to review is the change in http error codes. Pine now defaults to `500` on unknown errors instead of `404` so the changes in this PR are also meant to make sure we are still returning expected/correct error codes.


src/features/applications/hooks/slug.ts
```diff
@@ -60,16 +64,26 @@ hooks.addPureHook('PATCH', 'resin', 'application', {
...
+			const appSlugs = new Map<number, string>();
+			try {
+				apps.forEach((app) => {
+					appSlugs.set(
+						app.id,
+						getApplicationSlug(app.organization[0].handle, app.app_name),
+					);
+				});
+			} catch (e) {
+				throw new errors.NotFoundError(e.message);
```
thgreasi, line 77: This might be a legitimate 500 actually, since we are passing in the handle & the app_name that we retrieved from the DB, and there shouldn't be a way for them to be nullable.
So I think we can revert this part.

src/features/applications/hooks/slug.ts
```diff
@@ -17,10 +17,14 @@ hooks.addPureHook('POST', 'resin', 'application', {
...
-					organization.handle,
-					request.values.app_name,
-				);
+				try {
+					request.values.slug = getApplicationSlug(
+						organization.handle,
+						request.values.app_name,
+					);
+				} catch (e) {
+					throw new errors.NotFoundError(e.message);
```
thgreasi, line 26: :+1: 
I saw that we don't have a test case for this, not even in the SDK...
I'm fine leaving it as a 404 to preserve the current behavior, but let's add comment that this was done for backwards compatiblity.
On the other hand, I would be fine even if you changed this to a bad request, if you also had thought about doing so, since I'm not expecting that there would be users depending on the 404 status, when they pass in a null app_name.

src/features/applications/hooks/slug.ts
```diff
@@ -17,10 +17,14 @@ hooks.addPureHook('POST', 'resin', 'application', {
 				},
 			});
 			if (organization != null) {
-				request.values.slug = getApplicationSlug(
-					organization.handle,
-					request.values.app_name,
-				);
+				try {
+					request.values.slug = getApplicationSlug(
```
thgreasi, line 21: What do you think about adding a explicit, instead of relying on the try-catch?
```
	if (request.values.app_name == null) {
		throw new BadRequestError/NotFoundError('Missing application name');
	}
```
We can then leave the `throw new Error()`s as is in `getApplicationSlug` as they are, to signal w/ a 500 that we haven't checked the parameters in one of our codepaths.

src/features/applications/hooks/slug.ts
joshbwlng: Won't revert it, but will make it an explicit `errors.InternalServerError(e.message);`

src/features/applications/index.ts
```diff
@@ -1,13 +1,15 @@
+import { errors } from '@balena/pinejs';
+
 export const getApplicationSlug = (
 	orgHandle: string,
 	appName: string,
 ): string => {
 	if (orgHandle == null) {
-		throw new Error('Missing organization handle');
+		throw new errors.BadRequestError('Missing organization handle');
```
thgreasi, line 8: These are obviously behavior changes, but I do agree that it is for the better and don't expect anyone to be depending on this returning a 404.

thgreasi (approved):
Marked it as draft so that we can merge it once bC is also reviewed so that we don't block the pipeline :+1:

PR #1738, opened 2024-08-27
Fix v6 image.image_size no longer being a number

thgreasi:
Change-type: patch
See: https://github.com/balena-io/open-balena-api/pull/1730
See: https://github.com/balena-io-modules/sbvr-types/pull/106
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/v6.20image.2Eimage_size.20is.20now.20a.20string/near/465509436
See: https://balena.fibery.io/Work/Project/Fix-image-delta-size-no-longer-being-a-number-in-versions-v6-810

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -265,6 +265,10 @@ export const envVarsConfig = {
 	DEVICE_TYPE_SPECIFIC_CONFIG_VAR_PROPERTIES,
 };
 
+// Needed so that the augmented `@balena/sbvr-types` typings
+// automatically become available to consumer projects.
+import './translations/v6/numeric-big-integer-hack.js';
+
 export const translations = {
 	v7: {
 		getTranslations: getV7Translations,
--- a/src/translations/v6/numeric-big-integer-hack.ts
+++ b/src/translations/v6/numeric-big-integer-hack.ts
@@ -1,0 +1,31 @@
+import SbvrTypes from '@balena/sbvr-types';
+import type * as TypeUtils from '@balena/sbvr-types/out/type-utils.js';
+
+// We use the TsTypes as the type name here so that it doesn't conflict later
+// when we augment the @balena/sbvr-types Types type.
+type TsTypes = TypeUtils.TsTypes<number, number>;
+export { TsTypes as Types };
+export type DbWriteType = number;
+
+// Augmenting is necessary so that the TS v6-model.ts compiles properly
+declare module '@balena/sbvr-types' {
+	export interface Types {
+		'Numeric Big Integer': TsTypes;
+	}
+}
+
+// @ts-expect-error we are augmenting SbvrTypes w/ Numeric Big Integer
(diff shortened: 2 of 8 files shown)


src/index.ts
```diff
@@ -265,6 +265,17 @@ export const envVarsConfig = {
...
 };
 
+// Needs to be `import *` so that the import path is preserved in the generated .d.ts
+import * as numericBigInteger from './translations/v6/numeric-big-integer-hack.js';
+/**
+ * This automatically augments the declarations of the consumer projects.
+ */
+export const augmentedTypes = {
+	'@balena/sbvr-types': {
+		'Numeric Big Integer': numericBigInteger,
```
thgreasi, line 275: This allows consumers to have their v6 model typings work without having to manually import the numeric-big-integer-hack (or redeclare the augmented typed).
We can perhaps in the future can start adding more typings here which we atm have only in oB, like balena-device-config.ts or even common.ts

src/translations/v6/numeric-big-integer-hack.ts
```diff
@@ -0,0 +1,34 @@
...
+		if (data == null || typeof data === 'number') {
+			return data;
+		}
+		if (typeof data === 'string') {
+			return parseInt(data, 10);
+		}
+		throw new Error(
+			'Fetched Numeric Big Integer value is not valid: ' + typeof data,
+		);
+	},
```
Page-, line 31: I would lean towards simplifying this function by reusing as much of the `Big Integer` processing as possible:
```suggestion
	fetchProcessing(data) {
		const processedData = SbvrTypes['Big Integer'].fetchProcessing(data);
		return parseInt(processedData, 10);
	},
```
Performance should be equivalent since the `Big Integer` processing should get a string and then directly return it for us but it means less custom code/clearer intention (ie "big integer, but as a number")

src/translations/v6/v6.sbvr
```diff
@@ -140,7 +140,7 @@ Term: image location
 	Concept Type: Short Text (Type)
 
 Term: image size
-	Concept Type: Big Integer (Type)
+	Concept Type: Integer (Type)
```
Page-, line 143: Any reason not to leave this as a big integer type here? It shouldn't affect anything other than knowing this is actually a big integer in the db rather than an integer when looking at it
```suggestion
	Concept Type: Big Integer (Type)
```

src/index.ts
```diff
@@ -265,6 +265,17 @@ export const envVarsConfig = {
...
+	},
+};
```
Page-, line 277: You should be able to do this with just:
```suggestion
import './translations/v6/numeric-big-integer-hack.js';
```

thgreasi: Loved it!

Page- (commented):
Looks good but I'll wait to approve as I expect that merging atm would be a bit of a pain

Page- (approved):

PR #1747, opened 2024-09-04
Use `should_be_running__release` to handle the device/app release cascade for state-get

Page-:


thgreasi:
I would prefer it if we just deprecated the `getReleaseForDevice` and added a `TODO` to drop it in the next major, rather than doing a major just for this.

thgreasi (approved):

PR #1750, opened 2024-09-06
Update dependencies

Page-:
Update @balena/pinejs from 18.2.1 to 18.2.3

Change-type: patch


src/balena-model.ts
```diff
@@ -736,11 +736,6 @@ export interface Device {
 			| [Release['Read']]
 			| []
 			| null;
-		should_be_running__release:
```
thgreasi, line 739: ???? I don't think that should have gone away

Page-: It's a computed field, it cannot be written to

Page-: Oh wait, it's also affected read typings 🤦

src/balena-model.ts
```diff
@@ -1076,10 +1071,6 @@ export interface Release {
 		known_issue_list: Types['Text']['Write'] | null;
 		note: Types['Text']['Write'] | null;
 		invalidation_reason: Types['Text']['Write'] | null;
-		is_final: Types['Boolean']['Write'];
-		semver: Types['Short Text']['Write'];
```
thgreasi, line 1080: Just noting that these two are actually write-able on the OData API level (handled in hooks), but of course the model can't know that and we would need to augment them manually if needed.

thgreasi (approved):

PR #1761, opened 2024-09-16
Improve getTargetStateV2 typings

otaviojacobi:
Change-type: patch

### DIFF
--- a/src/features/device-state/routes/state-get-v2.ts
+++ b/src/features/device-state/routes/state-get-v2.ts
@@ -24,6 +24,8 @@ const { api } = sbvrUtils;
 
 type CompositionService = AnyObject;
 type LocalStateApp = StateV2['local']['apps'][string];
+type ExpandedDevice = Awaited<ReturnType<typeof getDevice>>;
+
 export type StateV2 = {
 	local: {
 		name: string;
@@ -63,9 +65,9 @@ export type StateV2 = {
 };
 
 function buildAppFromRelease(
-	device: AnyObject,
-	application: AnyObject,
-	release: AnyObject,
+	device: ExpandedDevice,
+	application: ExpandedDevice['belongs_to__application'][number],
+	release: ExpandedDevice['should_be_running__release'][number],
 	config: Dictionary<string>,
 ): LocalStateApp {
 	let composition: AnyObject = {};
@@ -82,7 +84,7 @@ function buildAppFromRelease(
 		}
 	}
 
-	for (const ipr of release.contains__image as AnyObject[]) {
+	for (const ipr of release.contains__image) {
 		// extract the per-image information
 		const image = ipr.image[0];
 
@@ -90,7 +92,7 @@ function buildAppFromRelease(
 		if (si == null) {
 			throw new Error(
 				`Could not find service install for device: '${
-					application.uuid
+					application.id
 				}', image: '${image?.id}', service: '${JSON.stringify(
 					image?.is_a_build_of__service,
 				)}', service installs: '${JSON.stringify(device.service_install)}'`,
@@ -109,7 +111,7 @@ function buildAppFromRelease(
 		for (const { label_name, value } of [
 			...ipr.image_label,
 			...svc.service_label,
-		] as Array<{ label_name: string; value: string }>) {
+		]) {
 			labels[label_name] = value;
 		}
 
@@ -253,7 +255,7 @@ const getStateV2 = async (req: Request, uuid: string): Promise<StateV2> => {
 	});
 
 	const userApp = getUserAppForState(device, config);
-	const userAppFromApi: AnyObject = device.belongs_to__application[0];
+	const userAppFromApi = device.belongs_to__application[0];
 
 	const local: StateV2['local'] = {
 		name: device.device_name!,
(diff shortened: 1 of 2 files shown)


src/features/device-state/routes/state-get-v2.ts
```diff
@@ -82,15 +84,15 @@ function buildAppFromRelease(
...
+	for (const ipr of release.contains__image) {
 		// extract the per-image information
 		const image = ipr.image[0];
 
 		const si = serviceInstallFromImage(device, image);
 		if (si == null) {
 			throw new Error(
 				`Could not find service install for device: '${
-					application.uuid
+					application.id
```
otaviojacobi, line 95: This should be the only (tiny) runtime change: It is a mistake captured by the typings, the `uuid` from application was not previously `$select`ed

src/features/device-state/state-get-utils.ts
```diff
@@ -65,10 +72,44 @@ export const getConfig = (
...
+): ExpandedServiceInstall | undefined;
+
+export function serviceInstallFromImage(
+	deviceOrFleet: { service: ExpandedService[] },
+	image?: {
+		is_a_build_of__service: number | { __id: number };
+	},
+): ExpandedService | undefined;
+
+export function serviceInstallFromImage(
```
otaviojacobi, line 100: I kept the `AnyObject` overload to preserve backwards compability of the typings as there are some (possible) type mistakes on get state v3 that are not happy with these typings (which should be correct).

src/features/device-state/state-get-utils.ts
```diff
@@ -39,8 +39,15 @@ export const setDefaultConfigVariables = (config: Dictionary<string>): void => {
 };
 
 export const getConfig = (
-	device: AnyObject | undefined,
-	application: AnyObject = device?.belongs_to__application[0],
+	device:
```
otaviojacobi, line 42: Theoretically this could be considered a breaking change as it is narrowing the types, however, it is narrowing to a more correct typing (e.g. if you were passing something that does not satisfy these typings, you would probably be running into runtime errors), so I decided to keep this in a patch

src/features/device-state/state-get-utils.ts
```diff
@@ -86,11 +127,9 @@ export const serviceInstallFromImage = (
 			(si) => si.service[0].id === id,
 		);
 	} else if ('service' in deviceOrFleet) {
-		return deviceOrFleet.service.find(
```
otaviojacobi, line 89: This is also slightly missleading as the fn is called `serviceInstallFromImage` but can return a service (rather than a serviceInstall). I decided not to change the function name for now as it would be a major and this is used on balena-api

src/features/device-state/state-get-utils.ts
Page-: Yeah, imo patch is fine for things like this where these properties were always required but were just not typed narrowly enough to cover the actual requirements. It's part of the progressive typing we have where `AnyObject` is purely better than `any` (guards against more mistakes) and so we use that to avoid potential "perfect being the enemy of good" issues but it's even more better to do this specific typing

src/features/device-state/state-get-utils.ts
Page-, line 100: Could you remove the empty lines between overloads here please, with the empty line it implies to me they are separate functions rather than a single (overloaded) function
```suggestion
export function serviceInstallFromImage(
	deviceOrFleet: { service_install: ExpandedServiceInstall[] },
	image?: {
		is_a_build_of__service: number | { __id: number };
	},
): ExpandedServiceInstall | undefined;
export function serviceInstallFromImage(
	deviceOrFleet: { service: ExpandedService[] },
	image?: {
		is_a_build_of__service: number | { __id: number };
	},
): ExpandedService | undefined;
export function serviceInstallFromImage(
```

Page-: In that case can you add a `/** @deprecated **/` comment so it'll show in editors when this overload is being used and make it easier to phase it out

src/features/device-state/state-get-utils.ts
```diff
@@ -65,10 +72,42 @@ export const getConfig = (
...
+		is_a_build_of__service: number | { __id: number };
+	},
+): ExpandedServiceInstall | undefined;
+export function serviceInstallFromImage(
+	deviceOrFleet: { service: ExpandedService[] },
+	image?: {
+		is_a_build_of__service: number | { __id: number };
+	},
+): ExpandedService | undefined;
+/** @deprecated Use the strongly-typed overloads with `{ service_install }` or `{ service }` instead */
```
Page-, line 98: Is this actually still needed now that the overloads cover more/different cases?

otaviojacobi: It is if we want to keep runtime code (in state v3) as is. The reason is in short this line: https://github.com/balena-io/open-balena-api/blob/master/src/features/device-state/routes/state-get-v3.ts#L108

Even if we overload this function so it knows that when device is undefined application will have the a correct "service" typescript is not able to to infer the proper type with a ?? so this line would have to become something like

```typescript
const si = device != null ? serviceInstallFromImage(device, image) : serviceInstallFromImage(application, image);
```

Or even a better approach for narrower type would be to on top level separate a path where device is null and one where it isn't, however, I am slightly scared about how this would affect performance (but would for sure allow better typings). Right now what I was trying to do was bend the typings to the current code instead of the code to the force correct typings.

otaviojacobi: So I would say at least for now we should keep it and then maybe we can improve when we are actually touching state-get-V3.

Page- (approved):

PR #1762, opened 2024-09-16
Reduce usage of service_install form state get v2

otaviojacobi:


src/features/device-state/routes/state-get-v2.ts
```diff
@@ -98,7 +99,7 @@ function buildAppFromRelease(
 				)}', service installs: '${JSON.stringify(device.service_install)}'`,
 			);
 		}
-		const svc = si.service[0];
+		const svc = image.is_a_build_of__service[0];
```
Page-, line 102: If you move this further up then you can change the error from
```ts
			throw new Error(
				`Could not find service install for device: '${
					application.id
				}', image: '${image?.id}', service: '${JSON.stringify(
					image?.is_a_build_of__service,
				)}', service installs: '${JSON.stringify(device.service_install)}'`,
			);
```
to
```ts
			throw new Error(
				`Could not find service install for device: '${
					application.id
				}', image: '${image?.id}', service: '${svc}', service installs: '${JSON.stringify(device.service_install)}'`,
			);

otaviojacobi: This slightly change the error message from `JSON.stringify([service])`  to basically `JSON.stringify(service)` e.g. it removes the wrapping list, which should always have a single elemtn, so I think we are fine with it

otaviojacobi: FWIW, `image?.id`  <- the `?` should not be necessary (at least the typings say that)

Page- (approved):

PR #1763, opened 2024-09-17
Improve getDeviceTargetStateV3/getFleetTargetStateV3 typings

otaviojacobi:
Change-type: patch

### DIFF
--- a/src/features/device-state/routes/fleet-state-get-v3.ts
+++ b/src/features/device-state/routes/fleet-state-get-v3.ts
@@ -9,8 +9,10 @@ import {
 
 import { sbvrUtils, errors } from '@balena/pinejs';
 import { getConfig, readTransaction } from '../state-get-utils.js';
-import type { StateV3 } from './state-get-v3.js';
+import type { ExpandedRelease, StateV3 } from './state-get-v3.js';
 import { buildAppFromRelease, releaseExpand } from './state-get-v3.js';
+import type { OptionsToResponse } from 'pinejs-client-core';
+import type { Application } from '../../../balena-model.js';
 const { api } = sbvrUtils;
 const { UnauthorizedError } = errors;
 
@@ -45,6 +47,16 @@ const fleetExpand = {
 		},
 	},
 } as const;
+
+export type ExpandedApplicationWithService = NonNullable<
+	OptionsToResponse<
+		Application['Read'],
+		{
+			$expand: typeof fleetExpand;
+		},
+		number
+	>
+>;
 
 const stateQuery = _.once(() =>
 	api.resin.prepare(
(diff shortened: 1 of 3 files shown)


src/features/device-state/routes/fleet-state-get-v3.ts
```diff
@@ -46,6 +48,14 @@ const fleetExpand = {
 	},
 } as const;
 
+export type ExpandedApplicationWithService = OptionsToResponse<
+	Application['Read'],
+	{
+		$expand: typeof fleetExpand;
+	},
+	undefined
+>[number];
```
Page-, line 57: I think you can also just pass a type for the id to show that you will be getting a specific one rather than the array
```suggestion
export type ExpandedApplicationWithService = OptionsToResponse<
	Application['Read'],
	{
		$expand: typeof fleetExpand;
	},
	number
>;
```

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -19,11 +19,31 @@ import { sbvrUtils } from '@balena/pinejs';
...
 type ServiceComposition = AnyObject;
+type ExpandedDevice = Awaited<ReturnType<typeof getDevice>>;
+type ExpandedApplication = OptionsToResponse<
+	Application['Read'],
+	{
+		$select: typeof appSelect;
+		$expand: typeof appExpand;
+	},
+	undefined
+>[number];
```
Page-, line 38: Same here:
```suggestion
type ExpandedApplication = OptionsToResponse<
	Application['Read'],
	{
		$select: typeof appSelect;
		$expand: typeof appExpand;
	},
	number
>;
```

src/features/device-state/routes/fleet-state-get-v3.ts
otaviojacobi: @Page- These are not the same, from my testing
```ts
export type ExpandedApplicationWithService = OptionsToResponse<
	Application['Read'],
	{
		$expand: typeof fleetExpand;
	},
	number
>;
```

Will be the same as my version above `| undefined` and my understanding is that we don't want it to be nullable. Ofc I can just wrap it in a `NonNullable` but I am not 100% sure which one is best considering this

src/features/device-state/state-get-utils.ts
```diff
@@ -95,6 +95,14 @@ export function serviceInstallFromImage(
...
+export function serviceInstallFromImage(
+	deviceOrFleet:
+		| { service_install: ExpandedServiceInstall[] }
+		| { service: ExpandedService[] },
+	image?: {
+		is_a_build_of__service: number | { __id: number };
+	},
+): ExpandedServiceInstall | ExpandedService | undefined;
 /** @deprecated Use the strongly-typed overloads with `{ service_install }` or `{ service }` instead */
 export function serviceInstallFromImage(
```
otaviojacobi, line 107: Still keeping the `AnyObject` overload not to break balena-api (for now), but the follow up might remove it

Page- (approved):

PR #1765, opened 2024-09-17
Add test cases for different environment_variables scopes in state get

otaviojacobi:
https://github.com/balena-io/open-balena-api/pull/1762 should use a different pass (not via service_install) to find the `service` env vars and service labels. We currently have no tests for the correct behaviour of env var overwriting. 

Change-type: patch

### DIFF
--- a/test/test-lib/fixtures.ts
+++ b/test/test-lib/fixtures.ts
@@ -13,6 +13,7 @@ import { requestAsync } from '../../src/infra/request-promise/index.js';
 import { supertest } from './supertest.js';
 import type { Organization } from '../../src/balena-model.js';
 import type Model from '../../src/balena-model.js';
+import { assertExists } from './common.js';
 
 const { api } = sbvrUtils;
 const version = 'resin';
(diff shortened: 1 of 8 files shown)


test/test-lib/fixtures.ts
```diff
@@ -448,6 +448,29 @@ const loaders: types.Dictionary<LoaderFunc> = {
 			user,
 		});
 	},
+	service_installs: async (jsonData, fixtures) => {
```
Page-, line 451: Is this actually necessary? Service installs should always be automatically created whenever they're necessary and if they're not then I'd expect that's a bug that needs to be fixed

otaviojacobi: @Page-  I think we might be able to have the service_install be created automatically but I think something in either the ordering of the mocks or maybe some missing property I could not figure it out yet causes it not to be created. I see that this is tested in cloud api more or less the same way so I followed from here: https://github.com/balena-io/balena-api/blob/master/test/20_state_endpoint.ts#L544-L553

test/test-lib/fixtures.ts
```diff
@@ -471,6 +471,60 @@ const loaders: types.Dictionary<LoaderFunc> = {
...
+			logErrorAndThrow(`Could not find service: ${jsonData.service}`);
+		}
+
+		const {
+			d: [si],
+		} = await requestResource({
+			resource: `service_install(device=${device.id},installs__service=${service.id})`,
+			method: 'GET',
+			user,
+		});
```
Page-, line 516: It makes more sense imo to just fetch this with pinejs-client, eg
```suggestion
		const si = await api.resin.get({
			resource: 'service_install'
			passthrough: { req: permissions.root },
			id: {
				device: device.id,
				installs__service: service.id,
			},
		});
```
rather than the weird resource with embedded params.

Page- (approved):

PR #1776, opened 2024-09-30
WIP

otaviojacobi:
Change-type: patch


src/lib/auth.ts
```diff
@@ -40,39 +63,141 @@ export const ROLES: {
...
-		`resin.user__has__public_key.all?${matchesUser}`,
-		'resin.release_asset.all',
+		// 'resin.actor.delete?id eq @__ACTOR_ID',
+		resinAuth({
+			resource: 'actor',
+			access: 'delete',
+			options: {
+				$filter: {
+					id: {
+						'@': '__ACTOR_ID',
```
Page-, line 73: I would definitely separate this out as a reusable var, eg `actorId` or something as it would help a bit with deduplication and since it is a constant that has to match precisely there's no sense in retyping it each time

src/lib/auth.ts
```diff
@@ -40,39 +63,141 @@ export const ROLES: {
...
+				$filter: {
+					is_of__actor: {
+						'@': '__ACTOR_ID',
+					},
+				},
+			},
+		}),
+
+		// 'resin.application.all',
+		resinAuth({ resource: 'application', access: 'all' }),
```
Page-, line 93: Things are fairly verbose like this, I think it's better though and worth adding the verbosity but I do wonder if in future we'll add shortcuts to reduce the verbosity slightly whilst keeping the benefits. For now I think I'd say live with the minor pain for the greater good it brings

otaviojacobi: I this this can be easily reduced by making the parameters on the return warpper 3 values e.g.
`balenaAuth('application', 'all')`

PR #1785, opened 2024-10-08
Add the ability to enable/disable scheduled jobs

Page-:
Change-type: minor

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -9,6 +9,7 @@ import {
 	SECONDS,
 	hostPortsVar,
 	trustProxyVar,
+	arrayVar,
 } from '@balena/env-parsing';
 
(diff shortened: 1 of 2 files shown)


src/lib/config.ts
```diff
@@ -486,6 +487,16 @@ export const WEBRESOURCES_CLOUDFRONT_HOST = optionalVar(
 	'WEBRESOURCES_CLOUDFRONT_HOST',
 );
 
+export const DISABLED_SCHEDULED_JOBS = new Set(
+	arrayVar('DISABLED_SCHEDULED_JOBS') ?? [],
+);
+export const disableScheduledJob = (jobName: string) => {
```
thgreasi, line 493: I guess it would be useful to also have a way to disable all the jobs, but we can do that later if needed :+1:

thgreasi (approved):

PR #1793, opened 2024-10-16
Multi level store

Page-:
These are various improvements aiming to improve the use of the multi level store and also to avoid needing to repopulate caches from scratcg on a version bump when that is not necessary

### DIFF
--- a/src/features/device-state/middleware.ts
+++ b/src/features/device-state/middleware.ts
@@ -25,7 +25,9 @@ const checkDeviceExistsIsFrozenQuery = _.once(() =>
 	),
 );
 export const checkDeviceExistsIsFrozen = multiCacheMemoizee(
-	async (uuid: string) => {
+	async (
+		uuid: string,
+	): Promise<Pick<Device['Read'], 'id' | 'is_frozen'> | undefined> => {
 		return await checkDeviceExistsIsFrozenQuery()({ uuid });
 	},
 	{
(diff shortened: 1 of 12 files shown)


src/features/device-state/middleware.ts
```diff
@@ -35,6 +35,7 @@ export const checkDeviceExistsIsFrozen = multiCacheMemoizee(
 		undefinedAs: false,
 		maxAge: DEVICE_EXISTS_CACHE_TIMEOUT,
 	},
+	{ useVersion: false },
```
thgreasi, line 38: A comment above the query that in case we change the query we need to change the `useVersion` to `true` would be nice.

src/features/device-state/state-patch-utils.ts
```diff
@@ -142,18 +144,15 @@ export type ImageInstallUpdateBody = {
 const shouldUpdateImageInstall = (() => {
 	const lastImageInstallReport = createMultiLevelStore<
 		ImageInstallUpdateBody & { updateTime: number }
```
thgreasi, line 146: A comment that if the structure of the content changes, then we also need to mark it as `useVersion: true` would have been nice to have.

src/features/device-state/middleware.ts
Page-: I've added an explicit return type to the memoized function so that if the query changes without updating the memoized fn then typescript will raise it

thgreasi (approved):

PR #1794, opened 2024-10-17
Skip triggering device update checks for image env var POSTs of non-successful releases

thgreasi:
Change-type: patch


src/features/vars-schema/hooks/vars-update-trigger.ts
```diff
@@ -318,6 +318,39 @@ addEnvHooks('device_service_environment_variable', async (args) => {
 
 addEnvHooks('image_environment_variable', async (args) => {
 	if (args.req.body.release_image != null) {
+		// This is a optimization that allows skipping the following
+		// device->image_install->...->release_image query in most real world
+		// cases, since when user permissions are used it creates a complex query
+		// with high planning time. The idea is that image env vars are created
+		// while the release is still 'running', in which case we don't expect
+		// any device to possibly be running that release.
+		const isPartOfSuccessfulRelease =
```
thgreasi, line 327: Someone could even argue that we could just never check for affected devices on image env var POSTs at all,
since if someone does end up doing image env var POSTs after the release gets finalized (which we don't expect), then the devices will anyway fetch them on their next scheduled state GET poll.

thgreasi:
#1796 might be a better approach

thgreasi:
Closing in favor of #1796

PR #1796, opened 2024-10-17
var-update-trigger: Update only devices that should be running the release that the image env var is part of

thgreasi:
…

Change-type: patch


src/features/vars-schema/hooks/vars-update-trigger.ts
```diff
@@ -316,24 +316,25 @@ addEnvHooks('device_service_environment_variable', async (args) => {
...
 		return {
-			image_install: {
+			should_be_running__release: {
 				$any: {
-					$alias: 'ii',
+					$alias: 'r',
 					$expr: {
-						installs__image: {
+						status: 'success',
+						release_image: {
```
Page-, line 332: Can you use `contains_image` to make this match the state get endpoint, and add a comment to the top of this file that the having the queries match the path the state get endpoint uses to get env vars for a device is the way to go. Essentially this captures if a subset of the state-get query will change so the correct thing is to follow the same path (and also means it'll have the same performance characteristics, which we take real care to optimize for in the state-get)

thgreasi:
Recheked this query and it seems that this way it becomes as slow as the service_environment_variables one (3+ seconds)...
The mistake was that I used a release id in place of a release_image id in my profiled queries...

thgreasi:
Marking as draft for now since we shouldn't get it in given that it regresses performance.
Should we reconsider https://github.com/balena-io/open-balena-api/pull/1794  ?

PR #1806, opened 2024-10-21
Reduce time transaction is open for `getWriteContext`

Page-:
Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/store.ts
+++ b/src/features/device-logs/lib/store.ts
@@ -52,27 +52,27 @@ const getWriteContext = (() => {
 			uuid: string,
 			req: permissions.PermissionReq,
 		): Promise<false | LogContext> => {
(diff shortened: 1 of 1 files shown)


thgreasi (approved):

src/features/device-logs/lib/store.ts
```diff
@@ -52,27 +52,27 @@ const getWriteContext = (() => {
 			uuid: string,
 			req: permissions.PermissionReq,
 		): Promise<false | LogContext> => {
-			return await sbvrUtils.db.readTransaction(async (tx) => {
-				try {
-					const result = await api.resin.request({
+			try {
+				const result = await sbvrUtils.db.readTransaction(async (tx) => {
```
thgreasi, line 56: I guess we don't even need to open the TX manually (just checked that a POST to canAccess creates a readTx anyway)

PR #1816, opened 2024-10-23
Add `LOKI_READ_PCT` to enable reading logs from loki

Page-:


src/lib/config.ts
```diff
@@ -305,6 +305,12 @@ export const LOKI_WRITE_PORT = intVar('LOKI_WRITE_PORT', LOKI_PORT);
 export const LOKI_READ_PORT = intVar('LOKI_READ_PORT', LOKI_PORT);
 // control the percent of logs written to Loki while scaling up
 export const LOKI_WRITE_PCT = intVar('LOKI_WRITE_PCT', 0);
+/**
+ * This is the percent of log read requests that will go to loki, however the number of logs fetched from loki
+ * will vary based upon the type of those read requests, eg it could be a long streaming request or a one-off fetch
+ * Note: It is recommended to only set this if `LOKI_WRITE_PCT` is at 100
```
thgreasi, line 311: Maybe it would make sense to actually throw in case LOKI_WRITE_PCT is not 100, b/c otherwise customers would be missing logs in an inconsistent way :thinking:

thgreasi (approved):
Converted to draft to give you the chance to check the comment, but LGTM even as is.

otaviojacobi (approved):

PR #1826, opened 2024-10-30
Loki

Page-:


otaviojacobi (approved):

src/features/device-logs/lib/store.ts
```diff
@@ -132,6 +132,8 @@ function handleStoreErrors(req: Request, res: Response, err: Error) {
 	res.status(500).end();
 }
 
+const lokiBackend = LOKI_ENABLED ? await getLokiBackend() : undefined;
```
thgreasi, line 135: I hope the top level await would not cause any unexpected issues :grimacing:

thgreasi (dismissed):
~~Since `getLokiBackend` is already memoized with an `once`, I don't expect this to have any measurable impact but lgtm.~~

nvm, I just read the commit message...

PR #1837, opened 2024-11-08
device-types: Sync only device-types with DB device_type records

### LINKED ISSUE #1060: Should drop the /device-types/v1/:deviceType/images endpoint after 2022-12-23
[thgreasi] See: https://jel.ly.fish/milestone-drop-device-types-v1-devicetype-images-9efbc48
See: https://monitor.balena-cloud.com/explore?orgId=1&left=%7B%22datasource%22:%22000000002%22,%22queries%22:%5B%7B%22expr%22:%22%7Bname%3D%5C%22balena-api%5C%22%7D%20%5Cn%7C~%20%5C%22GET%20%2Fdevice-types%2Fv1%2F%28%5B%5E%20%5D%29%2B%2Fimages%20%5C%22%20%5Cn%7C%20pattern%20%60%3C_%3E%20%3C_%3E:%20%3C_%3E%20%3C_%3E%20%3C_%3E%20%3Cmethod%3E%20%3Curi%3E%20%3CstatusCode%3E%20%3Clatency%3Ems%20%3C_%3E%60%22,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%22now-7d%22,%22to%22:%22now%22%7D%7D

### LINKED ISSUE #1433: Failed to get device type build data for...
_(moving https://github.com/balena-io/open-balena/issues/176 to here)_

Attempting to start an open-balena instance results in open-balena-api having hundreds of errors during startup that like:
```
Failed to get device type build data for imx8mmebcrs08a1/2.103.3 Access Denied AccessDenied: Access Denied
    at Request.extractError (/usr/src/app/node_modules/aws-sdk/lib/services/s3.js:711:35)
    at Request.callListeners (/usr/src/app/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/usr/src/app/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/usr/src/app/node_modules/aws-sdk/lib/request.js:686:14)
    at Request.transition (/usr/src/app/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/usr/src/app/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /usr/src/app/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/usr/src/app/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/usr/src/app/node_modules/aws-sdk/lib/request.js:688:12)
    at Request.callListeners (/usr/src/app/node_modules/aws-sdk/lib/sequential_executor.js:116:18)
    at Request.emit (/usr/src/app/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/usr/src/app/node_modules/aws-sdk/lib/request.js:686:14)
    at Request.transition (/usr/src/app/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/usr/src/app/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /usr/src/app/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/usr/src/app/node_modules/aws-sdk/lib/request.js:38:9)
```

As detailed in https://github.com/balena-io/balena-cli/issues/2322 and https://github.com/balena-io/open-balena/issues/129 by @pdcastro, open-balena points to balena-cloud's S3 bucket full of endpoint images.

At startup, open-belena-api grabs https://resin-production-img-cloudformation.s3.amazonaws.com/ to get a list of all available device types - this succeeds. open-balena-api then attempts to fetch `device-type.json` under each key. S3 returns `AccessDenied` for every one of those requests when `IMAGE_STORAGE_PREFIX` is set to `images` - which is the default at https://github.com/balena-io/open-balena/blob/ea07d43c5c7e98bb8b11c808b12428a46019c8d5/compose/services.yml#L34

If I change `IMAGE_STORAGE_PREFIX` to `esr-images`, then open-balena-api starts up without all the access denied errors ... but I'm not sure if that is giving me the right set of images or not.

---

For anyone else following along, `esr-images` contains the "Extended Support Release" images [as detailed here](https://blog.balena.io/balenaos-esr-peace-of-mind-with-only-two-updates-a-year/) and is a very limited set of devices:
- beaglebone-black
- ccon-01
- fincm3
- genericx86-64-ext
- intel-nuc
- jetson-tx2
- raspberry-pi
- raspberrypi3-64
- raspberrypi3
- raspberrypi4-64
- revpi-connect
- rockpi-4b-rk3399
- skx2

... so, `esr-images` is not what I wanted, as it is missing the 3 device types I actually need 😅

As for the errors, there seem to be certain device types where everything has permissions set to not allow public access for reading files, but still allows listing files. For example, the error above is trying to read https://resin-production-img-cloudformation.s3.amazonaws.com/images/imx8mmebcrs08a1/2.100.3/device-type.json which gives `access denied` (as does every other file in `images/imx8mmebcrs08a1/*`), this causes `contractSync` to be able to enumerate the files but not fetch the `device-type.json` and thus the open-balena-api log gets flooded with these errors on first sync.

Seems like there are two possible fixes:
1. Fix `resin-production-img-cloudformation` permissions on these device types to either prohibit both LIST and GET, or permit GET.
2. Create a new ENV variable that can be set for open-balena-api that allows for a list of device-types we actually want. For example: `INCLUDE_IMAGES=iot-gate-imx8,generic-amd64,generic-aarch64` which could filter both the `device type contracts` pulled from https://github.com/balena-io/contracts and the `device type information` pulled from https://resin-production-img-cloudformation.s3.amazonaws.com/images/ ... as I would assume no open-balena instance actually needs every possible device type - and if they do, they can just not set `INCLUDE_IMAGES` and it would fall back to the current behavior of grabbing everything.

### LINKED ISSUE #1696: Access Denied Errors in API Service
**Description**
I'm experiencing repeated Access Denied errors in the API service of OpenBalena. Below is a screenshot of the errors:

The errors look like this:
image

**Questions**
What is causing these Access Denied errors?
How can I resolve these errors to ensure the API service functions correctly?
Any guidance or suggestions to resolve this issue would be greatly appreciated.

Thank you!

Best regards,
Matias Alvarez Sabate

thgreasi:
Fixes: #1696
Fixes: #1433
Fixes: #1060

Change-type: patch
See: https://balena.fibery.io/Work/Project/Migrate-legacy-hostapp-S3-artifacts-to-web-resources-1674

### DIFF
--- a/src/features/device-types/device-types-list.ts
+++ b/src/features/device-types/device-types-list.ts
@@ -1,5 +1,5 @@
 import type { DeviceTypeJson } from './device-type-json.js';
-import { errors } from '@balena/pinejs';
+import { errors, permissions, sbvrUtils } from '@balena/pinejs';
 import * as semver from 'balena-semver';
 const { InternalRequestError } = errors;
 
@@ -18,6 +18,8 @@ import {
 	CONTRACT_ALLOWLIST,
 	IMAGE_STORAGE_DEBUG_REQUEST_ERRORS,
 } from '../../lib/config.js';
+
+const { api } = sbvrUtils;
 
 export interface DeviceTypeInfo {
 	latest: DeviceTypeJson;
@@ -61,25 +63,41 @@ const getFirstValidBuild = async (
 	}
 };
 
+const ALLOWLISTED_DT_SLUGS: string[] = [];
+for (const contractPath of CONTRACT_ALLOWLIST) {
+	const allowListedDtSlug = /^hw.device-type\/([\w-]+)$/.exec(
+		contractPath,
+	)?.[1];
+	if (allowListedDtSlug != null) {
+		ALLOWLISTED_DT_SLUGS.push(allowListedDtSlug);
+	}
+}
+
 export const getDeviceTypes = multiCacheMemoizee(
 	async (): Promise<Dictionary<DeviceTypeInfo>> => {
 		const result: Dictionary<DeviceTypeInfo> = {};
-		let slugs = await listFolders(IMAGE_STORAGE_PREFIX);
+		let deviceTypes = await api.resin.get({
+			resource: 'device_type',
+			passthrough: { req: permissions.rootRead },
+			options: {
+				$select: 'slug',
+				...(ALLOWLISTED_DT_SLUGS.length > 0 && {
+					$filter: {
+						slug: { $in: ALLOWLISTED_DT_SLUGS },
+					},
+				}),
+			},
+		});
 
-		// If there are explicit includes, then everything else is excluded so we need to
-		// filter the slugs list to include only contracts that are in the CONTRACT_ALLOWLIST map
-		if (CONTRACT_ALLOWLIST.size > 0) {
-			const before = slugs.length;
-			slugs = slugs.filter((slug) =>
-				CONTRACT_ALLOWLIST.has(`hw.device-type/${slug}`),
-			);
-			console.log(
-				`CONTRACT_ALLOWLIST reduced device type slugs from ${before} to ${slugs.length}`,
-			);
+		if (deviceTypes.length > 0) {
+			// This is an optimization to avoid multiple 404 queries for DTs that
+			// have a DB record but do not have an OS release published yet.
+			const s3DtSlugs = new Set(await listFolders(IMAGE_STORAGE_PREFIX));
+			deviceTypes = deviceTypes.filter(({ slug }) => s3DtSlugs.has(slug));
 		}
 
 		await Promise.all(
-			slugs.map(async (slug) => {
+			deviceTypes.map(async ({ slug }) => {
 				try {
 					const builds = await listFolders(getImageKey(slug));
 					if (builds.length === 0) {
@@ -111,7 +129,7 @@ export const getDeviceTypes = multiCacheMemoizee(
 			}),
 		);
 
-		if (slugs.length > 0 && Object.keys(result).length === 0) {
+		if (deviceTypes.length > 0 && Object.keys(result).length === 0) {
 			throw new InternalRequestError('Could not retrieve any device type');
 		}
 		return result;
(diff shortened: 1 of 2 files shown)


src/features/device-types/hooks/device-types-cache-invalidation.ts
```diff
@@ -1,6 +1,16 @@
 import { hooks } from '@balena/pinejs';
 import { getDeviceTypes } from '../device-types-list.js';
 
+hooks.addPureHook('POST', 'resin', 'device_type', {
+	POSTRUN: ({ result }) => {
+		if (typeof result !== 'number') {
+			return;
+		}
+		// no need to wait for the cache invalidation
```
otaviojacobi, line 12: NIT: should we add some logging so in case this for whatever reasons runs more frequently than we expected we could get this info from grafana?

thgreasi: I don't think it's necessary but sure

src/features/device-types/device-types-list.ts
```diff
@@ -19,6 +19,8 @@ import {
 	IMAGE_STORAGE_DEBUG_REQUEST_ERRORS,
 } from '../../lib/config.js';
 
+const { api } = sbvrUtils;
```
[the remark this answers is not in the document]
thgreasi: Pinejs-client is our ORM (ie DB sdk) so we shouldn't need to wrap plain DB queries into separate functions, especially when the whole goal of this function is to combine DB & S3 data.

otaviojacobi (approved):

PR #1839, opened 2024-11-12
Add device service fks

otaviojacobi:


src/features/ci-cd/hooks/backfill-device-service-environment-variable.ts
```diff
@@ -0,0 +1,35 @@
+import { hooks, errors, type sbvrUtils } from '@balena/pinejs';
```
Page-, line 1: Moving this to eg a `service-install` feature would make more sense to me as it doesn't actually relate to ci-cd really

otaviojacobi: done

src/migrations/00095-add-device-service-fks-to-device-service-environment-variable.sql
```diff
@@ -0,0 +1,40 @@
+ALTER TABLE "device service environment variable"
+ADD COLUMN IF NOT EXISTS "device" INTEGER NULL;
+
+ALTER TABLE "device service environment variable"
+ADD COLUMN IF NOT EXISTS "service" INTEGER NULL;
+
+CREATE UNIQUE INDEX IF NOT EXISTS "device service environment variable_device_service_name_key"
```
thgreasi, line 7: This should also go to `balena-init.sql`

PR #1840, opened 2024-11-14
Backfill service device on dsev

otaviojacobi:


src/migrations/00097-backfill-device-service-on-device-service-environment-variable.async.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const migration: Migrator.AsyncMigration = {
+	asyncSql: `\
+		UPDATE "device service environment variable" AS dsev
+		SET
+			"device" = si."device",
+			"service" = si."installs-service"
+		FROM "service install" AS si
+		WHERE dsev."service install" = si."id"
+		AND dsev.id IN (
```
thgreasi, line 11: NIT but other than that lgtm :+1: 

```suggestion
		AND dsev."id" IN (
```

src/migrations/00097-backfill-device-service-on-device-service-environment-variable.async.ts
```diff
@@ -0,0 +1,36 @@
...
+			SELECT id
+			FROM "device service environment variable"
```
thgreasi, line 13: Can we also add an alias here and use it in the WHERE just to be super sure?

PR #1848, opened 2024-11-19
Service installs: Allow GCing the missing service filters earlier

thgreasi:
Change-type: patch

### DIFF
--- a/src/features/ci-cd/tasks/service-installs.ts
+++ b/src/features/ci-cd/tasks/service-installs.ts
@@ -101,60 +101,63 @@ const createServiceInstalls = async ({
 
 	const serviceIds = [...new Set([...targetServicesByDevice.values()].flat())];
 
-	const missingServiceFilters = serviceIds.map((serviceId) => ({
-		$not: {
-			service_install: {
-				$any: {
-					$alias: 'si',
-					$expr: {
-						si: {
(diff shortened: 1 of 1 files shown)


src/features/ci-cd/tasks/service-installs.ts
```diff
@@ -101,60 +101,63 @@ const createServiceInstalls = async ({
...
-			passthrough: { req: permissions.rootRead },
-			options: {
-				$select: 'id',
-				$expand: {
-					service_install: {
-						$select: 'installs__service',
-						$filter: {
-							installs__service: { $in: serviceIds },
+	const devicesToAddServiceInstalls = await (async () => {
+		// Creating the filters inside an iife so that they get GCed right after the query.
```
Page-, line 110: Fwiw my understanding is they can be GCed the moment they are no longer referenced so inside or outside the IIFE should make no difference to V8 and primarily serves to block usage outside of it and enforce that it is not referenced later on

thgreasi: @Page- Afaik a variable isn't freed until it goes out of scope, which happens after the function (and all other closures) finish.
I can see though that this will probably not have much impact since the number of service IDs is expected to be low, so I'm closing this to keep the complexity low.

PR #1853, opened 2024-11-20
V2 state get wo si

otaviojacobi:


src/features/device-state/routes/state-get-v2.ts
```diff
@@ -188,6 +183,21 @@ const stateQuery = _.once(() =>
 											is_a_build_of__service: {
 												$select: ['id', 'service_name'],
 												$expand: {
+													device_service_environment_variable: {
+														$select: ['name', 'value'],
+														$filter: {
+															device: {
```
otaviojacobi, line 189: We need this $filter (or some variation of it) as otherwise the expand will get all `device_service_environment_variable` for that server and ignore the `device` filter

otaviojacobi: Some other idea would be to directly expand from the device on the top level to `dsev` but then we need to do something similar for service (which isa bit longer)

test/fixtures/03-device-state/device_service_environment_variables.json
```diff
@@ -5,5 +5,12 @@
 		"service": "app1_service1",
 		"name": "name_si",
 		"value": "value_si"
+	},
+	"si2_svc": {
```
otaviojacobi, line 9: This forces the state get tests to consider the case why I added the $filter on the device expand above

PR #1854, opened 2024-11-20
Add synonymous forms for reverse navigation on device and service to …

otaviojacobi:
…device service environment variable

Change-type: patch

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -395,6 +395,8 @@ Term: device
 		Term Form: service install
 		Database Table Name: service install
 
+		-- Target form for this table is:
+		-- Fact type: device has service has env var name
 		Fact type: service install has name (Auth)
 			Term Form: device service environment variable
 			Database Table Name: device service environment variable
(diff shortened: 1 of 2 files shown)


src/balena.sbvr
```diff
@@ -814,8 +816,14 @@ Fact type: device service environment variable has value
 	Necessity: each device service environment variable has exactly one value.
 Fact type: device service environment variable has service
 	Necessity: each device service environment variable has exactly one service.
+	-- This is the default name that will be given to the synonymous form
+	-- once the device service environment variable origin fact type changes
+	Synonymous form: service has device service environment variable
```
thgreasi, line 821: NIT: Usually we place the `Synonymous form` right below the `Fact type`, and Necessities come right after

thgreasi (approved):
I would expect to also see the changes that use this in the same PR  (since these are not exposed atm on a stable versioned endpoint) which also proves that the changes work, but I'm also fine w/ this PR as is :+1:

src/balena.sbvr
otaviojacobi: Done, thanks

thgreasi:
go for it

thgreasi:
mmmmmmmerge it

PR #1862, opened 2024-11-26
Remove usage of service_install resource on state get v2

otaviojacobi:
This is an alternative approach to https://github.com/balena-io/open-balena-api/pull/1853 which I think it is better as it does use a bit more of API cpu but should reduce DB read cpu in favor of it.

Change-type: patch

### DIFF
--- a/src/features/device-state/routes/state-get-v2.ts
+++ b/src/features/device-state/routes/state-get-v2.ts
@@ -87,18 +87,16 @@ function buildAppFromRelease(
 		// extract the per-image information
 		const image = ipr.image[0];
 
-		const si = device.service_install.find(
-			({ service }) => service.__id === image.is_a_build_of__service[0].id,
-		);
 		const svc = image.is_a_build_of__service[0];
 		const environment: Dictionary<string> = {};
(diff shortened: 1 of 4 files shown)


src/features/device-state/routes/state-get-v2.ts
```diff
@@ -87,18 +87,16 @@ function buildAppFromRelease(
...
 		const svc = image.is_a_build_of__service[0];
 		const environment: Dictionary<string> = {};
 		varListInsert(ipr.image_environment_variable, environment);
 		varListInsert(application.application_environment_variable, environment);
 		varListInsert(svc.service_environment_variable, environment);
 		varListInsert(device.device_environment_variable, environment);
-		if (si != null) {
-			varListInsert(si.device_service_environment_variable, environment);
-		}
+		const dsev = device.device_service_environment_variable.filter(
```
otaviojacobi, line 96: If this proves to be problematic (e.g. we are filtering (O(n)) inside another for loop) we could consider groupby service before the for loop and then get the services with O(1) but I don't think we need to do this prematurely

Page- (approved):

PR #1863, opened 2024-11-26
Update device service environment variable hook to navigate via devic…

otaviojacobi:
…e rather than service install

Change-type: patch


src/features/vars-schema/hooks/vars-update-trigger.ts
```diff
@@ -293,43 +293,7 @@ addEnvHooks('service_environment_variable', async (args) => {
...
-								},
-							},
-						},
-					},
-				},
-			},
-		}),
-	];
-});
+addDeviceEnvHooks('device_service_environment_variable');
```
thgreasi, line 296: How about moving it next to the other two call? :grin:  :
```
addDeviceEnvHooks('device_config_variable');
addDeviceEnvHooks('device_environment_variable');
```

PR #1866, opened 2024-11-27
Remove usage of service_install resource on state get v3

otaviojacobi:
Change-type: patch

### DIFF
--- a/src/features/device-state/routes/state-get-v3.ts
+++ b/src/features/device-state/routes/state-get-v3.ts
@@ -12,7 +12,6 @@ import {
 	getStateDelayingEmpty,
 	getStateEventAdditionalFields,
 	readTransaction,
-	serviceInstallFromImage,
 	varListInsert,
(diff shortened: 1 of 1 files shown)


src/features/device-state/routes/state-get-v3.ts
```diff
@@ -12,7 +12,6 @@ import {
 	getStateDelayingEmpty,
 	getStateEventAdditionalFields,
 	readTransaction,
-	serviceInstallFromImage,
```
thgreasi, line 15: @Page- I think that you can also drop this as part of your major #5477 after this gets merged

thgreasi (approved):

src/features/device-state/routes/state-get-v3.ts
otaviojacobi: @thgreasi It would probably be better to first stop using this on state get v1 before droping this one, but yes, I am aware we need a major to remove this one :)

PR #1867, opened 2024-11-27
Service installs: Avoid creating duplicate tasks on device registration

thgreasi:
Change-type: patch

### DIFF
--- a/src/features/ci-cd/hooks/service-installs.ts
+++ b/src/features/ci-cd/hooks/service-installs.ts
@@ -260,6 +260,13 @@ hooks.addPureHook('POST', 'resin', 'device', {
 		const app = request.values.belongs_to__application;
 		if (app != null) {
 			await createAppServiceInstalls(rootApi, app, [deviceId], tx);
+			if (ASYNC_TASKS_ENABLED && ASYNC_TASK_CREATE_SERVICE_INSTALLS_ENABLED) {
+				// When creating service installs via the async tasks, a single call to createServiceInstallsAsync()
+				// creates all service installs (target release, hostApp & SV release), so we don't need
+				// to run the extra code below to handle them separately, and doing so would create
+				// duplicate & unnecessary tasks.
+				return;
+			}
 		}
 
 		const release = request.values.is_pinned_on__release;
--- a/src/features/ci-cd/hooks/service-installs.ts
+++ b/src/features/ci-cd/hooks/service-installs.ts
@@ -273,6 +273,23 @@ hooks.addPureHook('POST', 'resin', 'device', {
 				tx,
 			);
 		}
+
+		const svAndHostAppReleaseIds = [
+			request.values.should_be_managed_by__release,
+			request.values.should_be_operated_by__release,
+		].filter((releaseId) => releaseId != null);
+
+		if (svAndHostAppReleaseIds.length > 0) {
+			// Create supervisor/hostApp service installs when the supervisor/hostApp is pinned on device creation
+			await createReleaseServiceInstalls(
+				rootApi,
+				[deviceId],
+				{
+					id: { $in: svAndHostAppReleaseIds },
+				},
+				tx,
+			);
+		}
 	},
 });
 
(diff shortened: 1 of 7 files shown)


src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -346,40 +370,20 @@ hooks.addPureHook('PATCH', 'resin', 'device', {
 	},
 });
 
-const addSystemAppServiceInstallHooks = (
-	fieldName: 'should_be_managed_by__release' | 'should_be_operated_by__release',
-) => {
-	hooks.addPureHook('POST', 'resin', 'device', {
```
thgreasi, line 352: Moved this higher, in the `hooks.addPureHook('POST', 'resin', 'device')` that we have for user apps.

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -260,6 +260,13 @@ hooks.addPureHook('POST', 'resin', 'device', {
 		const app = request.values.belongs_to__application;
 		if (app != null) {
 			await createAppServiceInstalls(rootApi, app, [deviceId], tx);
+			if (ASYNC_TASKS_ENABLED && ASYNC_TASK_CREATE_SERVICE_INSTALLS_ENABLED) {
+				// When creating service installs via the async tasks, a single call to createServiceInstallsAsync()
+				// creates all service installs (target release, hostApp & SV release), so we don't need
+				// to run the extra code bellow to handle them separately, and doing so would create
```
joshbwlng, line 266:
```suggestion
				// to run the extra code below to handle them separately, and doing so would create
```

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -260,6 +260,13 @@ hooks.addPureHook('POST', 'resin', 'device', {
...
+				// duplicate & unecessary tasks.
```
joshbwlng, line 267:
```suggestion
				// duplicate & unnecessary tasks.
```

src/features/ci-cd/hooks/service-installs.ts
```diff
@@ -260,6 +260,13 @@ hooks.addPureHook('POST', 'resin', 'device', {
 		const app = request.values.belongs_to__application;
 		if (app != null) {
 			await createAppServiceInstalls(rootApi, app, [deviceId], tx);
+			if (ASYNC_TASKS_ENABLED && ASYNC_TASK_CREATE_SERVICE_INSTALLS_ENABLED) {
```
joshbwlng, line 263: Shouldn't `ASYNC_TASK_CREATE_SERVICE_INSTALLS_ENABLED` only ever equal `true` if `ASYNC_TASKS_ENABLED` is also `true`? Maybe there was a discussion about this at some point, but I feel like system init should fail if `ASYNC_TASKS_ENABLED === false && ASYNC_TASK_CREATE_SERVICE_INSTALLS_ENABLED === true`.

src/features/ci-cd/hooks/service-installs.ts
thgreasi: TIL what that means XD

src/features/ci-cd/hooks/service-installs.ts
thgreasi: Yea I agree on that this was strange. I expected it to only have to check one of the two, maybe something like:
```
export let ASYNC_TASK_CREATE_SERVICE_INSTALLS_ENABLED = ASYNC_TASKS_ENABLED && boolVar(
	'ASYNC_TASK_CREATE_SERVICE_INSTALLS_ENABLED',
	false,
);
```
But since I don't have background on why it was done this way, I just copied it "as is" from above where we do the same check. I guess that it might be this way to make mocking in tests easier. I would prefer to PR such a change separately and fix all occurrences and whichever tests might break after having a short chat on it.

otaviojacobi: This makes sense and it is an improvement on its own, thanks for the suggestion.

otaviojacobi (approved):

PR #1870, opened 2024-11-28
tests: Fix flakyness of service install task expectations

thgreasi:
The hypothesis is that b/c the tasks are
updated by a different write TX, there is a
race condition between waiting for the service
installs to be created and then expected for
the task to have already been marked as
successful.

Change-type: patch


test/25_service-installs.ts
```diff
@@ -212,18 +229,10 @@ export default () => {
 							ctx.app2Service1.id,
 						);
 					});
-					await expectNewTasks(
+					await expectNewSettledTasks(
 						'create_service_installs',
 						isServiceInstallEnabled
 							? [
-									// the first one is from unpinning
```
thgreasi, line 219: Rather than commenting why we expect 2 tasks here, I moved each one right after the action that creates them.

otaviojacobi (approved):

PR #1871, opened 2024-11-28
Remove dsev fk to si

otaviojacobi:


thgreasi:
You should probably also rename the commit to something like:
Add a v7 translation for device_service_environment_variable.service_install

otaviojacobi:
> You should probably also rename the commit to something like: Add a v7 translation for device_service_environment_variable.service_install

This is just a test branch where I am accumulating changes. I will PR each possibly independent change in its standalone PRs

PR #1872, opened 2024-11-29
Rewrite service install FKs on balena model

otaviojacobi:
Change-type: patch

### DIFF
--- a/src/balena-init.sql
+++ b/src/balena-init.sql
@@ -81,10 +81,7 @@ ON "device family" ("is manufactured by-device manufacturer");
 
 -- "device tag"."device" is the first part of an automated unique index
 
--- "device environment variable"."device" is created with the unique index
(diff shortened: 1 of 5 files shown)


src/balena-init.sql
```diff
@@ -82,9 +82,7 @@ ON "device family" ("is manufactured by-device manufacturer");
 -- "device tag"."device" is the first part of an automated unique index
 
 -- "device environment variable"."device" is created with the unique index
-CREATE UNIQUE INDEX IF NOT EXISTS "device service environment variable_device_service_name_key"
-ON "device service environment variable" ("device", "service", "name");
-
+-- created by the "device service environment variable_device_service_name_key" constraint
```
Page-, line 85: Can you combine this with the existing comment, right now they're a bit disjointed as it's two halfs of a single comment written separately

PR #1887, opened 2024-12-09
Make `service install` FK nullable on `device service environment var…

otaviojacobi:
…iable`

Change-type: patch


src/features/cascade-delete/hooks.ts
```diff
@@ -41,13 +43,15 @@ setupDeleteCascade('release', {
...
+	device_service_environment_variable: 'service',
 	service_environment_variable: 'service',
 	service_install: 'installs__service',
 	image: 'is_a_build_of__service',
 	service_label: 'service',
 });
 
-setupDeleteCascade('service_install', {
+// Service install resource should only be used for <= v7 translations
+addDeleteHookForDependents('v7', 'service_install', {
```
Page-, line 54: I would like to see this in the v7 translations so that all v7 translation related hooks/logic is under the one umbrella (folder) rather than potentially spread like this.

Page- (approved):

PR #1892, opened 2024-12-10
Stop filling service install on device service environment variable

otaviojacobi:
Change-type: patch


src/features/service-install/hooks/nullify-si-on-device-service-environment-variable.ts
```diff
@@ -0,0 +1,13 @@
+import { hooks, type sbvrUtils } from '@balena/pinejs';
+
+function backfillDeviceAndService({ request }: sbvrUtils.HookArgs<'resin'>) {
```
thgreasi, line 3: Let's also add a
```
TODO: Drop this once `device_service_environment_variable.service_install` gets removed from the sbvr
```

PR #1908, opened 2024-12-18
Logs: switch to an agnostic primary/secondary backend paradigm

Page-:
Change-type: minor


src/features/device-logs/lib/backends/loki.ts
```diff
@@ -400,12 +400,14 @@ export class LokiBackend implements DeviceLogsBackend {
...
 		return logs.map((log) => {
 			this.validateLog(log);
-			log.version = VERSION;
 			const timestamp = new loki.Timestamp();
 			timestamp.setSeconds(Math.floor(Number(log.nanoTimestamp / 1000000000n)));
 			timestamp.setNanos(Number(log.nanoTimestamp % 1000000000n));
 			// store log line as JSON
-			const logJson = JSON.stringify(log, omitNanoTimestamp);
+			const logJson = JSON.stringify(
+				{ ...log, version: VERSION },
```
Page-, line 408: This change is to not mutate the log, it isn't a problem when loki is the secondary but when it's the primary it can cause problems for the secondary as they're getting a loki versioned log which they don't expect

config/confd/templates/env.tmpl
```diff
@@ -64,6 +64,10 @@ REDIS_LOGS_AUTH={{getenv "REDIS_LOGS_AUTH"}}
 REDIS_LOGS_RO_HOST={{getenv "REDIS_LOGS_RO_HOST"}}
 REDIS_LOGS_RO_AUTH={{getenv "REDIS_LOGS_RO_AUTH"}}
 REDIS_LOGS_SHARDED_PUBSUB={{getenv "REDIS_LOGS_SHARDED_PUBSUB"}}
+LOGS_REDIS_READ_PCT={{getenv "LOGS_REDIS_READ_PCT"}}
```
otaviojacobi, line 67: Can we follow the convention of `REDIS_LOGS` instead of `LOGS_REDIS` used in other places?

src/lib/config.ts
```diff
@@ -342,6 +366,39 @@ export const LOGS_SUBSCRIPTION_EXPIRY_HEARTBEAT_SECONDS =
...
+if (LOGS_PRIMARY_BACKEND === 'loki' && !LOGS_LOKI_ENABLED) {
+	throw new Error(
+		'LOGS_PRIMARY_BACKEND cannot be "loki" if LOGS_LOKI_ENABLED is false',
+	);
+} else if (LOGS_PRIMARY_BACKEND === 'redis' && !LOGS_REDIS_ENABLED) {
+	throw new Error(
+		'LOGS_PRIMARY_BACKEND cannot be "redis" if LOGS_REDIS_ENABLED is false',
+	);
+}
+switch (LOGS_PRIMARY_BACKEND) {
```
otaviojacobi, line 385: Is there any reason for swich case in here? Imo it is just more readable to continue `if/else` logic flow as above 

```
if (LOGS_PRIMARY_BACKEND === 'redis') {
...
} else if (LOGS_PRIMARY_BACKEND === 'loki') {
 ...
}

Comment (Page-) [reply src/lib/config.ts:385]: It mostly comes from typescript complaining if I used `!== 'redis' elseif !== 'loki'` but it's ok with `===`

File: src/features/device-logs/lib/config.ts [hunk 4]
```diff
@@ -1,21 +1,47 @@
...
-	LOKI_ENABLED && LOKI_READ_PCT > Math.random() * 100;
+const LOGS_SECONDARY_BACKEND =
+	LOGS_PRIMARY_BACKEND === 'loki' ? 'redis' : 'loki';
+
+export const LOGS_SECONDARY_BACKEND_ENABLED =
+	LOGS_SECONDARY_BACKEND === 'loki'
+		? LOGS_LOKI_ENABLED
+		: LOGS_SECONDARY_BACKEND === 'redis'
+			? LOGS_REDIS_ENABLED
+			: false;
```
thgreasi, line 23: NIT: Isn't this just v ?
```suggestion
export const LOGS_SECONDARY_BACKEND_ENABLED =
	(LOGS_SECONDARY_BACKEND === 'loki' && LOGS_LOKI_ENABLED) ||
	(LOGS_SECONDARY_BACKEND === 'redis' && LOGS_REDIS_ENABLED);
```

Page-: sure

Page-: Your suggestions fails linting with
```
  19:59  error  Prefer using nullish coalescing operator (`??`) instead of a logical or (`||`), as it is a safer operator  @typescript-eslint/prefer-nullish-coalescing

Comment (Page-) [reply src/features/device-logs/lib/config.ts]: Fixed it

File: src/features/device-logs/lib/store.ts [hunk 5]
```diff
@@ -132,38 +128,31 @@ function handleStoreErrors(req: Request, res: Response, err: Error) {
...
-		) => {
-			await backend.publish(ctx, buffer);
-		};
+const publishBackend = async (ctx: LogContext, buffer: DeviceLog[]) => {
+	const primaryBackendPromise = primaryBackend.publish(ctx, buffer);
+	const secondaryBackendPromise = shouldPublishToSecondary()
+		? secondaryBackend?.publish(ctx, buffer).catch((err) => {
+				captureException(
+					err,
+					`Failed to publish logs to ${LOGS_PRIMARY_BACKEND === 'loki' ? 'redis' : 'loki'}`,
```
thgreasi, line 142: NIT
```suggestion
					`Failed to publish logs to ${LOGS_SECONDARY_BACKEND}`,
```

PR #1918, opened 2024-12-24
Loki: use structured metadata rather than a JSON line

Page-:
Change-type: major


src/features/device-logs/lib/backends/loki.ts
```diff
@@ -384,23 +407,36 @@ export class LokiBackend implements DeviceLogsBackend {
...
 			timestamp.setNanos(Number(log.nanoTimestamp % 1000000000n));
-			// store log line as JSON
-			const logJson = JSON.stringify(
-				{ ...log, version: VERSION },
-				omitNanoTimestamp,
-			);
+			const structuredMetadata = [
+				deviceId,
+				new loki.LabelPairAdapter()
+					.setName('timestamp')
```
Page-, line 421: Should we name this piece of metadata as "device_timestamp" instead to make it clearer what it actually is?

Page-, line 421: This is also questionable as to whether we should store it as structured metadata because every entry for it will be unique and possibly make a bunch of other things more painful and is probably not ergonomic or even useful to query by

PR #1933, opened 2025-01-17
Use opentelemetry to add trace ids to request and error logs

Page-:


index.js
```diff
@@ -2,6 +2,20 @@ const start = async () => {
...
+	);
+	const { HttpInstrumentation } = await import(
+		'@opentelemetry/instrumentation-http'
+	);
+
+	const sdk = new NodeSDK({
+		instrumentations: [new ExpressInstrumentation(), new HttpInstrumentation()],
+	});
+
+	sdk.start();
```
joshbwlng, line 17: The docs also mention a `shutdown()` that can be called to cleanly close the SDK client, but I suppose it's not required and we don't have an existing SIGTERM/etc handler here.

joshbwlng (approved):

PR #1936, opened 2025-01-22
Fix PATCH /state/v3 not clearing the device's is_running__release

thgreasi:
Change-type: patch


test/03_device-state.ts
```diff
@@ -1799,6 +1799,108 @@ export default () => {
...
+						{
+							[stateKey]:
+								stateVersion === 'v2'
+									? {
+											is_on__commit: null,
+										}
+									: {
+											apps: {
+												[applicationUuid]: {
+													release_uuid: null,
```
pipex, line 1844: Not sure this would ever happen. The `release_uuid` property would just be undefined

thgreasi: I realized, but I wanted to make sure that in case we explicitly pass `null` in this prop, then it is indeed getting stored, b/c that's what makes sense to do any way.
Effectively I treat the supervisor omitting the `release_uuid` when it's `null`, just as an optimization, and make sure that the basic semantics work as well (even though we don't  expect it to happen).

PR #1959, opened 2025-02-16
Loki: switch to the http/websockets api for tailing logs

Page-:
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "@types/semver": "^7.5.8",
     "@types/statuses": "^2.0.5",
     "@types/validator": "^13.12.2",
+    "@types/ws": "^8.5.14",
     "avsc": "^5.7.7",
     "aws-sdk": "^2.1692.0",
     "balena-device-config": "^6.3.0",
@@ -90,6 +91,7 @@
     "basic-auth": "^2.0.1",
     "bluebird": "^3.7.2",
     "body-parser": "^1.20.3",
+    "bufferutil": "^4.0.9",
     "cache-manager": "^4.1.0",
     "cache-manager-ioredis": "^2.1.0",
     "common-tags": "^1.8.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -83,13 +83,15 @@
...
     "@types/validator": "^13.12.2",
+    "@types/ws": "^8.5.14",
     "avsc": "^5.7.7",
     "aws-sdk": "^2.1692.0",
     "balena-device-config": "^6.3.0",
     "balena-semver": "^2.3.5",
     "basic-auth": "^2.0.1",
     "bluebird": "^3.7.2",
     "body-parser": "^1.20.3",
+    "bufferutil": "^4.0.9",
```
thgreasi, line 94: Is this used anywhere?

src/features/device-logs/lib/backends/loki.ts
```diff
@@ -297,37 +276,65 @@ export class LokiBackend implements DeviceLogsBackend {
...
 			});
-			this.tailCalls.set(key, call);
+
+			ws.on('open', function open() {
+				console.log('connected');
+			});
+
+			ws.on('message', (data) => {
+				try {
+					const result = JSON.parse(data.toString()) as {
```
Page-, line 308: https://grafana.com/docs/loki/latest/reference/loki-http-api/#stream-logs

package.json
Page-: It's an opt-in for performance for ws: https://github.com/websockets/ws?tab=readme-ov-file#opt-in-for-performance

test/13_loki-backend.ts
```diff
@@ -60,27 +60,6 @@ export default () => {
 			expect(history.at(-1)).to.deep.equal(convertToOutputLog(log));
 		});
 
-		it('should convert multiple logs with different labels to streams and then back to logs', function () {
```
Page-, line 63: I deleted this test as it no longer makes sense and tests the internals (which no longer exist), so it makes more sense to test via the published/supported APIs

PR #1965, opened 2025-02-24
Loki: improve error handling when fetching historical logs

Page-:
Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/backends/loki.ts
+++ b/src/features/device-logs/lib/backends/loki.ts
@@ -184,7 +184,7 @@ export class LokiBackend implements DeviceLogsBackend {
 	): Promise<OutputDeviceLog[]> {
 		const ctx = await assertLokiLogContext($ctx);
 
-		const [, body] = await requestAsync({
+		const [{ statusCode }, body] = await requestAsync({
 			url: `http://${lokiQueryAddress}/loki/api/v1/query_range`,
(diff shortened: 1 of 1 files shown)


src/features/device-logs/lib/backends/loki.ts
```diff
@@ -198,6 +198,12 @@ export class LokiBackend implements DeviceLogsBackend {
 			gzip: LOKI_HISTORY_GZIP,
 		});
 
+		if (statusCode !== 200) {
+			throw new Error(
```
otaviojacobi, line 202: @Page- will this unhandled error show up on sentry (e.g. generic type Error rather than a properly HttpError?)

Page-: Yes, every path that makes it here has a generic error handler with reporting to sentry, this is mostly just changing the error from
```
     TypeError: Cannot read properties of undefined (reading 'result')
      at LokiBackend.history (file:///usr/src/app/src/features/device-logs/lib/backends/loki.ts:129:28)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async Context.<anonymous> (file:///usr/src/app/test/13_loki-backend.ts:102:29)
```
to the actual error returned from loki

otaviojacobi (approved):

PR #1966, opened 2025-02-25
minor: add secureboot to device-config endpoint

aethernet:
To provision device with secureboot, we need the config.json to contain the secureboot installer entry.

- [FIBERY](https://balena.fibery.io/Work/Improvement/Update-the-device-config-backend-for-config.json-2609)

### DIFF
--- a/src/features/device-config/device-config.ts
+++ b/src/features/device-config/device-config.ts
@@ -150,5 +150,13 @@ export const generateConfig = async (
 		config.developmentMode = true;
 	}
 
+	const securebootInstaller = getBodyOrQueryParam(
+		req,
+		'secureboot',
+	)?.toString();
+	if (['true', 'on', '1'].includes(securebootInstaller || '')) {
+		config.installer = { secureboot: true };
+	}
+
 	return config;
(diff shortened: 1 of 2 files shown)


src/features/device-config/device-config.ts
```diff
@@ -150,5 +150,13 @@ export const generateConfig = async (
 		config.developmentMode = true;
 	}
 
+	const securebootInstaller = getBodyOrQueryParam(
+		req,
+		'secureboot',
+	)?.toString();
+	if (['true', 'on', '1'].includes(securebootInstaller || '')) {
+		config.installer = { secureboot: true };
```
Page-, line 158: I've checked the fibery link but I don't see where it's documented that the format should be
```json
{ installer: { secureboot: true } }
```
Do you know where I would confirm that format is correct?

aethernet: It's here https://github.com/balena-os/meta-balena/blob/master/docs/uefi-secure-boot.md#provisioning

Page- (approved):

PR #1967, opened 2025-02-26
Check access using supervisor-proxy-write permission

joshbwlng:
Use the supervisor-proxy-write permission to determine who
can execute POST/PUT/PATCH/DELETE actions via the supervisor
proxy on device and application resources.

Change-type: major

---

See: https://balena.fibery.io/Work/Project/Reduce-frequency-of-accidental-device-deletions---new-roles-750

Depends on:
- https://github.com/balena-io/open-balena-api/pull/1978

Note: Marking as a major as consumers (such as balena-api) will need to make sure they have `supervisor-proxy-write` permissions properly configured or supervisor proxy requests may begin failing.

### DIFF
--- a/src/features/device-proxy/device-proxy.ts
+++ b/src/features/device-proxy/device-proxy.ts
@@ -125,7 +125,16 @@ export const proxy = async (req: Request, res: Response) => {
 			throw new BadRequestError('At least one filter must be specified');
 		}
 
-		const responses = await requestDevices({ url, req, filter, data, method });
+		const responses = await requestDevices({
+			url,
+			req,
+			filter,
+			data,
+			method,
+			// Only pass appId if we're not filtering by deviceId or uuid
+			// This means we can check permissions on the application level only
+			...(deviceId == null && uuid == null && { appId }),
+		});
 		if (responses.length === 1) {
 			validateSupervisorResponse(responses[0], req, res, filter);
 			return;
@@ -146,10 +155,28 @@ interface FixedMethodRequestDevicesOpts {
 	data?: AnyObject;
 	req?: sbvrUtils.Passthrough['req'];
 	wait?: boolean;
+	appId?: number;
 }
 
 interface RequestDevicesOpts extends FixedMethodRequestDevicesOpts {
 	method: string;
+}
+
+// Check user resource access via /canAccess
+async function checkResourceAccess(
+	client: typeof api.resin,
+	resource: 'application' | 'device',
+	id: number,
+): Promise<void> {
+	const res = (await client.request({
+		method: 'POST',
+		url: `${resource}(${id})/canAccess`,
+		body: { action: 'supervisor-proxy-write' },
+	})) as { d?: Array<{ id: number }> };
+
+	if (res?.d?.[0]?.id !== id) {
+		throw new errors.ForbiddenError();
+	}
 }
 
 // - req is the express req object, if passed then(the permissions of the user making
@@ -178,6 +205,7 @@ async function requestDevices({
 	req,
 	wait = true,
 	method = 'POST',
+	appId,
 }: RequestDevicesOpts): Promise<undefined | RequestResponse[]> {
 	if (url == null) {
 		throw new BadRequestError('You must specify a url to request!');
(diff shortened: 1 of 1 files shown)


src/lib/auth.ts
```diff
@@ -51,6 +51,10 @@ export const ROLES: {
 		'resin.application_tag.all',
 		'resin.application_type.all',
 		'resin.device.all',
+		'resin.device.purge',
```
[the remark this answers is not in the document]
joshbwlng: Need to check if a device API key can be used to execute these actions via the device proxy (can do this via the SDK in a browser tab when not logged in). Because we need to know if we also need to grant devices access to themselves or if we can limit it to Application Membership roles.

src/features/device-proxy/device-proxy.ts
```diff
@@ -217,12 +217,16 @@ async function requestDevices({
 		// Check for device update permission, except for
 		// internal operation of the platform.
 		if (method !== 'GET' && req !== permissions.root) {
+			const urlAction = url.split('/').pop() ?? '';
+			const action = ['purge', 'shutdown'].includes(urlAction)
```
joshbwlng, line 221: Can do simple `===` check instead of `.includes()`

src/lib/auth.ts
```diff
@@ -56,6 +56,15 @@ export const ROLES: {
 		'resin.device_environment_variable.all',
 		'resin.device_tag.all',
 		'resin.device_service_environment_variable.all',
+
+		// These are for the supervisor proxy
+		'resin.device.purge',
+		'resin.device.shutdown',
+
+		// These are for the actions proxy
```
joshbwlng, line 64: These may not be necessary to add here as I believe open-balena users do not have the actions proxy anyway. Will rebase on https://github.com/balena-io/open-balena-api/pull/1978 once merged, adding permissions here so we can more easily test https://github.com/balena-io/balena-api/pull/5655

src/features/device-proxy/device-proxy.ts
```diff
@@ -214,22 +289,27 @@ async function requestDevices({
...
+			// Otherwise we will break things by checking application.update permissions for things like Operators calling application.reboot()
+			if (appId != null && ['shutdown', 'purge'].includes(permission)) {
+				await checkResourceAccess(resinApi, 'application', appId, permission);
+			} else {
+				await Promise.all(
+					deviceIds.map(async (deviceId) =>
+						checkResourceAccess(resinApi, 'device', deviceId, permission),
+					),
+				);
+			}
```
joshbwlng, line 312: I'd prefer to allow application level permission checks as much as possible for performance reasons, but that will require that we add `resin.application.reboot` etc to `auth.ts` and allow those currently able to execute those actions via `resin.device.update` so we don't break anything.

src/features/device-proxy/device-proxy.ts
```diff
@@ -125,7 +125,16 @@ export const proxy = async (req: Request, res: Response) => {
...
-		const responses = await requestDevices({ url, req, filter, data, method });
+		const responses = await requestDevices({
+			url,
+			req,
+			filter,
+			data,
+			method,
+			// Only pass appId if we're not filtering by deviceId or uuid
+			// This means we can check permissions on the application level only
+			...(deviceId == null && uuid == null ? { appId } : {}),
```
thgreasi, line 136: NIT
```suggestion
			...(deviceId == null && uuid == null && { appId }),
```

src/features/device-proxy/device-proxy.ts
```diff
@@ -214,22 +243,20 @@ async function requestDevices({
...
-					})) as { d?: Array<{ id: number }> };
-
-					if (res?.d?.[0]?.id !== deviceId) {
-						throw new errors.ForbiddenError();
-					}
-				}),
-			);
+			const permission = ['PUT', 'PATCH', 'POST', 'DELETE'].includes(method)
+				? 'supervisor-proxy-write'
+				: 'update';
```
thgreasi, line 250: Isn't this covering all cases really? I mean, unless I'm forgetting something, I would expect `['PUT', 'PATCH', 'POST', 'DELETE'].includes(method)` to be always true, so we could just do:
```suggestion
			const permission = 'supervisor-proxy-write';
```

joshbwlng: Yeah, it was about `HEAD` requests. If it's fine to not check `/canAccess` for `HEAD` requests then I'd prefer to go with your suggestion and add it next to `method !== 'GET'`.

thgreasi: I think so, especially given that the supervisor doesn't support any HEAD requests atm afaict https://docs.balena.io/reference/supervisor/supervisor-api/

src/features/device-proxy/device-proxy.ts
```diff
@@ -146,12 +155,30 @@ interface FixedMethodRequestDevicesOpts {
...
 
+// Check user resource access via /canAccess
+async function checkResourceAccess(
+	client: typeof api.resin,
+	resource: 'application' | 'device',
+	id: number,
+): Promise<void> {
+	const res = (await client.request({
+		method: 'POST',
+		url: `${resource}(${id})/canAccess`,
```
Page-, line 173: @otaviojacobi isn't there a nice way of doing `canAccess` calls via pinejs-client now? Or was that only specific to the permissions?

Page- (commented):
Seems good

src/features/device-proxy/device-proxy.ts
thgreasi: Afaict it's only supported inside `$filter` (since the implementation was added in `handleFilterOperator`).
See: https://github.com/balena-io-modules/pinejs-client-js/pull/183/files

thgreasi (approved):

PR #1977, opened 2025-03-14
minor: extend device state for secureboot and FDE status

aethernet:
https://balena.fibery.io/Work/Improvement/Spec-for-Supervisor-+-API-changes-2388

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -193,8 +193,14 @@ Term: memory usage
 Term: memory total
 	Concept Type: Integer (Type)
 
+Term: secureboot enabled
+	Concept Type: Boolean (Type)
+
 Term: storage block device
 	Concept Type: Short Text (Type)
+
+Term: storage encrypted
+	Concept Type: Boolean (Type)
 
 Term: storage usage
 	Concept Type: Integer (Type)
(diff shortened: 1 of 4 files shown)


src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -49,6 +49,9 @@ export type StatePatchV2Body = {
 		api_secret?: string;
 		memory_usage?: number;
 		memory_total?: number;
+		is_secureboot_enabled?: boolean;
+		is_storage_encrypted?: boolean;
+		secureboot_keys_metadata?: string | null;
```
Page-, line 54: State-patch v2 will never report these props:
```suggestion
```

src/balena.sbvr
```diff
@@ -193,6 +193,9 @@ Term: memory usage
 Term: memory total
 	Concept Type: Integer (Type)
 
+Term= secureboot keys metadata
+	Concept Type: Short Text (Type)
```
Page-, line 197:
```suggestion
	Concept Type: JSON (Type)
```

src/migrations/00102-device-add-secureboot-fde-status.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "device" ADD COLUMN IF NOT EXISTS "is secureboot enabled" BOOLEAN NULL;
+ALTER TABLE "device" ADD COLUMN IF NOT EXISTS "is storage encrypted" BOOLEAN NULL;
+ALTER TABLE "device" ADD COLUMN IF NOT EXISTS "secureboot keys metadata" JSONB NULL;
```
Page-, line 3: This doesn't actually match the generated schema from the sbvr so this would be an invalid migration, it's better to fix at the sbvr level but it's worth noting that you can generate the schema via `npm run generate-model-schema` and make sure the column types match

src/translations/v6/v6.sbvr
```diff
@@ -187,6 +187,9 @@ Term: memory usage
 Term: memory total
 	Concept Type: Integer (Type)
 
+Term= secureboot keys metadata
```
Page-, line 190: Personally I wouldn't expose this on v6, if it's info people want then they can update to using v7

PR #1978, opened 2025-03-18
Add supervisor-proxy-write permission

joshbwlng:
Add new permission for supervisor proxy POST/PUT/PATCH/DELETE
operations to give us more granular access control.

Change-type: minor

---

See: https://balena.fibery.io/Work/Project/Reduce-frequency-of-accidental-device-deletions---new-roles-750

Required for:
- https://github.com/balena-io/open-balena-api/pull/1967

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -43,6 +43,7 @@ export const ROLES: {
 		'resin.actor.delete?id eq @__ACTOR_ID',
 		'resin.api_key.read?is_of__actor eq @__ACTOR_ID',
 		'resin.application.all',
+		'resin.application.supervisor-proxy-write',
 		'resin.device_type.read',
 		'resin.device_type_alias.read',
 		'resin.cpu_architecture.read',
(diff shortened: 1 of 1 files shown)


src/lib/auth.ts
```diff
@@ -56,6 +56,15 @@ export const ROLES: {
 		'resin.device_environment_variable.all',
 		'resin.device_tag.all',
 		'resin.device_service_environment_variable.all',
+
+		// These are for the supervisor proxy
+		'resin.device.purge',
+		'resin.device.shutdown',
+
+		// These are for the actions proxy
```
joshbwlng, line 64: These may not be necessary to add here as I believe open-balena users do not have the actions proxy anyway

thgreasi: Indeed I think you can drop these

thgreasi (dismissed):
Unless @Page- wants to suggest a different name (eg a prefix), LGTM :+1:  (I suggest a zulip ping)

thgreasi (approved):

PR #1980, opened 2025-03-19
Update to sentry 9.x

Page-:
Change-type: patch

### DIFF
--- a/src/infra/error-handling/index.ts
+++ b/src/infra/error-handling/index.ts
@@ -2,7 +2,6 @@ import _ from 'lodash';
 import * as Sentry from '@sentry/node';
 import escapeHtml from 'escape-html';
 
-import type { hooks } from '@balena/pinejs';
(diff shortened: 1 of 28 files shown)


src/infra/error-handling/index.ts
```diff
@@ -47,15 +45,10 @@ export function captureException(
...
 	Sentry.withScope((scope) => {
 		if (options != null) {
-			const { tags, req, extra } = options;
+			const { tags, extra } = options;
 			if (tags != null) {
 				scope.setTags(tags);
 			}
-			if (req != null) {
-				scope.addEventProcessor((evt) =>
-					Sentry.addRequestDataToEvent(evt, req),
```
Page-, line 56: My understanding is that nowadays this should be automatically handled by https://docs.sentry.io/platforms/javascript/guides/node/configuration/integrations/requestdata/ and we should not need to manually do it

thgreasi (approved):

PR #1981, opened 2025-03-19
Switch to native promises for the request wrapper/jwt-passport handling

Page-:


src/infra/auth/jwt-passport.ts
```diff
@@ -95,11 +94,14 @@ export const createStrategy = (
 					const processedJwtUser = jwtUser as ResolvedUserPayload;
 					processedJwtUser.actor = user.actor.__id;
 					processedJwtUser.permissions = userPermissions;
-					return processedJwtUser;
+					done(null, processedJwtUser);
```
thgreasi, line 97: Heads-up that there are more `return` statements higher.
How about wrapping the whole function in an iife so that we leave all the `return`s as is and we call `done` just once?

Page-: Yeah I've changed the approach to callback-ifying this so it is much easier to get right now/hard to get wrong

src/infra/request-promise/index.ts
```diff
@@ -1,54 +1,27 @@
...
 import request from 'request';
 
 import { EXTERNAL_HTTP_TIMEOUT_MS } from '../../lib/config.js';
 
-type Request = typeof request;
-
 export type RequestResponse = [request.Response, any];
 
-interface PromisifiedRequest extends Request {
-	getAsync(
```
thgreasi, line 11: I'm going to miss some of there :cry:

thgreasi (approved):

src/infra/request-promise/index.ts
Page-: In practice we barely use them and they can be replaced with `requestAsync({ method, ...opts })

PR #1983, opened 2025-03-21
Tests: use faster config for postgres

Page-:
These settings are not safe for data we care about as we might lose or corrupt data in the case of power failure, but for test data where that is not an issue they offer a performance benefit

Change-type: patch


joshbwlng:
@Page- Oh yeah, I remember using these for Jellyfish: https://github.com/product-os/autumndb/blob/master/docker-compose.yml#L21

joshbwlng (approved):

PR #2014, opened 2025-04-25
Update nock to v14

joshbwlng:
Change-type: patch


test/20_supervisor_notification.ts
```diff
@@ -146,6 +147,7 @@ export default () => {
 
 				// Just to confirm that the MAX_SAFE_SQL_BINDS limitation is still valid
 				it('should fail when providing more parameters to $in than PG support', async function () {
+					nock.restore();
```
joshbwlng, line 150: I'm not sure why disabling interceptors make any difference here. I'm wondering if it is more about extra processing time due to interceptors and not necessarily the interceptors themselves causing problems?

PR #2017, opened 2025-05-08
Allow customizing webresource objects storage class

otaviojacobi:
Ideally we would default to INTELLIGENT_TIERING, however, Minio does not support this storage class and fails with "Storage class not supported". In order to avoid breaking open-balena & balena-machine installations I am defaulting to standard and we can override on the staging/production configmap.

See: https://balena.fibery.io/search/V9pZn#Work/Improvement/Allow-configurable-storage-classes-for-uploads-2816

Change-type: patch


joshbwlng:
@otaviojacobi I guess this doesn't require any test additions/changes?

otaviojacobi:
> Member

The current release asset tests should be enough to ensure this is  customizable

joshbwlng:
@otaviojacobi 
Happy to approve, but noticed a couple typos in your commit message. Not a big deal, but thought it worth fixing before we merge: `Ideally we woul default to INTELLGENT_TIERING` -> `Ideally we would default to INTELLIGENT_TIERING`

otaviojacobi:
thanks @joshbwlng my (low level) dyslexia always bites me in the commit message as I haven't yet found a good spell checker for the terminal :)

thgreasi (approved):

PR #2022, opened 2025-05-13
Allow customizing the fetching of authorized keys based on username

Page-:
Change-type: patch

### DIFF
--- a/src/features/auth/public-keys.ts
+++ b/src/features/auth/public-keys.ts
@@ -1,12 +1,50 @@
 import type { RequestHandler } from 'express';
-
-import { sbvrUtils } from '@balena/pinejs';
+import { type permissions, sbvrUtils } from '@balena/pinejs';
 
 import {
 	captureException,
 	handleHttpErrors,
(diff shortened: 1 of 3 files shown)


src/features/auth/public-keys.ts
```diff
@@ -1,43 +1,57 @@
...
+	return data.map((e) => e.public_key).join('\n');
+};
+
+let getAuthorizedKeys = defaultGetAuthorizedKeysFn;
+/**
+ * Set the function used to get the authorized keys for a user.
+ */
+export const setGetAuthorizedKeysFn = (
+	$getAuthorizedKeys: typeof getAuthorizedKeys,
+) => (getAuthorizedKeys = $getAuthorizedKeys);
```
otaviojacobi, line 45: Is there any reason to return the operation result here rather than just returning `Promise<void>` ?
```suggestion
) => {getAuthorizedKeys = $getAuthorizedKeys};
```

otaviojacobi (approved):

PR #2026, opened 2025-05-20
Add size constraints to string fields

thgreasi:
Change-type: major
See: https://balena.fibery.io/Work/Project/API-Limit-size-of-large-fields-975


joshbwlng:
I'm sure it was scripted, but just in case I double-checked that all lengths mentioned in the SBVR match what's in the migration and didn't find any issues. :+1:

src/features/device-state/state-patch-utils.ts
```diff
@@ -61,32 +61,56 @@ const truncateText = (
...
+			if (typeof object[key] !== 'string' || object[key].length <= maxLength) {
+				continue;
+			}
+			ThisShouldNeverHappenError(
+				`Received ${resource}.${key} for id ${recordId} that was too long and had to be truncated: ${object[key].length} chars`,
+			);
+			const truncatedText = truncateText(object[key], maxLength, delimiter);
+			// @ts-expect-error We have already checked that object[key] is a string, but TS complains that string is not assignable to O[TextFieldsOf<T>]
+			object[key] = truncatedText;
+		}
```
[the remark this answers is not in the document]
thgreasi: I can do the `const value = object[key];` part to save the extra prop accesses, but that does allow removing the ts-expect error. The only reason you were able to remove the ts-expect error was actually the destructuring, but that would cause extra GC, which I would prefer to avoid.

src/features/field-size-limits/setup-field-size-limits.ts
```diff
@@ -0,0 +1,51 @@
...
+import { odataNameToSqlName } from '@balena/odata-to-abstract-sql';
+
+const { BadRequestError } = pinejsErrors;
+
+const vowels = ['a', 'e', 'i', 'o', 'u'];
+
+/**
+ * This is a HACH and should only be used when pre-existing data blocks us from adding the respective Necessity directly in the sbvr.
+ */
+export function addHooksForFieldSizeLimitChecks(
```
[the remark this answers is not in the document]
thgreasi: Added tests for api keys in oB.
 Since the rests of the resources are atm tested only on bC, and in order to avoid duplicating & fragmenting the tests, I decided to add the extra tests there (in bC).
This should make it easier for us to later move all of them to oB

test/07_versioned-releases.ts
```diff
@@ -137,6 +137,48 @@ export default () => {
...
+			}
+
+			for (const [propName, maxSize] of [
+				// express is configured to not parse json request bodies larger than 512kb,
+				// but it seems that that's before decompressing so we are able to use values
+				// slightly larger than that.
+				['build_log', 524_200],
+				['dockerfile', 524_200],
+			] as const) {
+				it(`should be able to set the a ${maxSize} chars long ${propName} on an image`, async () => {
```
joshbwlng, line 169:
```suggestion
				it(`should be able to set ${maxSize} chars long ${propName} on an image`, async () => {
```

test/07_versioned-releases.ts
```diff
@@ -235,6 +277,45 @@ export default () => {
 				);
 			});
 
+			for (const [propName, maxSize] of [
+				['note', 1_000_000],
+				['build_log', 1_000_000],
+			] as const) {
+				it(`should not be able to set the a ${propName} longer than ${maxSize} on a release`, async () => {
```
joshbwlng, line 284:
```suggestion
				it(`should not be able to set a ${propName} longer than ${maxSize} chars on a release`, async () => {
```

test/08_api-keys.ts
```diff
@@ -502,6 +510,121 @@ export default () => {
 						expect(apiKey).to.be.a('string');
 						expect(apiKey).to.not.be.empty;
 					});
+
+					it('should not allow requests with a name longer than 50 characters', async function () {
+						const name = 'too-long-name-'.padEnd(51, 'a');
+						const description = 'long description '.padEnd(150, 'a');
```
joshbwlng, line 516: Is the `description` padding here necessary?

thgreasi: No, I just liked the word separation :)

thgreasi: No, it isn't, since I later added an explicit test case for the max size of both fields :+1:

thgreasi:
Since you only raised comments about the tests, I went ahead and squashed after addressing your comments.
This should be ready for review.

joshbwlng (approved):

PR #2049, opened 2025-06-19
Migrate to timestamptz

joshbwlng:
Change-type: patch

---

See: https://balena.fibery.io/Work/Project/Use-time-zone-based-timestamps-for-PineJS-dates-965


src/balena-init.sql
```diff
@@ -189,9 +189,9 @@ CREATE INDEX "application_slug_public_host_idx"
 ON "application" ("slug" varchar_pattern_ops, "is public", "is host");
 
 CREATE INDEX IF NOT EXISTS "scheduled_job_run_start_timestamp_idx"
-ON "scheduled job run" (DATE_TRUNC('milliseconds', "start timestamp"));
+ON "scheduled job run" (DATE_TRUNC('milliseconds', "start timestamp", 'UTC'));
```
joshbwlng, line 192: Assuming 'UTC' is my only concern here as that may not be the case for everyone

Page-: I don't think we query by timezones other than UTC do we? So it shouldn't be an issue

Page- (approved):

PR #2050, opened 2025-06-19
Update lf-to-abstract-sql to support multiple necessities

thgreasi:
Change-type: patch
See: https://balena.fibery.io/Work/Project/API-Limit-size-of-large-fields-975

### DIFF
--- a/src/migrations/00104-update-index-constraint-for-lf-to-abstract-sql-bump.sql
+++ b/src/migrations/00104-update-index-constraint-for-lf-to-abstract-sql-bump.sql
@@ -1,0 +1,46 @@
+DO $$
+BEGIN
+	IF NOT EXISTS (
+		SELECT 1
+		FROM information_schema.table_constraints tc
+		WHERE tc.CONSTRAINT_TYPE = 'CHECK'
+			AND tc.table_schema = CURRENT_SCHEMA()
+			AND tc.table_name = 'application'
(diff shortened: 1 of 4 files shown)


src/migrations/00104-update-index-constraint-for-lf-to-abstract-sql-bump.sql
```diff
@@ -0,0 +1,46 @@
...
+	IF NOT EXISTS (
+		SELECT 1
+		FROM information_schema.table_constraints tc
+		WHERE tc.CONSTRAINT_TYPE = 'CHECK'
+			AND tc.table_schema = CURRENT_SCHEMA()
+			AND tc.table_name = 'application'
+			AND tc.constraint_name = 'application$QqBV+60jro8TubDvmFVw6BwX9diAf/SNdJ1816Yxpx0='
+	) THEN
+		ALTER TABLE "application" ADD CONSTRAINT "application$QqBV+60jro8TubDvmFVw6BwX9diAf/SNdJ1816Yxpx0=" CHECK (
+			((32 = LENGTH("uuid")) AND (LENGTH("uuid") IS NOT NULL) AND ("uuid" IS NOT NULL))
```
thgreasi, line 12: I realize that since the only change is switching sides in the operands of `LENGTH("uuid") = 32` comparison, we could have just renamed the constraint.
On the other hand I decided that it would make things easier to maintain & double-check if we recreated the constraint so that the SQL in the catalog is exactly the same between a migrated & a fresh DB.

Page- (approved):

PR #2070, opened 2025-07-28
Add size constraints to JSON fields

thgreasi:
Change-type: major
See: https://balena.fibery.io/Work/Project/re-pitching-API-Limit-size-of-large-fields-audit-the-outliers-975
See: https://balena.fibery.io/Work/Project/API-Limit-the-size-of-JSON-fields-1723

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -677,6 +677,8 @@ Fact type: device type has logo
 
 Fact type: device type has contract
 	Necessity: each device type has at most one contract
+	-- 5MB
+	Necessity: each device type that has a contract, has a contract that is represented by a Text (Type) that has a Length (Type) that is less than or equal to 5000000.
 
 Fact type: device type belongs to device family
 	Synonymous form: device family has device type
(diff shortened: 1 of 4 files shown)


joshbwlng:
Didn't see any issues, but will leave it open for now to give others a chance to look. :+1:

thgreasi:
added some extra tests

src/balena.sbvr
```diff
@@ -796,6 +802,8 @@ Fact type: release has asset key
 -- release asset
 Fact type: release asset has asset
 	Necessity: each release asset has at most one asset.
+	-- The content_type size is as per RFC6838 recommendation: "<type-name> and <subtype-name> SHOULD be limited to 64 characters."
+	Necessity: each release asset that has an asset, has an asset that has a Filename (Type) that has a Length (Type) that is less than or equal to 255 and has a Content Type (Type) that has a Length (Type) that is less than or equal to 129.
```
otaviojacobi, line 806: These look fine to me. You might want to check with product if 255 is not too big (but I think it is fine as it is also kind of similar to what windows allows)

PR #2072, opened 2025-08-01
Update TypeScript to 5.9.2

thgreasi:
Change-type: patch
See: https://balena.fibery.io/Work/Project/1863

### DIFF
--- a/src/features/device-logs/lib/backends/redis.ts
+++ b/src/features/device-logs/lib/backends/redis.ts
@@ -270,7 +270,7 @@ export class RedisBackend implements DeviceLogsBackend {
 		payload: string,
 	): Promise<RedisDeviceLog | undefined> {
 		try {
-			let decompressedBuffer = Buffer.from(payload, BUFFER_ENCODING);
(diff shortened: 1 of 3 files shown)


src/features/device-logs/lib/backends/redis.ts
```diff
@@ -270,7 +270,7 @@ export class RedisBackend implements DeviceLogsBackend {
 		payload: string,
 	): Promise<RedisDeviceLog | undefined> {
 		try {
-			let decompressedBuffer = Buffer.from(payload, BUFFER_ENCODING);
+			let decompressedBuffer: Buffer = Buffer.from(payload, BUFFER_ENCODING);
```
thgreasi, line 273: `Buffer.from` returns `Buffer<ArrayBuffer>, but `compression.uncompress` returns `Promise<Buffer>`, and now `Buffer` has a default generic of `Buffer<ArrayBufferLike>`, so we had to define decompressedBuffer with the wider `Buffer<ArrayBufferLike>`, which then eslint simplifies to `Buffer`, since `ArrayBufferLike` is the default.

Page- (approved):

PR #2087, opened 2025-08-25
contractSync: Add support for authenticating using PAT

thgreasi:
Change-type: minor
See: https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28
See: https://balena.fibery.io/Work/Project/Add-support-for-using-PAT-for-authenticating-in-the-conractsSync-of-the-API-1789


src/features/contracts/contracts-directory.ts
```diff
@@ -125,9 +125,15 @@ const prepareContractDirectory = async (repo: RepositoryInfo) => {
...
 const getRequestOptions = (repo: RepositoryInfo) => {
-	const auth = repo.token
-		? `Basic ${Buffer.from(repo.token).toString('base64')}`
-		: '';
+	const auth =
+		repo.token == null
+			? ''
+			: // legacy `username:token` Basic authentication.
+				// TODO: Consider dropping in the next major
+				/^\w+:\w+$/.test(repo.token)
```
otaviojacobi, line 133: I am slightly concerned with this match e.g. it won't match the `-` characters and several others. Do we have the correct regex to match our db format for usernames in the db?

PR #2113, opened 2025-10-02
Mark images for deletion on registry

joshbwlng:
Change-type: minor

---

See: https://balena.fibery.io/Work/Project/Mark-associated-resources-for-deletion-when-a-release-is-deleted-1779


src/features/images/tasks/delete-registry-images.ts
```diff
@@ -0,0 +1,174 @@
...
+// Make an API call to the registry service to mark images for deletion on next garbage collection
+async function markForDeletion(repo: string, hash: string) {
+	// Generate an admin-level token with delete permission
+	const token = generateToken('admin', REGISTRY2_HOST, [
+		{
+			name: repo,
+			type: 'repository',
+			actions: ['delete'],
+		},
+	]);
```
joshbwlng: We may instead want to generate a wildcard token with delete access to all repos

src/features/images/tasks/delete-registry-images.ts
```diff
@@ -0,0 +1,176 @@
...
+				items: {
+					type: 'string',
+				},
+				maxItems: 2,
+				minItems: 2,
+			},
+		},
+	},
+	required: ['images'],
+};
```
otaviojacobi: @joshbwlng just for my own information, is there any specific reason you chose this schema instead of

```suggestion
const schema = {
	type: 'object',
	properties: {
		images: {
			type: 'array',
			items: {
				type: 'object',
				properties: {
					repo: { type: 'string' },
					hash: { type: 'string' },
				},
				required: ['repo', 'hash'],
				additionalProperties: false,
			},
		},
	},
	required: ['images'],
};
```
where the properties have explicit naming? (slightly easier to read imo)

src/features/images/tasks/delete-registry-images.ts
```diff
@@ -0,0 +1,176 @@
...
+			name: repo,
+			type: 'repository',
+			actions: ['delete'],
+		},
+	]);
+
+	// Need to make requests one image at a time, no batch endpoint available
+	for (let retries = 0; retries < RATE_LIMIT_RETRIES; retries++) {
+		const [{ statusCode, statusMessage, headers }] = await requestAsync({
+			url: `https://${REGISTRY2_HOST}/v2/${repo}/manifests/${hash}`,
```
otaviojacobi: Are all repo/hashes we have here guaranteed to belong to our own registry or could these be images on custom customer registry?

src/features/images/hooks/delete-registry-images.ts
```diff
@@ -0,0 +1,77 @@
...
+
+			// Get information required to mark the image for deletion on
+			// the registry before the record is deleted from our database.
+			const images = await api.get({
+				resource: 'image',
+				options: {
+					$select: ['is_stored_at__image_location', 'content_hash'],
+					$filter: {
+						id: { $in: affectedIds },
+						content_hash: { $ne: null },
```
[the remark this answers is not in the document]
joshbwlng: `content_hash` is the "reference" value needed to make the call to the registry's delete endpoint: https://distribution.github.io/distribution/spec/api/#deleting-an-image

src/features/images/hooks/delete-registry-images.ts
```diff
@@ -0,0 +1,77 @@
...
+						resource: 'task',
+						passthrough: { req: permissions.root, tx },
+						body: {
+							is_executed_by__handler: 'delete_registry_images',
+							is_executed_with__parameter_set: {
+								images: chunk,
+							} satisfies DeleteRegistryImagesTaskParams,
+							is_scheduled_to_execute_on__time: new Date(
+								Date.now() +
+									ASYNC_TASK_DELETE_REGISTRY_IMAGES_OFFSET_SECONDS * 1000,
```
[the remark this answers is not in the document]
joshbwlng: No, updating to `ASYNC_TASK_DELETE_IMAGES_OFFSET_MS`, thanks.

src/index.ts
```diff
@@ -178,6 +178,7 @@ export * as redis from './infra/redis/index.js';
 export * as scheduler from './infra/scheduler/index.js';
 export * as cache from './infra/cache/index.js';
 export * as config from './lib/config.js';
+export * as registry from './lib/registry.js';
```
joshbwlng: Exporting the utility to make delete calls to the registry so we can reuse it to delete deltas from the registry in balena-api

Page- (approved):

src/features/images/hooks/delete-registry-images.ts
```diff
@@ -0,0 +1,88 @@
...
+					images.map((image) => [
+						image.is_stored_at__image_location,
+						image.content_hash!,
+						image.dockerfile != null
+							? dockerParser
+									.parse(image.dockerfile, {
+										includeComments: false,
+									})
+									.filter((e) => e.name.toUpperCase() === 'FROM').length
+							: 0,
```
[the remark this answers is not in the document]
joshbwlng: When the dockerfile isn't stored we have no way to figure out how many stage cache images are expected to exist. Like I mentioned above we could blindly make deletion attempts for cache images until the registry returns a 404.

src/features/images/tasks/delete-registry-images.ts
```diff
@@ -0,0 +1,138 @@
...
+					// image.is_stored_at__image_location
+					{ type: 'string' },
+					// image.content_hash
+					{ type: 'string' },
+					// number of stages in image.dockerfile
+					{ type: 'number', minimum: 0 },
+				],
+				additionalItems: false,
+				maxItems: 3,
+				minItems: 3,
```
Page-: Any reason for this to be a fixed length 3 array rather than an object?

src/features/images/hooks/delete-registry-images.ts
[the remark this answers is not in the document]
Page-: I think we do need to ensure any image that is created can also be deleted or otherwise it's problematic

src/features/images/tasks/delete-registry-images.ts
joshbwlng: @Page- The only reason is to save a bit of space per task record in the DB. May be an over optimization. If you'd prefer I change it to an object I'm happy to do that - I do like the improved typings and readability the object approach provides.

src/features/images/hooks/delete-registry-images.ts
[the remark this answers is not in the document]
joshbwlng: @Page- I guess you're referring to if an image record doesn't have a dockerfile? The only other approach I can think of is to loop through -0, -1, etc suffixes until a registry delete API call returns a 404. I had written this before and I think it can work. I'm open to other ideas though.

src/features/images/tasks/delete-registry-images.ts
Page-: Yeah I wouldn't optimize for that space savings unless it turns out to be a real problem - in which case the savings from object -> array probably isn't going to make or break it anyway

src/features/images/hooks/delete-registry-images.ts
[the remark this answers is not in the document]
Page-: We mostly just need to ensure they get deleted, that could also be done via changing how we track images in the first place such that the image cannot be created without the API explicitly knowing about it. It's not a blocker for this PR per se as the PR still improves things but this won't be sufficient alone to ensure all images that should be deleted are deleted without further work, and so no assumptionns can be made that all deleted images are also deleted in the registry

src/features/images/tasks/delete-registry-images.ts
```diff
@@ -0,0 +1,145 @@
...
+		}
+
+		// Delete cache images first so they aren't orphaned if the task
+		// fails after deleting the main image.
+		const cacheRepos = await s3Client.listCacheRepos(repo);
+		for (const cacheRepo of cacheRepos) {
+			const cacheDigests = await s3Client.listRepoDigests(cacheRepo);
+			for (const digest of cacheDigests) {
+				await deleteImage(subject, cacheRepo, digest);
+			}
```
Page-, line 147: Was there an intentional choice of serial vs parallel for these deletions?

src/features/registry/registry.ts
```diff
@@ -711,3 +720,125 @@ const getSubject = async (
...
+		: undefined;
+
+// Make an API call to the registry service to mark images for deletion on next garbage collection
+export async function deleteImage(
+	subject: string,
+	repo: string,
+	digest: string,
+) {
+	// Generate a new token for each attempt to ensure we don't run into expiry issues
+	const registryToken = generateToken(subject, REGISTRY2_HOST, [
```
Page-: I think these token generations are somewhat cpu expensive (cryptography), enough that when deleting 100s of images (multiple services * cache+release) then I expect it could add up to being noticeable

src/features/registry/registry.ts
```diff
@@ -711,3 +720,125 @@ const getSubject = async (
...
+	}
+
+	// List all multi-stage build cache image repositories for a given repository.
+	async listCacheRepos(repo: string) {
+		const cacheRepos: string[] = [];
+		const reposPath = `${this.rootPath}/repositories/`;
+		for (const child of await this.listChildPrefixes(`${reposPath}${repo}-`)) {
+			cacheRepos.push(child.replace(reposPath, '').replace(/\/$/, ''));
+		}
+		return cacheRepos;
```
Page-: This might be better as a map:
```suggestion
		const reposPath = `${this.rootPath}/repositories/`;
		return (await this.listChildPrefixes(`${reposPath}${repo}-`)).map((child) =>
			child.replace(reposPath, '').replace(/\/$/, '');
		);
```
```

File: src/features/images/tasks/delete-registry-images.ts (hunk 9 above)
Comment (joshbwlng) [reply src/features/images/tasks/delete-registry-images.ts:147]: It was mostly to avoid potentially overloading the registry with too many requests at once.

File: src/features/registry/registry.ts (hunk 10 above)
Comment (joshbwlng) [reply src/features/registry/registry.ts]: I updated the logic to instead generate a delete token per repo, instead of generating a delete token per image. Thanks for catching that.

File: src/features/images/tasks/delete-registry-images.ts (hunk 9 above)
Comment (Page-) [reply src/features/images/tasks/delete-registry-images.ts:147]: It might be worth adding a comment as to the reasoning, and that it's a guess rather than a tested thing

File: src/features/registry/registry.ts [hunk 12]
```diff
@@ -4,10 +4,12 @@
 import type { Request, RequestHandler } from 'express';
 import jsonwebtoken from 'jsonwebtoken';
 import _ from 'lodash';
+import AWS from '../device-types/storage/aws-sdk-wrapper.js';
```
Page-, line 7: I would really love if new aws code was using aws sdk v3 but I know the mocking situation will need a rethink

Page- (approved):

src/features/registry/registry.ts
joshbwlng: I agree, and I also want to move the wrapper to a more generic location. I'll come back to this in a follow-up.

src/features/images/tasks/delete-registry-images.ts
```diff
@@ -0,0 +1,158 @@
...
+	// should be enforced as unique at the database level, but just in case
+	const filters = images.map((image) => ({
+		is_stored_at__image_location: image.location,
+	}));
+	const stillReferenced = await api.resin.get({
+		resource: 'image',
+		passthrough: { req: permissions.rootRead },
+		options: {
+			$select: ['is_stored_at__image_location'],
+			$filter: filters.length === 1 ? filters[0] : filters,
```
thgreasi: How about a `$in`?

src/features/images/tasks/delete-registry-images.ts
```diff
@@ -0,0 +1,158 @@
...
+		},
+	});
+
+	// Define what images are actually safe to delete
+	const safeToDelete = images.filter(
+		(image) =>
+			!stillReferenced.some(
+				(ref) => ref.is_stored_at__image_location === image.location,
+			),
```
thgreasi: How about making a Set out of the `stillReferenced/is_stored_at__image_location`s so that this becomes O(N) :smiling_imp:

test/test-lib/api-helpers.ts
```diff
@@ -338,6 +347,23 @@ const expectTasks = async (
 				task.is_executed_with__parameter_set.devices.sort((a, b) => a - b);
 			}
 		}
+	} else if (handler === 'delete_registry_images') {
+		const sortImages = (images: DeleteRegistryImagesTaskParams['images']) =>
+			images.sort((a, b) => a.location.localeCompare(b.location));
+		for (const task of actual) {
+			if (
+				isDeleteRegistryImageTaskParam(task.is_executed_with__parameter_set)
```
thgreasi, line 355: Maybe we should be throwing when this (and all other isSomethingTaskParam) returns false

joshbwlng: Agreed, let's do that in a follow-up

test/26_registry-image-deletion.ts
```diff
@@ -0,0 +1,384 @@
...
+			delete ctx.fixtures.applications['app1'];
+			registryMock.reset();
+			await fixtures.clean(ctx.fixtures);
+		});
+
+		it('should mark for deletion when an image is deleted', async () => {
+			const image = await createImage(ctx.service1.id);
+			await api.resin.delete({
+				resource: 'image',
+				passthrough: { req: permissions.root },
```
thgreasi: I guess it would have been better if we used user credentials (via a pineTest pineUser) for all these, so that it's more realistic permissions-wise.
Ie a user has enough permissions for all things the hook needs to do.

test/26_registry-image-deletion.ts
```diff
@@ -0,0 +1,384 @@
...
+				consoleSpy.restore();
+			} finally {
+				config.TEST_MOCK_ONLY.ASYNC_TASK_DELETE_REGISTRY_IMAGES_MAX_TIME_MS =
+					originalMaxTime;
+				await resetLatestTaskIds('delete_registry_images');
+			}
+		});
+
+		it('should mark for deletion when images are deleted via a cascade', async () => {
+			const { images } = await createRelease(ctx.app1.id, ctx.service1.id);
```
thgreasi: How about adding an extra release here?

test/26_registry-image-deletion.ts
```diff
@@ -0,0 +1,384 @@
...
+
+			await waitFor({
+				delayMs: 500,
+				checkFn: () =>
+					imageA.registryImage.isDeleted === true && imageB.isDeleted === true,
+			});
+			await expectSettledTasks([imageA]);
+		});
+
+		it('should mark for deletion when multiple images are deleted', async () => {
```
thgreasi: How about changing this to deleting a release w/ 2 images, to be more realistic of how this case happens?

thgreasi (approved):

PR #2226, opened 2026-03-25
Add `/device/v3/tags` endpoint

Page-:
https://balena.fibery.io/Work/Project/Supervisor-API-for-device-tags-1031

Related: https://github.com/balena-os/balena-supervisor/pull/2501

Change-type: minor

### DIFF
--- a/src/features/device-state/index.ts
+++ b/src/features/device-state/index.ts
@@ -12,6 +12,7 @@ import type { StatePatchV3Body } from './routes/state-patch-v3.js';
 import { resolveDeviceUuids, statePatchV3 } from './routes/state-patch-v3.js';
 import { fleetStateV3 } from './routes/fleet-state-get-v3.js';
 import type { Device } from '../../balena-model.js';
+import { tagsPatchV3 } from './routes/tags-patch-v3.js';
 
 export {
 	getStateEventAdditionalFields,
@@ -63,6 +64,14 @@ export const setup = (app: Application) => {
 		'/device/v3/fleet/:fleetUuid/state',
 		middleware.authenticated,
 		fleetStateV3,
+	);
+	app.patch(
+		'/device/v3/tags',
+		resolveOrDenyDevicesWithStatus(401, (req) =>
+			resolveDeviceUuids(req.body as StatePatchV3Body),
+		),
+		middleware.authenticatedApiKey,
+		tagsPatchV3,
 	);
 };
 
--- a/src/features/device-state/routes/tags-patch-v3.ts
+++ b/src/features/device-state/routes/tags-patch-v3.ts
@@ -1,0 +1,194 @@
+import type { RequestHandler } from 'express';
+
+import {
+	captureException,
+	handleHttpErrors,
+} from '../../../infra/error-handling/index.js';
+import { sbvrUtils, errors } from '@balena/pinejs';
+import { getIP } from '../../../lib/utils.js';
+import type { DeviceTag } from '../../../balena-model.js';
+import type { ResolveDeviceInfoCustomObject } from '../middleware.js';
+import { checkTagKeyValidity } from '../../tags/validation.js';
+import { resolveDeviceUuids } from './state-patch-v3.js';
+
+const { BadRequestError, UnauthorizedError } = errors;
+const { api } = sbvrUtils;
+
+/**
+ * These typings should be used as a guide to what should be sent, but cannot be trusted as what actually *is* sent.
+ */
+export type TagsPatchV3Body = {
+	[uuid: string]: {
+		[tagKey: string]: string;
+	};
+};
+
+const fetchData = async (
+	req: Express.Request,
+	custom: AnyObject,
+	deviceIds: number[],
+	deviceTagKeys: {
+		[deviceUuid: string]: string[];
+	},
+) =>
+	await sbvrUtils.db.readTransaction(async (tx) => {
+		const resinApiTx = api.resin.clone({
+			passthrough: { req, custom, tx },
+		});
+		const devices = await resinApiTx.get({
+			resource: 'device',
+			options: {
+				$select: ['id', 'uuid'],
+				$filter: {
+					id: { $in: deviceIds },
+				},
+			},
+		} as const);
+		if (devices.length !== deviceIds.length) {
+			throw new UnauthorizedError();
+		}
+
+		const tagsByDevice: {
+			[deviceUuid: string]: {
+				id: number;
+				tags: Record<DeviceTag['Read']['tag_key'], DeviceTag['Read']['value']>;
+			};
+		} = {};
+		for (const { id, uuid } of devices) {
+			tagsByDevice[uuid] = {
+				id,
+				tags: Object.fromEntries(
+					(
+						await resinApiTx.get({
+							resource: 'device_tag',
+							options: {
+								$select: ['tag_key', 'value'],
+								$filter: {
+									device: id,
+									tag_key: { $in: deviceTagKeys[uuid] },
+								},
+							},
+						} as const)
+					).map(({ tag_key, value }) => [tag_key, value]),
+				),
+			};
+		}
+
+		return tagsByDevice;
+	});
+
+export const tagsPatchV3: RequestHandler = async (req, res) => {
+	try {
+		const body = req.body as TagsPatchV3Body;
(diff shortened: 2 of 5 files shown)


src/features/device-state/routes/tags-patch-v3.ts
```diff
@@ -0,0 +1,174 @@
...
+			resource: 'device',
+			options: {
+				$select: ['id', 'uuid'],
+				$filter: {
+					id: { $in: deviceIds },
+				},
+			},
+		} as const);
+		if (devices.length !== deviceIds.length) {
+			throw new UnauthorizedError();
```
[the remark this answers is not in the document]
Page-: That's a pretty niche edge case but also 401 is what we do for security reasons and also as it stands I think it would currently only ever be an array of 1 device (and if that 1 is deleted I think it'll bail earlier, but even if not then the 401 would still be good for consistency)

src/features/device-state/routes/tags-patch-v3.ts
```diff
@@ -0,0 +1,194 @@
...
+		}
+
+		const tagsByDevice: {
+			[deviceUuid: string]: {
+				id: number;
+				tags: Record<DeviceTag['Read']['tag_key'], DeviceTag['Read']['value']>;
+			};
+		} = {};
+		for (const { id, uuid } of devices) {
+			// for (const uuid of Object.keys(deviceTagKeys)) {
```
joshbwlng: Leftover comment?

src/features/device-state/index.ts
```diff
@@ -64,6 +65,14 @@ export const setup = (app: Application) => {
 		middleware.authenticated,
 		fleetStateV3,
 	);
+	app.patch(
+		'/device/v3/tags',
+		resolveOrDenyDevicesWithStatus(401, (req) =>
+			resolveDeviceUuids(req.body as StatePatchV3Body),
```
joshbwlng, line 71: Is this meant to be using the `resolveDeviceUuids()` exported from `tags-patch-v3.ts`?

src/features/device-state/routes/tags-patch-v3.ts
```diff
@@ -0,0 +1,194 @@
...
+		const updateFns: Array<
+			(resinApiTx: typeof sbvrUtils.api.resin) => Promise<void>
+		> = [];
+
+		for (const uuid of uuids) {
+			const tags = body[uuid];
+			const tagKeys = deviceTagKeys[uuid];
+			if (tagKeys.length === 0) {
+				continue;
+			}
```
joshbwlng, line 133: If `deviceTagKeys[uuid]` is `undefined`, won't `tagKeys.length` throw?

src/features/device-state/index.ts
Page-: No, I think the idea was that it should become shared code but I didn't leave a TODO/forgot - I'll need to fix that before merge

src/features/device-state/routes/tags-patch-v3.ts
```diff
@@ -0,0 +1,191 @@
...
+					checkTagKeyValidity(tagKey);
+					if (typeof body[uuid][tagKey] !== 'string') {
+						throw new BadRequestError('Tag values must be strings.');
+					}
+				}
+
+				deviceTagKeys[uuid] = tagKeys;
+			}
+		}
+		const tagsByDevice = await fetchData(
```
joshbwlng, line 117: Just in case maybe it would make sense to return early if `deviceTagKeys` is empty before calling `fetchData()`.

Page-: Done, although I treated it as an error

joshbwlng (approved):

src/features/device-state/routes/tags-patch-v3.ts
```diff
@@ -0,0 +1,194 @@
...
+				id: number;
+				tags: Record<DeviceTag['Read']['tag_key'], DeviceTag['Read']['value']>;
+			};
+		} = {};
+		for (const { id, uuid } of devices) {
+			tagsByDevice[uuid] = {
+				id,
+				tags: Object.fromEntries(
+					(
+						await resinApiTx.get({
```
thgreasi, line 62: Was there any reason that we didn't prefer to just $expand from query that fetches the device(s) above?
That would benefit from the `/canAccess()` permission chain as well, while the separate query has to do the join w/ the device again.
If so, then I would love a comment so that we remember the reason.

src/features/device-state/routes/tags-patch-v3.ts
```diff
@@ -0,0 +1,194 @@
...
+							resource: 'device_tag',
+							options: {
+								$select: ['tag_key', 'value'],
+								$filter: {
+									device: id,
+									tag_key: { $in: deviceTagKeys[uuid] },
+								},
+							},
+						} as const)
+					).map(({ tag_key, value }) => [tag_key, value]),
```
thgreasi: Would it be that bad (memory wise) if we used the pine result object instead of allocating N more arrays?

src/features/device-state/routes/tags-patch-v3.ts
thgreasi: I guess the reason was so that we can $filter the tags of each device independently.
But since we no longer have dependent devices and we atm expect all devices to be patching just their own tags , we could be doing a single `.get` per device w/ a device_tag `$expand`.

src/features/device-state/routes/tags-patch-v3.ts
```diff
@@ -0,0 +1,194 @@
...
+	deviceIds: number[],
+	deviceTagKeys: {
+		[deviceUuid: string]: string[];
+	},
+) =>
+	await sbvrUtils.db.readTransaction(async (tx) => {
+		const resinApiTx = api.resin.clone({
+			passthrough: { req, custom, tx },
+		});
+		const devices = await resinApiTx.get({
```
thgreasi, line 38: Since `resolveOrDenyDevicesWithStatus` already has both the id & uuid of each device, wdyt about something like v, so that `fetchData()` only has to fetch the `device_tags`?
See: https://github.com/balena-io/open-balena-api/pull/2253

or if we still prefer to re-fetch all devices in a single query, we could do:
```ts
		const deviceTagFilters = deviceIds.map((id) => ({
			device: id,
			tag_key: { $in: deviceTagKeysByDeviceId[id] },
		}));
		const devices = await resinApiTx.get({
			resource: 'device',
			options: {
				$select: ['id', 'uuid'],
				$expand: {
					device_tag: {
						$select: ['tag_key', 'value'],
						$filter: deviceTagFilters.length === 1
							? deviceTagFilters[0]
							: { $or: deviceTagFilters },
					}
				},
				$filter: {
					id: { $in: deviceIds },
				},
			},
		} as const);
```

test/03_device-state.ts
```diff
@@ -2077,5 +2077,115 @@ export default () => {
...
+							},
+						);
+					}
+				});
+
+				it('should update multiple device tags, leaving others intact', async () => {
+					const devicePatchBody = {
+						[stateKey]: {
+							tagKey2: 'updatedTagValue2',
+							tagKey3: 'updatedTagValue3',
```
thgreasi, line 2164: Can we also add a new tag here so that we can test that we can both update existing & insert new ones w/ the same request?

thgreasi (approved):

src/features/device-state/routes/tags-patch-v3.ts
```diff
@@ -0,0 +1,194 @@
...
+const { api } = sbvrUtils;
+
+/**
+ * These typings should be used as a guide to what should be sent, but cannot be trusted as what actually *is* sent.
+ */
+export type TagsPatchV3Body = {
+	[uuid: string]: {
+		[tagKey: string]: string;
+	};
+};
```
thgreasi, line 24: Would it make sense to wrap the tags in an extra object, just in case we later want to add more stuff in this endpoint (eg what if users later also want a way to also delete keys)
```suggestion
export type TagsPatchV3Body = {
	[uuid: string]: {
		PUT: { // or upsert or something
			[tagKey: string]: string;
		};
		// eg: maybe we also add DELETE in the future
	};
};
```

Page-: I was actually thinking an explicit `null` could work for removing tags if we decide to add that

PR #2242, opened 2026-04-03
Test OS updates from and to draft OS releases

thgreasi:
Change-type: patch
See: https://balena.fibery.io/Work/Project/1704


kb2ma (approved):
Walked through code with Thodoris. LGTM.

PR #2255, opened 2026-04-20
target-hostapp: Move the downgrade check to the POSTRUN phase

thgreasi:
Change-type: patch
See: https://balena.fibery.io/Work/Project/1704

### DIFF
--- a/test/15_target-hostapp.ts
+++ b/test/15_target-hostapp.ts
@@ -483,8 +483,7 @@ export default () => {
 					})
 					.expect(
 						400,
-						// TODO: This should ideally be: '"It is necessary that each release that should operate a device, has a status that is equal to \\"success\\"."'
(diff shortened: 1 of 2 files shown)


test/15_target-hostapp.ts
```diff
@@ -722,6 +721,41 @@ export default () => {
 						);
 				});
 
+				it(`should block queuing a downgrade when the device.should_be_operated_by__release is updated alogn with os_version & os_variant (when downgrading ${higherTitle} -> ${lowerTitle})`, async () => {
```
joshbwlng:
```suggestion
				it(`should block queuing a downgrade when the device.should_be_operated_by__release is updated along with os_version & os_variant (when downgrading ${higherTitle} -> ${lowerTitle})`, async () => {
```

joshbwlng (approved):

PR #2256, opened 2026-04-20
patch: Optional retry logic for requestAsync

ab77:
Adds exponential backoff retry to requestAsync with default 0 retries. Adds a single retry to device proxy, as a workaround for dev. env. virtual devices sometimes missing the initial target state notification.

https://balena.fibery.io/Work/Project/Capture-data-to-build-the-balena-reliability-pipeline-1422

### DIFF
--- a/src/infra/request-promise/index.ts
+++ b/src/infra/request-promise/index.ts
@@ -8,6 +8,28 @@ const defaultRequest = request.defaults({
 	timeout: EXTERNAL_HTTP_TIMEOUT_MS,
 });
 
+const delay = (ms: number) => new Promise((res) => setTimeout(res, ms));
+
+export const withRetry = async <T>(
+	fn: () => Promise<T>,
+	retries: number,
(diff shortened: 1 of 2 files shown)


src/infra/request-promise/index.ts
```diff
@@ -8,6 +8,28 @@ const defaultRequest = request.defaults({
...
+		return await fn();
+	} catch (error) {
+		if (retries <= 0) {
+			throw error;
+		}
+
+		const waitTime = baseDelay * Math.pow(2, attempt);
+
+		await delay(waitTime);
+		return withRetry(fn, retries - 1, baseDelay, attempt + 1);
```
otaviojacobi, line 29: Can we go with a iterative version rather than recrusive? In general less problematic and better usage of memory

ab77:
instead: https://github.com/balena-io/open-balena-vpn/pull/679

PR #2260, opened 2026-04-23
Improve typing of `getBodyOrQueryParam`

Page-:
https://balena.fibery.io/favorites/Work/Project/1002

Change-type: patch

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -143,8 +143,20 @@ export const groupByMap = <K, V>(entries: V[], iteratee: (item: V) => K) => {
 	return result;
 };
 
-export const getBodyOrQueryParam = (
-	req: Request,
-	paramName: string,
-	defaultValue?: string,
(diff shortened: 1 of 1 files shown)


src/lib/utils.ts
```diff
@@ -143,8 +143,20 @@ export const groupByMap = <K, V>(entries: V[], iteratee: (item: V) => K) => {
...
 
-export const getBodyOrQueryParam = (
-	req: Request,
-	paramName: string,
-	defaultValue?: string,
-) => req.body[paramName] ?? req.query[paramName] ?? defaultValue;
+export const getBodyOrQueryParam = <
+	R extends Request,
+	P extends (keyof R['body'] | keyof R['query']) & string,
+	D = undefined,
```
thgreasi, line 149: I would be also fine if we preserved the defaultValue only be a `string` but not a blocker given your TODO
```suggestion
	D extends string | undefined = undefined,
```

thgreasi (approved):

PR #2264, opened 2026-05-01
Mark registry images for deletion in parallel

joshbwlng:
Change-type: patch

---

See: https://balena.fibery.io/Work/Project/Mark-associated-resources-for-deletion-when-a-release-is-deleted-1779

After some testing on staging I found that we are only executing around 42-44 image deletion calls within the default 30 seconds the task is allowed to run for. S3 allows for 3,500 DELETE requests per minute so we should be safe to increase throughput.


src/features/images/tasks/delete-registry-images.ts
```diff
@@ -98,65 +139,51 @@ const deleteRegistryImages = async ({
 	// Mark images and any of their multi-stage cache images for deletion
 	// Don't let the task run for too long, and create a new task with
 	// the remaining image data if it does
-	let totalImagesDeleted = 0;
-	for (let i = 0; i < safeToDelete.length; i++) {
-		const { location } = safeToDelete[i];
+	const s3 = s3Client;
+	let processedCount = 0;
+	let totalManifestsDeleted = 0;
+	while (processedCount < safeToDelete.length) {
```
Page-: I think this would be much better approached using a library like https://github.com/sindresorhus/p-queue so something more like:
```ts
const queue = = new PQueue({
	concurrency: ASYNC_TASK_DELETE_REGISTRY_IMAGES_CONCURRENCY // With the value set according to the overall concurrency per process rather than per task, still not perfect but still a lot more consistent
});
...
const controller = AbortSignal.timeout(ASYNC_TASK_DELETE_REGISTRY_IMAGES_MAX_TIME_MS);
await Promise.map(safeToDelete, async ({ location }) => {
	await queue.add(({ signal }) => {
		// run delete
		...

		// maybe cancel the in-progress on abort?
		signal.addEventListener('abort', () => {
			// cancel
		});
	}, {
		signal: controller.signal
	})
})
```

This has the following benefits imo:
* the tasks can be added to a global queue per-process for more consistent parallelism than if done per-task (although isn't a requirement and equally could stick with a separate queue per task if there is an argument for that which I'm missing)
* the concurrency is maintained at a consistent N vs the current implementation which is spikey with it starting at N and then decreasing as each job completes until it hits 0 at which point it spikes back up to N and starts decreasing again
* it handles aborting based on time more smoothly with the abort controller/signals, allowing not yet started tasks to be removed from the queue immediately and in-progress ones the possibility of being cancelled partway through (which should be fine as we in general should be able to retry/restart wherever between temporary failures/deploys/etc)

The main issue I can see is the creation of a new task with a subset of images; however I suspect that should be fine as we should be able to pick up/continue from wherever and so keeping a full list wouldn't break anything, and for re-runs (or first runs) with already deleted images I'd expect it to be very quick to confirm they're already deleted and move on, and as such not a big deal.

Also there is some rate limiting support via `interval`/`intervalCap` but I suspect that is overkill for what we need; it does open up that possibility though if we wanted to limit more based on number of calls than on concurrency (I suspect any underlying limits would be concurrency based rather than request based so wouldn't help, but also potentially not)

joshbwlng: @Page- Thanks for the suggestion, I wasn't aware of `p-queue`. I've added a new commit for switching to `p-queue` - it appears to be working and passing tests. The only concern I have is if having this queue in memory could potentially add too much extra memory pressure. I do really prefer the per-process concurrency control though. There is the `.onSizeLessThan(limit)` option that we might could implement if we're worried about it growing too large. wdyt?

src/lib/config.ts
```diff
@@ -760,6 +764,12 @@ export const TEST_MOCK_ONLY = {
 		guardTestMockOnly();
 		PINEJS_QUEUE_INTERVAL_MS = v;
 	},
+	set ASYNC_TASK_DELETE_REGISTRY_IMAGES_CONCURRENCY(
+		v: typeof ASYNC_TASK_DELETE_REGISTRY_IMAGES_CONCURRENCY,
+	) {
+		guardTestMockOnly();
+		ASYNC_TASK_DELETE_REGISTRY_IMAGES_CONCURRENCY = v;
+	},
```
thgreasi: Since this is a ctor argument in p-queue, I guess that there is no way to modify it during runtime in tests (and we are also not using it am afaict), so I would argue to remove it, to avoid any confusion if someone later tries to use it and see that it's not working.
```suggestion
```

test/26_registry-image-deletion.ts
```diff
@@ -277,31 +277,33 @@ export default () => {
...
+							checkFn: () => checkIsDeleted(images),
+						});
+
+						sinon.assert.calledWithMatch(
+							consoleSpy,
+							sinon.match(
+								'[delete_registry_images_task] Task took too long. Created a new task for the remaining images',
+							),
+						);
+					} finally {
```
thgreasi: NIT: I guess you could also combine the two subsequent `finaly`s?

src/features/images/tasks/delete-registry-images.ts
Page-: The entries would be removed on success or timing out, and I wouldn't expect it to use much more memory than just having the tasks running via some other form so I'm not worried about extra memory use from p-queue - however if there were a lot of concurrent tasks running at once then potentially that could be an issue (but that's a separate standalone thing)

src/features/images/tasks/delete-registry-images.ts
```diff
@@ -91,72 +119,70 @@ const deleteRegistryImages = async ({
...
-					} satisfies DeleteRegistryImagesTaskParams,
-					attempt_limit: ASYNC_TASK_ATTEMPT_LIMIT,
+	const errorController = new AbortController();
+	const signal = AbortSignal.any([
+		AbortSignal.timeout(ASYNC_TASK_DELETE_REGISTRY_IMAGES_MAX_TIME_MS),
+		errorController.signal,
+	]);
+	let totalManifestsDeleted = 0;
+	await Promise.allSettled(
+		[...remaining].map((location) =>
```
Page-: I prefer using the iterator functions now, ie
```suggestion
		remaining.values().map((location) =>
```
It avoids the need to create an intermediate array and will also result in only iterating the entries once (vs twice, once to create the intermediate array and then once to do the actual map)

src/features/images/tasks/delete-registry-images.ts
```diff
@@ -91,72 +119,70 @@ const deleteRegistryImages = async ({
...
-						images: safeToDelete.slice(i),
-					} satisfies DeleteRegistryImagesTaskParams,
-					attempt_limit: ASYNC_TASK_ATTEMPT_LIMIT,
+	const errorController = new AbortController();
+	const signal = AbortSignal.any([
+		AbortSignal.timeout(ASYNC_TASK_DELETE_REGISTRY_IMAGES_MAX_TIME_MS),
+		errorController.signal,
+	]);
+	let totalManifestsDeleted = 0;
+	await Promise.allSettled(
```
Page-: What's the reasoning behind `Promise.allSettled` with the `throw errorController.signal.reason` vs `Promise.all` throwing directly?

joshbwlng: We could do it with `Promise.all`, but we'd need to wrap it in a try/catch and inspect the error to know if its from an expected timeout (`ASYNC_TASK_DELETE_REGISTRY_IMAGES_MAX_TIME_MS`), or an actual error from the registry/etc - I don't want the underlying Pine `task` attempt to be marked as failed just because it reached it's `MAX_TIME_MS`. I went with `Promise.allSettled` combined with a separate `errorController` channel so we can clearly know if it was a timeout or an actual error. I also updated `deleteRepo()` to accept `signal` and throw early if the signal has been aborted, so we can stop any other in-flight work earlier.

thgreasi: @Page- @joshbwlng does this look more like how you envision it?
```ts
	const errorController = new AbortController();
	const timeoutAbortSignal = AbortSignal.timeout(ASYNC_TASK_DELETE_REGISTRY_IMAGES_MAX_TIME_MS);
	const signal = AbortSignal.any([
		timeoutAbortSignal,
		errorController.signal,
	]);
	let totalManifestsDeleted = 0;
	try {
		await Promise.all(
			remaining.values().map((location) =>
				queue.add(
					async () => {
						const repo = location.replace(/^[^/]+\//, '');
						if (repo === '') {
							console.warn(
								`[${logHeader}] Skipping deletion of image with empty repo: ${location}`,
							);
						} else {
							const manifestsDeleted = await deleteRepo(s3Client!, repo, signal);
							totalManifestsDeleted += manifestsDeleted;
						}
						remaining.delete(location);
					},
					{ signal },
				),
			),
		);
	} catch (err) {
		// In all cases other than a timeout error, we fail the task
		if (
			!timeoutAbortSignal.aborted ||
			timeoutAbortSignal.reason !== err
		) {
			errorController.abort(err);
			throw err;
		}
	}
```
PS: I mean w/o any  `if (errorController.signal.aborted) { throw errorController.signal.reason; }` later.

joshbwlng: I think that's what we'd need if we wanted to go with `Promise.all`, but since we already have the `errorController` I think we can just use that and avoid the extra nesting in the try/catch if we stick with `Promise.allSettled`.

thgreasi: I would not block on this, but
* I think it's the same nesting level since I just moved the try...catch outside
* I feel that one I posted above w/ the `Promise.all` might be slightly more accurate since
  * we now queue a new task only when in case of a timeout error and in all other cases throw - handling the special expected case (which is exactly what we said we want to do), instead of handling the case that a specific range of code threw
  * we also remove the `location` from the `remaining` even when `repo === ''`
  * it probably doesn'y matter but fwiw: we now call the `errorController.abort()` only once

thgreasi (approved):

src/features/images/tasks/delete-registry-images.ts
joshbwlng: I'm happy with either approach, maybe @Page- can be the tie breaker vote.

> we now queue a new task only when in case of a timeout error and in all other cases throw - handling the special expected case (which is exactly what we said we want to do), instead of handling the case that a specific range of code threw

I think this is true for my implementation as well?

src/features/images/tasks/delete-registry-images.ts
```diff
@@ -91,72 +122,77 @@ const deleteRegistryImages = async ({
...
+	const timeoutAbortSignal = AbortSignal.timeout(
+		ASYNC_TASK_DELETE_REGISTRY_IMAGES_MAX_TIME_MS,
+	);
+	const signal = AbortSignal.any([timeoutAbortSignal, errorController.signal]);
+	let totalManifestsDeleted = 0;
+	try {
+		await Promise.all(
+			remaining.values().map((location) =>
+				queue.add(
+					async () => {
```
Page-: You can reduce the lines/indentation slightly with:
```suggestion
		await queue.addAll(
			remaining.values().map((location) =>
				async () => {
```

Page- (approved):

PR #2283, opened 2026-05-12
fix: re-export zod's `z` namespace so consumers can use `z.infer`

otaviojacobi:
Previously, the import type + re-export did not reliably carried the zod namespace

Change-type: patch


Page- (changes requested):
This doesn't actually seem to be necessary from my tests

PR #2297, opened 2026-05-13
Run the tests of each model version in parallel on CI

thgreasi:
Change-type: patch


.github/actions/test/action.yml
```diff
@@ -1,5 +1,5 @@
 # https://docs.github.com/en/actions/creating-actions/creating-a-composite-action
-name: "Docker tests"
+name: "Docker tests version scoped"
```
thgreasi, line 2: My goal was to align bC-api w/ oB-api as much as possible, so that we can have both repos work in a consistent way, and improvements on one of them can easily moved to the other.

.github/workflows/flowzone.yml
```diff
@@ -3,11 +3,15 @@ name: Flowzone
...
+    branches:
+      - "main"
+      - "master"
   # allow external contributions to use secrets within trusted code
   pull_request_target:
     types: [opened, synchronize, closed]
-    branches: [main, master]
+    branches:
+      - "main"
+      - "master"
```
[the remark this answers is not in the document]
thgreasi: yea that's how flowzone works

.github/workflows/flowzone.yml
```diff
@@ -31,5 +31,13 @@ jobs:
           "linux/amd64": ["self-hosted","X64"],
           "linux/arm64": ["self-hosted","ARM64"]
         }
+      custom_test_matrix: >
+        {
+          "value": [
+            "v6",
+            "v7",
+            "resin",
```
joshbwlng:
```suggestion
            "resin"
```

joshbwlng: Also I noticed that bC-api has an os array defined here, is that unnecessary?
```
"os": [
    ["self-hosted"]
]
```

thgreasi: I've left that out on purpose since there is a discussion on zulip on why we are not using it.
It doesn't seem to be necessary.
@otaviojacobi already has a follow-up just for the self-hosted, but this PR as it atm seems to be fixing the intermittent errors on its own.

joshbwlng (approved):

otaviojacobi:
:heart:

PR #2300, opened 2026-05-14
Switch the /device-types/v1/:deviceType/images/:version/download-size alias resolution to use the model

thgreasi:
Depends-on: https://github.com/balena-io/open-balena-api/pull/1607
Depends-on: https://github.com/balena-io/contracts/pull/480
Change-type: patch
See: https://balena.fibery.io/Work/Improvement/Stop-relying-on-device-types-v1-device-type.json-for-unrelated-things-257

### DIFF
--- a/src/features/device-types/device-types.ts
+++ b/src/features/device-types/device-types.ts
@@ -119,9 +119,8 @@ export const getImageSize = async (
 	slug: string,
 	buildId: string,
 ): Promise<number> => {
-	const deviceTypeInfo = await findDeviceTypeInfoBySlug(resinApi, slug);
-	const deviceTypeJson = deviceTypeInfo.latest;
-	const normalizedSlug = deviceTypeJson.slug;
+	const deviceType = await getDeviceTypeBySlug(resinApi, slug);
+	const normalizedSlug = deviceType.slug;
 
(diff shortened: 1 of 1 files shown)


src/features/device-types/device-types.ts
```diff
@@ -119,9 +119,8 @@ export const getImageSize = async (
 	slug: string,
 	buildId: string,
 ): Promise<number> => {
-	const deviceTypeInfo = await findDeviceTypeInfoBySlug(resinApi, slug);
-	const deviceTypeJson = deviceTypeInfo.latest;
-	const normalizedSlug = deviceTypeJson.slug;
+	const deviceType = await getDeviceTypeBySlug(resinApi, slug);
+	const normalizedSlug = deviceType.slug;
```
otaviojacobi, line 123: I see deviceType variable is only used here and for its id, wdyt of
```suggestion
const { id: dtId, slug: normalizedSlug } = await getDeviceTypeBySlug(resinApi, slug);
```

And update accordingly

thgreasi: I actually prefer it as it is :grimacing:

otaviojacobi (approved):

PR #2301, opened 2026-05-14
Add `zod` based validation for logs endpoints

Page-:

Change-type: patch

### DIFF
--- a/src/features/device-logs/lib/store.ts
+++ b/src/features/device-logs/lib/store.ts
@@ -1,4 +1,4 @@
-import type { Request, RequestHandler, Response } from 'express';
+import type { Request, Response } from 'express';
 import type { InternalDeviceLog, LogContext, SupervisorLog } from './struct.js';
 
 import onFinished from 'on-finished';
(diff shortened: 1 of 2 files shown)


src/features/device-logs/lib/store.ts
```diff
@@ -91,34 +95,48 @@ const getWriteContext = (() => {
...
+			z.object({
+				timestamp: z.number(),
+				message: z.string(),
+				isStdErr: z.boolean().optional(),
+				isSystem: z.boolean().optional(),
+				serviceId: z.number().optional(),
+				// This should never actually be sent, but if it is we just ignore the log line instead of throwing an error
+				uuid: z.string().optional(),
+			}),
+		),
```
otaviojacobi, line 110: Maybe a add a TODOor do right away but possibly we can do
```ts
export const supervisorSchema = z.object({
				timestamp: z.number(),
				message: z.string(),
				isStdErr: z.boolean().optional(),
				isSystem: z.boolean().optional(),
				serviceId: z.number().optional(),
				// This should never actually be sent, but if it is we just ignore the log line instead of throwing an error
				uuid: z.string().optional(),
			});
```

And on struct.ts SupervisorLog now becomes
```ts
export type SupervisorLog = z.infer<typeof supervisorSchema>; 
```

Page-: I think we should probably be doing more of this in future, but for the moment I think it's nice to make sure the zod schemas match with the existing expected typings

otaviojacobi (approved):

PR #2305, opened 2026-05-14
Add `zod` based validation for `/login_` endpoint

Page-:

Change-type: patch


joshbwlng (approved):

src/features/auth/login.ts
```diff
@@ -1,52 +1,62 @@
...
-					throw new BadRequestError('User not allowed to login.');
-				}
+export const login = (onLogin: SetupOptions['onLogin']) =>
+	createValidatedRequestHandler(
+		{
+			body: z.object({
+				username: z.string(),
+				password: z.string(),
+			}),
+		},
```
joshbwlng, line 22: Not requesting any changes, but I suppose we now will return a 400 on invalid request parameters whereas before we were returning a 401.

PR #2306, opened 2026-05-14
Use release assets for device types v1

otaviojacobi:
See: https://balena.fibery.io/Work/Project/Enable-balenaMachine-bok-users-to-use-release-assets-hostapp-images-2470

Change-type: patch

### DIFF
--- a/src/features/device-types/device-types-list.ts
+++ b/src/features/device-types/device-types-list.ts
@@ -18,6 +18,7 @@ import {
 	CONTRACT_ALLOWLIST,
 	IMAGE_STORAGE_DEBUG_REQUEST_ERRORS,
 } from '../../lib/config.js';
+import { ESR_MIN_MAJOR } from './device-types.js';
 
 const { api } = sbvrUtils;
 
@@ -58,6 +59,18 @@ const getFirstValidBuild = async (
 	}
 };
 
+const getDeviceTypeJsonFromAsset = async (
+	href: string,
+): Promise<DeviceTypeJson> => {
+	const response = await fetch(href);
+	if (!response.ok) {
+		throw new InternalRequestError(
+			'Failed to retrieve device type json from release asset',
+		);
+	}
+	return (await response.json()) as DeviceTypeJson;
+};
+
 const ALLOWLISTED_DT_SLUGS: string[] = [];
 for (const contractPath of CONTRACT_ALLOWLIST) {
 	const allowListedDtSlug = /^hw.device-type\/([\w-]+)$/.exec(
@@ -71,11 +84,64 @@ for (const contractPath of CONTRACT_ALLOWLIST) {
 export const getDeviceTypeJsons = multiCacheMemoizee(
 	async (): Promise<Dictionary<DeviceTypeJson>> => {
 		const result: Dictionary<DeviceTypeJson> = {};
-		let deviceTypes = await api.resin.get({
+		const deviceTypes = await api.resin.get({
 			resource: 'device_type',
 			passthrough: { req: permissions.rootRead },
 			options: {
 				$select: 'slug',
+				$expand: {
+					is_default_for__application: {
+						$top: 2,
+						$select: ['id'],
+						$expand: {
+							// TODO: this could possibly use should_be_running__release
+							// to simplify the filter/orderby semantics
+							owns__release: {
+								$top: 1,
+								$select: ['id', 'raw_version'],
+								$expand: {
+									release_asset: {
+										$select: ['asset_key', 'asset'],
+										$filter: {
+											asset_key: 'device-type.json',
+										},
+									},
+								},
+								$filter: {
+									status: 'success',
+									is_final: true,
+									is_invalidated: false,
+									$and: [
+										{ semver_major: { $lt: ESR_MIN_MAJOR } },
+										{ semver_major: { $gt: 0 } },
+									],
+								},
+								$orderby: [
+									{ semver_major: 'desc' },
+									{ semver_minor: 'desc' },
+									{ semver_patch: 'desc' },
+									{ revision: 'desc' },
+									// prefer prod over dev
+									{ variant: 'desc' },
+									{ created_at: 'desc' },
+								],
+							},
+						},
+						$filter: {
+							is_host: true,
+							$not: {
+								application_tag: {
+									$any: {
+										$alias: 'at',
+										$expr: {
+											at: { tag_key: 'release-policy' },
+										},
+									},
+								},
+							},
+						},
+					},
+				},
 				...(ALLOWLISTED_DT_SLUGS.length > 0 && {
 					$filter: {
 						slug: { $in: ALLOWLISTED_DT_SLUGS },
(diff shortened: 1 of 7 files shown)


src/features/device-types/device-types-list.ts
```diff
@@ -84,15 +148,61 @@ export const getDeviceTypeJsons = multiCacheMemoizee(
...
+				const release = dt.is_default_for__application
+					.map((app) => app.owns__release[0])
+					.find((rel) => rel?.release_asset[0]?.asset != null);
+				if (release == null) {
+					return;
+				}
+				const deviceTypeJson = await getDeviceTypeJsonFromAsset(
+					release.release_asset[0].asset!.href,
+				);
+				deviceTypeJson.buildId = release.raw_version;
```
otaviojacobi: @thgreasi do you think we really need this? This whole block of code would become much simpler without this (as we could simply just directly get the asset with a flatMap above)

src/features/device-types/device-types-list.ts
```diff
@@ -71,11 +84,62 @@ for (const contractPath of CONTRACT_ALLOWLIST) {
...
 			passthrough: { req: permissions.rootRead },
 			options: {
 				$select: 'slug',
+				$expand: {
+					is_default_for__application: {
+						$top: 2,
+						$select: ['id'],
+						$expand: {
+							owns__release: {
+								$top: 1,
```
thgreasi, line 100: I think that we should use v since we have agreed on the semantics of this field in a past cycle.
Let me know if you find any inconsistency that we might need to fix.
```suggestion
							should_be_running__release: {
```

I'm fine if you want to keep this for a later PR though.

src/features/device-types/device-types-list.ts
thgreasi: I would say keep it and add a TODO to drop it in the next major.
This way we can avoid having any user noticeable changes from the user's point of view.

src/features/device-types/device-types-list.ts
```diff
@@ -84,15 +148,61 @@ export const getDeviceTypeJsons = multiCacheMemoizee(
...
+
+		await Promise.all([
+			...dtsWithAssets.map(async (dt) => {
+				const release = dt.is_default_for__application
+					.map((app) => app.owns__release[0])
+					.find((rel) => rel?.release_asset[0]?.asset != null);
+				if (release == null) {
+					return;
+				}
+				const deviceTypeJson = await getDeviceTypeJsonFromAsset(
```
thgreasi: Failing to fetch the dtjson of one release should not fail the whole endpoint, like in the old implementation

src/features/device-types/device-types-list.ts
```diff
@@ -84,15 +148,61 @@ export const getDeviceTypeJsons = multiCacheMemoizee(
...
+					release.release_asset[0].asset!.href,
+				);
+				deviceTypeJson.buildId = release.raw_version;
+				setResult(dt.slug, deviceTypeJson);
```
thgreasi: How about doing just v here
```suggestion
				result[slug] = deviceTypeJson;
```
and handle the aliases after the `Promise.all`
```
		for (const [slug, dtJson] of Object.entries(result)) {
			if (dtJson.aliases != null) {
				for (const alias of dtJson.aliases) {
					result[alias] = result[slug];
				}
			}
		}
```

src/features/device-types/device-types-list.ts
otaviojacobi: I am not fully sure I trust `should_be_running__release` based on hooks consistency only

src/features/device-types/device-types-list.ts
```diff
@@ -84,15 +150,59 @@ export const getDeviceTypeJsons = multiCacheMemoizee(
...
-		await Promise.all(
-			deviceTypes.map(async ({ slug }) => {
+		await Promise.all([
+			...dtsWithAssets.map(async (dt) => {
+				try {
+					const release = dt.is_default_for__application
+						.map((app) => app.owns__release[0])
+						.find((rel) => rel?.release_asset[0]?.asset != null);
+					if (release == null) {
+						return;
```
thgreasi: Since we have checked it earlier in the `cost appWithAsset = ...` part, shouldn't this `throw ThisShouldNeverHappenError` instead of returning?

src/features/device-types/device-types-list.ts
```diff
@@ -84,15 +150,59 @@ export const getDeviceTypeJsons = multiCacheMemoizee(
...
+					const message = `Found ${hostApps.length} host applications for device type ${slug}, expected at most 1`;
+					captureException(new Error(message), message, {
+						tags: { slug },
+						extra: { hostAppIds: hostApps.map((app) => app.id) },
+					});
+				}
+
+				const appWithAsset = hostApps.find(
+					(app) => app.owns__release[0]?.release_asset[0]?.asset != null,
+				);
```
thgreasi: How about picking just one of the hostApps here,  :imp: , so that we don't have to repeat this `app.owns__release[0]?.release_asset[0]?.asset != null` check in the `Promise.all`?

```suggestion
				const appWithAsset = hostApps.find(
					(app) => app.owns__release[0]?.release_asset[0]?.asset != null,
				);
				if (appWithAsset != null) {
					deviceType.is_default_for__application = [appWithAsset]
				}
```
then in the `Promise.all()` we could do
```
					const release = dt.is_default_for__application[0]?.owns__release[0];
```

This way we separate the decision part of which hostApp is the correct early, in a dedicated spot, while the `Promise.all` focuses just on fetching the info.

otaviojacobi: I really don't like monkey patching and then this becomes even more tricky because it is monkey patching a nested object inside a groupBy which should (functionally speaking) not be modifying things on the fly but rather just categorizing the groups

otaviojacobi: I took a different apporach removing the groupBy and using a few flatMaps to keep the strict typings, lmk what do you think @thgreasi no override needed here

src/features/device-types/device-types-list.ts
```diff
@@ -84,15 +150,54 @@ export const getDeviceTypeJsons = multiCacheMemoizee(
...
+				.find((rel) => rel?.release_asset[0]?.asset != null);
+			return { dt, assetRelease };
+		});
+
+		const dtsWithAssets = dtsWithAssetRelease.flatMap(({ dt, assetRelease }) =>
+			assetRelease != null ? [{ slug: dt.slug, assetRelease }] : [],
+		);
+		let dtsWithoutAssets = dtsWithAssetRelease.flatMap(
+			({ dt, assetRelease }) => (assetRelease == null ? [dt] : []),
+		);
```
thgreasi, line 173: I guess it could be slightly more beautiful, but given that we intend to drop half of later, lgtm for now :+1:

src/features/device-types/device-types-list.ts
```diff
@@ -84,15 +150,54 @@ export const getDeviceTypeJsons = multiCacheMemoizee(
...
+					const deviceTypeJson = await getDeviceTypeJsonFromAsset(
+						assetRelease.release_asset[0].asset!.href,
+					);
+					// TODO: we can possibly drop this field in the next major
+					deviceTypeJson.buildId = assetRelease.raw_version;
+					result[slug] = deviceTypeJson;
+				} catch (err) {
+					captureException(
+						err,
+						`Failed to find a valid build for device type ${slug}`,
```
thgreasi: Maybe it would be useful to differentiate this error a bit from the S3 one, eg saing `hostApp release with assets` instead of `build`.

test/fixtures/02-device-types/releases.json
```diff
@@ -104,5 +104,13 @@
 		"status": "success",
 		"semver": "2024.7.1",
 		"user": "admin"
+	},
+	"releaseDtWithReleaseAsset_1_0_0": {
+		"application": "dt-with-release-asset",
+		"composition": {},
+		"source": "cloud",
+		"status": "success",
+		"semver": "1.0.0",
```
thgreasi: No Resin OS v1 plz :cry: use v7 or something

thgreasi (approved):
:muscle: :sparkles:

test/fixtures/02-device-types/releases.json
otaviojacobi: :robot:

src/features/device-types/device-types-list.ts
```diff
@@ -58,6 +59,18 @@ const getFirstValidBuild = async (
 	}
 };
 
+const getDeviceTypeJsonFromAsset = async (
+	href: string,
+): Promise<DeviceTypeJson> => {
+	const response = await fetch(href);
```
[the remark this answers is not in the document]
otaviojacobi: it cannot control the url, it is auto generated from our db

PR #2310, opened 2026-05-15
Zod patch state

Page-:


src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -115,206 +117,267 @@ const resolveReleaseId = async (
...
+					should_be_running__release: z.number(),
+					name: z.string(),
+					/**
+					 * @deprecated in favor of name, to match the GET endpoint
+					 */
+					device_name: z.string(),
+					status: z.string(),
+					is_online: z.boolean(),
+					note: z.string(),
+					os_version: z.string(),
```
pipex: both `os_version` and `os_variant` are typed as nullable on the supervisor https://github.com/balena-os/balena-supervisor/commit/e161888516e7f234154245cfc8c9f82b6e2b656b#diff-21f20d872d1ae8b1f9e7412a530bdea361cf17df8bf601c6606b829a430c4e28R13

src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -115,206 +117,267 @@ const resolveReleaseId = async (
...
+					os_variant: z.string().nullable(),
+					supervisor_version: z.string(),
+					provisioning_progress: z.number().nullable(),
+					provisioning_state: z.string().nullable(),
+					ip_address: z.string(),
+					mac_address: z.string(),
+					download_progress: z.number().nullable(),
+					api_port: z.number(),
+					api_secret: z.string(),
```
pipex: same with `api_secret` https://github.com/balena-os/balena-supervisor/commit/e161888516e7f234154245cfc8c9f82b6e2b656b#diff-21f20d872d1ae8b1f9e7412a530bdea361cf17df8bf601c6606b829a430c4e28R13

src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -115,206 +117,267 @@ const resolveReleaseId = async (
...
+					memory_usage: z.number(),
+					memory_total: z.number(),
+					storage_block_device: z.string(),
+					storage_usage: z.number(),
+					storage_total: z.number(),
+					cpu_temp: z.number(),
+					cpu_usage: z.number(),
+					cpu_id: z.string(),
+					is_undervolted: z.boolean(),
+					is_on__commit: z.string().nullable(),
```
pipex, line 153: I've never seen this, I assume it was reported at some point but not sure on what supervisor version

src/features/device-state/routes/state-patch-v2.ts
```diff
@@ -115,206 +117,267 @@ const resolveReleaseId = async (
...
-
-		const values = req.body;
-		// firstly we need to extract all fields which should be sent to the device
-		// resource which is everything but the service entries
+export const statePatchV2 = createValidatedRequestHandler(
+	{
+		body: z.object({
+			local: z
+				.object({
+					should_be_running__release: z.number(),
```
pipex, line 125: Never seen this, are there any supported supervisor/OS versions that can still report this?

src/features/device-state/routes/state-patch-v3.ts
```diff
@@ -218,265 +220,334 @@ const fetchData = async (
...
-		const custom: AnyObject = {}; // shove custom values here to make them available to the hooks
+export const statePatchV3 = createValidatedRequestHandler(
+	{
+		body: z.record(
+			// uuid
+			z.string(),
+			z
+				.object({
+					status: z.string(),
+					os_version: z.string(),
```
pipex: `os_version` and `os_variant` as defined as nullable on the supervisor code, in practice I'm not sure if we ever send `null`

pipex (commented):
Left a couple of comments. The types look good overall. We still probably want to test with a few old devices to avoid surprises

src/features/device-state/routes/state-patch-v2.ts
Page-: Interesting, we had it as `string | undefined` here, I've changed to also accept `null`

src/features/device-state/routes/state-patch-v2.ts
Page-: I have no idea honestly, but worst case scenario it's just not sent and nothing happens

src/features/device-state/routes/state-patch-v2.ts
Page-: Yeah this is v2 state so probably relatively old supervisors but we still see plenty of those - nowadays it'd be "is_running__release"

otaviojacobi (approved):

PR #2321, opened 2026-05-19
Simplify `/download-config` endpoint handling of query vs body params

Page-:
This simplifies using the mix of query/body inputs (POST can have both) and makes it clear that in all cases the body var will take priority

https://balena.fibery.io/favorites/Work/Project/1002

Change-type: patch


thgreasi (approved):
:heart:

PR #2322, opened 2026-05-19
Narrow typings of `getUser`/`prefetchAPIKey`/`retrieveAPIKey` to only what is needed

Page-:

Change-type: patch

### DIFF
--- a/src/infra/auth/api-keys.ts
+++ b/src/infra/auth/api-keys.ts
@@ -1,19 +1,18 @@
 import type { Request } from 'express';
 
-import type { hooks, sbvrUtils } from '@balena/pinejs';
+import type { hooks, sbvrUtils, types } from '@balena/pinejs';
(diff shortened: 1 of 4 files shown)


src/infra/auth/api-keys.ts
```diff
@@ -1,19 +1,18 @@
 import type { Request } from 'express';
 
-import type { hooks, sbvrUtils } from '@balena/pinejs';
+import type { hooks, sbvrUtils, types } from '@balena/pinejs';
 import { permissions } from '@balena/pinejs';
 
 import { isJWT } from './jwt-passport.js';
 
-const isRequest = (req: hooks.HookReq | Request): req is Request =>
```
Page-, line 8: I inlined this because it meant that typescript could automatically handle the type inference without me needing to jump through a load of hoops to manually type it to support generics

otaviojacobi (approved):

PR #2327, opened 2026-05-20
Add `zod` based validation for the `/supervisor/*` proxy endpoint

Page-:

Change-type: patch


src/features/device-proxy/device-proxy.ts
```diff
@@ -86,80 +90,101 @@ const validateSupervisorResponse = (
...
+						'Device ID must be a valid integer if specified',
+					);
+				}
 			}
-		}
-		if (uuid != null) {
-			if (typeof uuid !== 'string') {
-				throw new BadRequestError('UUID must be a valid string if specified');
+			if (uuid != null) {
+				if (typeof uuid !== 'string') {
```
joshbwlng: I don't think we need this typeof check any longer

joshbwlng (approved):

PR #2329, opened 2026-05-20
47.x

Page-:


src/features/device-config/device-config.ts
```diff
@@ -111,15 +128,15 @@ export const generateConfig = async (
...
-				60 *
-				1000,
-			network: getBodyOrQueryParam(req, 'network'),
-			wifiSsid: getBodyOrQueryParam(req, 'wifiSsid'),
-			wifiKey: getBodyOrQueryParam(req, 'wifiKey'),
-			ip: getBodyOrQueryParam(req, 'ip'),
-			gateway: getBodyOrQueryParam(req, 'gateway'),
-			netmask: getBodyOrQueryParam(req, 'netmask'),
+				typeof params.appUpdatePollInterval === 'string'
+					? parseInt(params.appUpdatePollInterval, 10) * 60 * 1000
```
joshbwlng: While we're at it, how about replacing `60 * 1000` with `MINUTES` from `@balena/env-parsing`?

Page-: In general I'm trying to keep the changes minimal so as to not accidentally break things, although I realized in this case I was no longer doing the minutes -> milliseconds conversion if a number was passed in directly so I fixed that

otaviojacobi (approved):

PR #2346, opened 2026-06-02
Add new scoped roles jwt type

otaviojacobi:
See: https://balena.fibery.io/Work/Project/Ingest-audit-logs-from-balena-proxy-2469
Change-type: minor

### DIFF
--- a/src/infra/auth/jwt-passport.ts
+++ b/src/infra/auth/jwt-passport.ts
@@ -10,12 +10,13 @@ import { JSON_WEB_TOKEN_SECRET } from '../../lib/config.js';
 import { captureException } from '../error-handling/index.js';
 import type {
 	ScopedAccessToken,
+	ScopedRolesToken,
 	ScopedToken,
 	TokenUserPayload,
 } from './jwt.js';
 import type { User } from '../../balena-model.js';
 import type { PickDeferred } from '@balena/abstract-sql-to-typescript';
-import { getGuestActorId } from './permissions.js';
+import { getGuestActorId, getRolePermissions } from './permissions.js';
 import { createUnvalidatedRequestHandler } from '../validation/index.js';
 
 export type { SignOptions } from 'jsonwebtoken';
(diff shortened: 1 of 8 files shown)


src/infra/auth/permissions.ts
```diff
@@ -76,6 +78,48 @@ export const revokeUserRole = async (user: number, role: number, tx: Tx) => {
...
+			},
+		});
+		return perms.map(({ name }) => name);
+	},
+	{
+		cacheKey: 'getRolePermissions',
+		promise: true,
+		primitive: true,
+		maxAge: ROLE_PERMISSIONS_CACHE_TIMEOUT,
+		normalizer: ([roleNames]) => JSON.stringify([...roleNames].sort()),
```
Page-:
```suggestion
		normalizer: ([roleNames]) => JSON.stringify(roleNames.toSorted()),
```

src/infra/auth/jwt.ts
```diff
@@ -196,6 +196,47 @@ export function createScopedAccessToken(
...
+	}
+	const base = {
+		actor: options.actor,
+		roles: options.roles,
+		bindings: options.bindings,
+	};
+	const payload: ScopedRolesToken =
+		userId != null && jwtSecret != null
+			? { ...base, userId, jwt_secret: jwtSecret }
+			: base;
```
Page-: Any reason to not mutate `base` instead? Rather than creating a duplicate object in the userId case

otaviojacobi: Just in general I prefer avoid mutations, but I am fine with adding the properties too :+1: 

I inlined it to make the typings work on our favor too

src/lib/config.ts
```diff
@@ -207,6 +207,10 @@ export const RATE_LIMIT_FACTOR = intVar('RATE_LIMIT_FACTOR', 1);
 export const RATE_LIMIT_MEMORY_BACKEND = optionalVar(
 	'RATE_LIMIT_MEMORY_BACKEND',
 );
+export const ROLE_PERMISSIONS_CACHE_TIMEOUT = intVar(
+	'ROLE_PERMISSIONS_CACHE_TIMEOUT',
+	15 * MINUTES,
```
Page-: I think we more commonly go with 5 minutes as the default

src/infra/auth/jwt-passport.ts
```diff
@@ -85,6 +90,17 @@ const processVerifiedJwtPayload = async (
 		const { service, apikey } = jwtUser;
 		const apiKeyPermissions = await permissions.getApiKeyPermissions(apikey);
 		return { service, apikey, permissions: apiKeyPermissions };
+	} else if ('roles' in jwtUser && jwtUser.roles != null) {
+		const { actor } = jwtUser;
+		if ('jwt_secret' in jwtUser && jwtUser.jwt_secret != null) {
+			await fetchUser({ actor: jwtUser.actor }, jwtUser.jwt_secret);
```
Page-: It might be worth renaming this from `fetchUser` to `verifyJwtAndFetchUser` or something because calling a `fetchX` function and not doing anything with the result looks like a mistake

Page- (approved):

PR #2354, opened 2026-06-15
Resolve JWT users in jwt-passport instead of via injection

otaviojacobi:
Change-type: major

### DIFF
--- a/src/infra/auth/jwt-passport.ts
+++ b/src/infra/auth/jwt-passport.ts
@@ -3,8 +3,7 @@ import passport from 'passport';
 import { ExtractJwt, Strategy as JwtStrategy } from 'passport-jwt';
 import { TypedError } from 'typed-error';
 
-import type { sbvrUtils } from '@balena/pinejs';
-import { permissions } from '@balena/pinejs';
+import { permissions, sbvrUtils } from '@balena/pinejs';
 
 import { JSON_WEB_TOKEN_SECRET } from '../../lib/config.js';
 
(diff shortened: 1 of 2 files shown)


src/infra/auth/jwt-passport.ts
```diff
@@ -21,7 +20,9 @@ import { createUnvalidatedRequestHandler } from '../validation/index.js';
 
 export type { SignOptions } from 'jsonwebtoken';
 
-class InvalidJwtSecretError extends TypedError {}
+const { api } = sbvrUtils;
+
+export class InvalidJwtSecretError extends TypedError {}
```
Page-: I don't think this needs to be exported?

src/infra/auth/jwt-passport.ts
```diff
@@ -47,13 +48,50 @@ const jwtFromRequest = ExtractJwt.versionOneCompatibility({
 	authScheme: 'Bearer',
 });
 
-type FetchUserFn = (
-	id: number,
-) => Promise<PickDeferred<User['Read'], 'jwt_secret' | 'actor'> | undefined>;
+function fetchUser(
+	searchId: { userId: number },
```
Page-: I think it would be better to do
```suggestion
	searchId: { id: number },
```
so that it would be in a format that would be valid to pass directly to the `id` property of a pinejs client query if we wanted, or perhaps
```suggestion
	searchId: Pick<User['Write'], 'id'>,
```

otaviojacobi: I am not sure I follow? You suggest we only change the searchId naming for specific the user case? we would have `{id: number}` or `{actorId: number}` ? I went with the current one so that use vs actor is explicit in the name property (so you don't have the case to be searching for a user byId when you have the actor, for example)

otaviojacobi: Got it, thanks! This is indeed much better

Page- (approved):

PR #2362, opened 2026-06-16
Use `actor` id instead of $filter

otaviojacobi:
Change-type: patch


thgreasi (approved):

src/features/auth/whoami.ts
```diff
@@ -70,16 +70,11 @@ export const actorWhoami = createValidatedRequestHandler(async (req, res) => {
...
 					resource: 'user',
 					passthrough: { req, tx },
-					options: {
-						$top: 1,
-						$select: 'id',
-						$filter: {
-							actor: req.user?.actor,
-						},
-					},
+					id: { actor: req.user?.actor },
```
thgreasi: Does it work like v ?
```suggestion
					id: { actor: req.user.actor },
```

PR #2367, opened 2026-06-23
Contracts: avoid setting assets to `undefined` when none are passed

Page-:
This saves a little work but also avoids issues if the contract data is ever expected to be valid JSON as JSON cannot contain undefined

Change-type: patch

### DIFF
--- a/src/features/contracts/contracts-directory.ts
+++ b/src/features/contracts/contracts-directory.ts
@@ -30,12 +30,8 @@ const handleLocalAssetUrl = async (assetUrl: string): Promise<string> => {
 
 const normalizeAssets = async (
 	contractFilepath: string,
-	assets: Contract['assets'],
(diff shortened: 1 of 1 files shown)


src/features/contracts/contracts-directory.ts
```diff
@@ -89,7 +89,12 @@ const normalizeContract = async (
 			contract.aliases.push(contract.slug);
 		}
 
-		contract.assets = await normalizeAssets(contractFilepath, contract.assets);
+		if (contract.assets != null) {
+			contract.assets = await normalizeAssets(
```
thgreasi, line 89: Can we then also change `normalizeAssets`'s param to `assets: NonNullable<Contract['assets']>` ?

thgreasi (approved):

PR #2370, opened 2026-06-26
Delete old cache images on new successful release

joshbwlng:
Change-type: minor

---

See: https://balena.fibery.io/Work/Project/Mark-multistage-cache-images-for-deletion-when-new-releases-are-created-2246

### DIFF
--- a/src/features/images/tasks/delete-registry-images.ts
+++ b/src/features/images/tasks/delete-registry-images.ts
@@ -17,7 +17,7 @@ const queue = new PQueue({
 	concurrency: ASYNC_TASK_DELETE_REGISTRY_IMAGES_CONCURRENCY,
 });
 
-const schema = {
+export const schema = {
 	type: 'object',
 	properties: {
 		images: {
@@ -33,10 +33,6 @@ const schema = {
 				additionalProperties: false,
 			},
 		},
-		onlyDeleteCache: {
-			type: 'boolean',
-			default: false,
-		},
 	},
 	required: ['images'],
 	additionalProperties: false,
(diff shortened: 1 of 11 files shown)


src/features/images/tasks/delete-registry-images.ts
```diff
@@ -33,6 +33,10 @@ const schema = {
 				additionalProperties: false,
 			},
 		},
+		onlyDeleteCache: {
```
thgreasi: Should we add this in a separate PR before starting adding records w/ the new property in the DB?
How would old instances behave if they see task w/ this new property?
Will they explode?
Will they ignore the new property and delete the whole image?

src/features/images/tasks/delete-registry-images.ts
thgreasi, line 53: I wouldn't be against a dedicated new task handler (which can share a bunch of code w/ this one), b/c the `onlyDeleteCache: true` seems weird. If we did have flags, I would prefer them to be explicit on what to do, instead of what to not do.

joshbwlng: That could work, maybe a dedicated task like `delete_registry_image_cache` would be good. I'd just want to avoid duplicating code between the two tasks.

src/features/images/tasks/delete-registry-images.ts
```diff
@@ -89,6 +95,7 @@ async function deleteRepo(
 
 const deleteRegistryImages = async ({
 	images,
+	onlyDeleteCache = false,
```
thgreasi: That's a scary param to have a default value for tbh.
Maybe we could make it super explicit w/ something like v or `deleteTarget: 'image' | 'cache-only'`?
```suggestion
	{
		deleteImage: boolean,
		deleteCache: boolean,
	}
```

joshbwlng: I agree being more explicit makes sense

joshbwlng:
Tested on BoK and cache images were cleaned up as expected, the latest release kept it's cache.

PR #2375, opened 2026-07-07
Registry: Delete repository _manifests directory

joshbwlng:
Instead of using Docker Distribution's delete endpoint, we need to delete the entire _manifests directory via the s3 client. This is because the delete endpoint could leave some things behind that can cause deleted images to still be listed in the registry's catalog.

Change-type: patch

---

See: https://balena.fibery.io/Work/Project/Mark-existing-registry-trash-data-for-deletion-via-a-migration-2245

ToDo: Should probably avoid removing exported functions for now.

### DIFF
--- a/src/features/registry/registry.ts
+++ b/src/features/registry/registry.ts
@@ -804,14 +804,6 @@ class S3Client {
 		return digests;
 	}
 
-	// List all existing manifest digests (tagged and untagged) for a given
-	// repository by listing objects in the registry's S3 bucket directly.
-	async listRepoDigests(repo: string) {
-		return await this.listDigestsAt(
-			`${this.rootPath}/repositories/${repo}/_manifests/revisions/sha256/`,
-		);
-	}
-
 	// List all manifest digests ever associated with a given tag of a
 	// repository (current and historical) by listing objects in the
 	// registry's S3 bucket directly.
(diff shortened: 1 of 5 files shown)


joshbwlng:
Tested on BoK and it worked as expected. Without the fix, random data sticking around in the `_manifests` subdirectory in a repositories directory in the registry kept the repository in the list returned from the `v2/_catalog` endpoint - this unexpected data can be caused in production by things like direct pushes from ancient docker clients. With the fix, the entire `_manifests` directory is removed and the repository correctly no longer shows up in the `v2/_catalog` endpoint results.

src/features/registry/registry.ts
```diff
@@ -828,6 +820,49 @@ class S3Client {
...
+				.listObjectsV2({
+					Bucket: this.bucket,
+					Prefix: prefix,
+					ContinuationToken: continuationToken,
+				})
+				.promise();
+
+			const objects = (res.Contents ?? [])
+				.map(({ Key }) => Key)
+				.filter((key): key is string => key != null)
```
Page-: This type guard is wrong, it only checks that it's not nullish (which I think typescript can maybe infer automatically?) and not that it's a string. It's possible with how things are today that it's equivalent but I much prefer typing these sorts of guards correctly so they continue to be correct if things change elsewhere

src/features/registry/registry.ts
```diff
@@ -828,6 +820,49 @@ class S3Client {
...
+			const res = await this.s3
+				.listObjectsV2({
+					Bucket: this.bucket,
+					Prefix: prefix,
+					ContinuationToken: continuationToken,
+				})
+				.promise();
+
+			const objects = (res.Contents ?? [])
+				.map(({ Key }) => Key)
```
Page-: Is this double map to filter out other properties so we end with just `{ Key }` or is there some other reason for it? It looks a bit strange at first look is all

joshbwlng: @Page- The goal was to get this shape that the `deleteObjects` is happy with:
```typescript
const objects: {
    Key: string;
}[]
```

I've pushed a new commit to change this to a simpler `flatMap()`. I also wouldn't hate a simple `for` loop implementation either.

Page- (approved):

PR #2388, opened 2026-07-17
Add build-time image profile resource

otaviojacobi:
See: https://balena.fibery.io/Work/Project/Supporting-docker-compose-profiles-on-balenaCloud-2480
Change-type: minor

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -203,6 +203,9 @@ Term: memory usage
 Term: memory total
 	Concept Type: Integer (Type)
 
+Term: profile name
+	Concept Type: Short Text (Type)
+
 Term: storage block device
 	Concept Type: Short Text (Type)
 
@@ -430,6 +433,11 @@ Fact type: image is part of release
 	Fact type: release image has label name
 		Term Form: image label
 		Database Table Name: image label
+
+	Fact type: release image has profile name
+		Term Form: image profile
+		Database Table Name: image profile
+		Necessity: each image profile has a profile name that has a Length (Type) that is greater than 1 and is less than or equal to 100.
 
 	Fact type: release image has name (Auth)
 		Term Form: image environment variable
--- a/src/infra/validation/index.ts
+++ b/src/infra/validation/index.ts
@@ -1,5 +1,7 @@
 import type { Request, RequestHandler } from 'express';
 import type { ParamsDictionary } from 'express-serve-static-core';
+import { errors } from '@balena/pinejs';
+import type { hooks } from '@balena/pinejs';
 import type { z } from 'zod';
 export { z } from 'zod';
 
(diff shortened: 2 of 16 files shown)


src/lib/auth.ts
```diff
@@ -144,6 +145,7 @@ export const DEVICE_API_KEY_PERMISSIONS = [
 	)})`,
 
 	'resin.image_label.read?release_image/canAccess()',
+	'resin.image_profile.read?release_image/canAccess()',
```
otaviojacobi, line 148: So that device api keys have permissions to read/expand into it on target state later on.

src/balena.sbvr
```diff
@@ -431,6 +434,11 @@ Fact type: image is part of release
 		Term Form: image label
 		Database Table Name: image label
 
+	Fact type: release image has profile name
+		Term Form: image profile
+		Database Table Name: image profile
+		Necessity: each image profile has a profile name that has a Length (Type) that is greater than 1 and is less than or equal to 100.
```
Page-, line 440: No profile names like `'x'`?

src/migrations/00113-add-image-profile.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE IF NOT EXISTS "image profile" (
```
Page-, line 1: This migration shouldn't be necessary since it's a completely new table and would be automatically created?

src/infra/validation/index.ts
```diff
@@ -106,3 +109,26 @@ export const createUnvalidatedRequestHandler = (
...
+> & {
+	request: Omit<hooks.HookArgs['request'], 'values'> & {
+		values: z.infer<Schema>;
+	};
+};
+
+export const withValidatedValues =
+	<Schema extends z.ZodType<AnyObject>>(
+		schema: Schema,
+		next?: (args: ValidatedHookArgs<Schema>) => hooks.HookResponse,
```
Page-: I'm not sure on the goal of this arg, and would leave it out until it's with associated code that would actually need it so that it can be reviewed together as/if needed

src/features/image-profiles/validation.ts
```diff
@@ -0,0 +1,20 @@
...
+
+// Matches docker-compose profile name semantics, see
+// https://docs.docker.com/compose/how-tos/profiles/
+const PROFILE_NAME_REGEX = /^[a-zA-Z0-9][a-zA-Z0-9_.-]+$/;
+
+const profileNameSchema = z.looseObject({
+	profile_name: z.string().regex(PROFILE_NAME_REGEX),
+});
+
+export const registerImageProfileHooks = (resource: string) => {
```
Page-: I'd recommend a stricter typing for `resource`,at a minimum `keyof BalenaModel` but probably just literals of the actual resources it can work for

src/balena.sbvr
otaviojacobi: No - @Page- it is not allowed it must be at least two chars as per "Valid profiles names follow the regex format of [a-zA-Z0-9][a-zA-Z0-9_.-]+." - so at least 2 chars

src/infra/validation/index.ts
otaviojacobi: Sure :+1: I wanted a generic thing for validation but agree it can be done when needed

src/features/image-profiles/validation.ts
otaviojacobi: This was just unnecessary abstraction from a previous version I had written, thanks

As thhis evolves, it will be reused and can be made more generic, but as it stands, a simpler version does the job so I will keep it simple

src/migrations/00113-add-image-profile.sql
otaviojacobi: @Page- I don't think it is strictly necessary for things to work however as far as I remember this was expected as a way for later readers to know the precise effect on the DB of adding this model, and done for all previously recent created tables, e.g. `0076-add-scheduled-job-run.sql` `0070-device-type-alias.sql` etc. 

I can remove, I just want to point out that this is apprently the current pattern

Page- (approved):

PR #2396, opened 2026-07-22
Switch the v6 conversion of string contracts to JSON into the validator

Page-:
This is also necessary as otherwise the pinejs validation will reject the invalid type before ever getting to the hook that does the conversion

Change-type: patch

### DIFF
--- a/src/translations/v6/v6.ts
+++ b/src/translations/v6/v6.ts
@@ -4,12 +4,15 @@ import type { ConfigLoader } from '@balena/pinejs';
 import {
 	aliasFields,
 	aliasTable,
+	extendValidator,
 	generateAbstractSqlModel,
 	overrideFieldType,
 	renameResourceField,
 } from '../../abstract-sql-utils.js';
 import * as userHasDirectAccessToApplication from '../../features/applications/models/user__has_direct_access_to__application.js';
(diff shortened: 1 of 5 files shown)


src/translations/v6/v6.ts
```diff
@@ -239,3 +242,21 @@ export const getV6Translations = (abstractSqlModel = v6AbstractSqlModel) => {
...
+export const extendValidators = () => {
+	extendValidator('v6', 'release', {
+		// In v6 we also accept contract as a string that we automatically parse, this used to be via a hook but we can now do it directly in the validator
+		contract: z
+			.preprocess((value) => {
+				if (typeof value === 'string') {
+					try {
+						return JSON.parse(value);
+					} catch {
+						// ignore
```
otaviojacobi, line 255: I know that is fine for now (as we only validate and don't throw any zod managed error yet) but in the future, I think we would want to look into actually capturing and the using cdx.addIssue and a return z.NEVER

Page-: Yeah I think that would be an option, although equally even with the current form you would get an error message pointing you to that it should be an object (which in general is what we'd want people to actually send) rather than a string - if we were exposing the error messages that is

otaviojacobi (approved):

PR #2397, opened 2026-07-22
Unfilter fleet hostapps services based on profiles

otaviojacobi:
See: https://balena.fibery.io/Work/Project/Supporting-docker-compose-profiles-on-balenaCloud-2480
Cahge-type: major

### DIFF
--- a/src/features/device-state/routes/state-get-v3.ts
+++ b/src/features/device-state/routes/state-get-v3.ts
@@ -93,6 +93,7 @@ export type StateV3 = {
 		};
 	};
 };
+export const noActiveProfiles: ReadonlySet<string> = new Set();
 
 export function buildAppFromRelease(
 	device: undefined,
@@ -100,6 +101,7 @@ export function buildAppFromRelease(
 	release: ExpandedRelease,
 	config: Record<string, string>,
 	defaultLabels?: Record<string, string>,
+	activeProfiles?: ReadonlySet<string>,
 ): NonNullable<LocalStateApp['releases']>;
 export function buildAppFromRelease(
 	device: ExpandedDevice,
@@ -107,6 +109,7 @@ export function buildAppFromRelease(
 	release: ExpandedRelease,
 	config: Record<string, string>,
 	defaultLabels?: Record<string, string>,
+	activeProfiles?: ReadonlySet<string>,
 ): NonNullable<LocalStateApp['releases']>;
 export function buildAppFromRelease(
 	device: ExpandedDevice | undefined,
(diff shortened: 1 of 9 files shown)


src/features/device-state/routes/state-get-v3.ts
```diff
@@ -155,6 +158,16 @@ export function buildAppFromRelease(
 				}`;
 
 	for (const ipr of release.release_image) {
+		if (
+			activeProfiles != null &&
```
otaviojacobi: undefined/null profiles means "this is not a hostapp" so we don't change the current behavior. Once we have fleet (userapp) profiles and supervisor profiles this becomes non-nullable

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -155,6 +158,16 @@ export function buildAppFromRelease(
...
+			ipr.image_profile.length > 0 &&
+			!ipr.image_profile.some(({ profile_name }) =>
+				activeProfiles.has(profile_name),
+			)
```
thgreasi: Is there any reason to not do it like this v upfront (which imo is a bit more readable), and rely just on the sbvr rule to block users from activating profiles to anything else than the hostApp?
```suggestion
			// When a release image has profiles set, then it ONLY runs if one of them is active
			ipr.image_profile.length > 0 &&
			(
				activeProfiles == null ||
				!ipr.image_profile.some(({ profile_name }) =>
					activeProfiles.has(profile_name),
				)
			)
```
Semantically, if somehow users do set profiles on their own images, then they should already not be returned by the target state.
If we don't do this upfront today, then once we later enable profiles for user app, it will then be a breaking change b/c suddenly any user app images with image_profiles set (which they can do atm afaict) will stop running.

otaviojacobi: I am fine doing the breaking change now but if any customer has already used a profile this would happen anyway now

otaviojacobi: If we go that route than I would prefer to make the activeProfiles non-nullable and always empty from the other incoming paths and then we stick the correctt logic from the get go without any `== null`  check necessary @thgreasi

src/features/device-state/routes/state-get-v3.ts
```diff
@@ -93,27 +93,31 @@ export type StateV3 = {
...
 	defaultLabels?: Record<string, string>,
+	activeProfiles?: ReadonlySet<string>,
 ): NonNullable<LocalStateApp['releases']>;
 export function buildAppFromRelease(
 	device: ExpandedDevice | undefined,
 	application: ExpandedApplication | ExpandedApplicationWithService,
 	release: ExpandedRelease,
 	config: Record<string, string>,
 	defaultLabels?: Record<string, string>,
+	activeProfiles: ReadonlySet<string> = noActiveProfiles,
```
otaviojacobi, line 120: Mimic docker behavior of defaulting to no noActiveProfiles

Page- (approved):

thgreasi (commented):
Even better :) 👍

PR #2401, opened 2026-07-29
Add onRefreshToken callback

JSReds:
Change-type: minor


src/features/auth/refresh-token.ts
```diff
@@ -6,38 +6,49 @@ import {
...
-	{
-		body: z.object({
-			password: z.string().nullish(),
-		}),
-	},
-	async (req, res) => {
-		try {
-			await sbvrUtils.db.readTransaction(async (tx) => {
-				const { password } = req.body;
+export const refreshToken = (onRefreshToken: SetupOptions['onRefreshToken']) =>
```
thgreasi, line 13: Since this is also exported and used in bC-api, how about retaining the old signature of `refreshToken`, and have `setup` store `onRefreshToken` to a local variable and then using it here?
Similar to how `setUserTokenDataCallback` works today, but set via `setup()` so that it's closer to `onLogin`.

src/features/auth/refresh-token.ts
```diff
@@ -6,38 +6,49 @@ import {
...
+						await checkUserPassword(password, creds.id, tx);
+						creds.authTime = Date.now();
 					}
-					await checkUserPassword(password, creds.id, tx);
-					creds.authTime = Date.now();
+					if (onRefreshToken) {
+						const creds = req.creds!;
+						if (!('id' in creds)) {
+							throw new BadRequestError('Can only refresh user tokens');
+						}
```
thgreasi, line 40: I think this should be left for the consumer to handle if needed
```suggestion
```

src/features/auth/refresh-token.ts
```diff
@@ -6,38 +6,49 @@ import {
...
+							throw new BadRequestError(
+								'Can only password refresh user tokens',
+							);
+						}
+						await checkUserPassword(password, creds.id, tx);
+						creds.authTime = Date.now();
 					}
-					await checkUserPassword(password, creds.id, tx);
-					creds.authTime = Date.now();
+					if (onRefreshToken) {
```
thgreasi, line 36: NIT to be consistent w/ the coding style elsewhere 
```suggestion
					if (onRefreshToken != null) {
```

src/features/auth/refresh-token.ts
thgreasi: which then can even become `await onRefreshToken?.(req, tx);`

src/index.ts
```diff
@@ -311,6 +311,10 @@ export interface SetupOptions {
 		tx: Tx,
 		req: RequestExcludingInput,
 	) => PromiseLike<void> | void;
+	onRefreshToken?: (
+		req: RequestExcludingInput,
+		tx: Tx,
+	) => PromiseLike<void> | void;
```
thgreasi: wdyt about v, since you can see `onLogin`'s params have gradually evolved to not be that great.
```suggestion
	onRefreshToken?: (params: {
		req: RequestExcludingInput,
		tx: Tx,
	}) => PromiseLike<void> | void;
```

thgreasi (approved):

PR #2423, opened 2026-08-14
Add device_profile_override model

otaviojacobi:
See: https://balena.fibery.io/Work/Project/Enabling-device-overrides-for-hostapp-extensions-2840
Change-type: minor

### DIFF
--- a/src/balena.sbvr
+++ b/src/balena.sbvr
@@ -1004,6 +1004,13 @@ Fact type: application1 activates profile name on application2
 	Necessity: each application profile has a profile name that has a Length (Type) that is greater than 1 and is less than or equal to 100.
 	Necessity: each application profile has an application2 that is host and is of a class that is not equal to "block".
 
+Fact type: device overrides profile name on application
+	Term Form: device profile override
+	Database Table Name: device profile override
+	Necessity: each device profile override has a profile name that has a Length (Type) that is greater than 1 and is less than or equal to 100.
+	Necessity: each device profile override has an application that is host and is of a class that is not equal to "block".
+Fact type: device profile override is active
+
 
 -- Rules
 
@@ -1023,3 +1030,4 @@ Rule: It is necessary that each release that should manage a device that is of a
 Rule: It is necessary that each application that is not host, is updated by no application.
 Rule: It is necessary that each application that is of a class that is not equal to "block", updates no application.
 Rule: It is necessary that each application1 that activates a profile name on an application2, is of a class that is equal to "fleet".
+Rule: It is necessary that each device profile override that has a device and has an application, has a device that should be operated by a release that belongs to the application.
--- a/src/features/cascade-delete/hooks.ts
+++ b/src/features/cascade-delete/hooks.ts
@@ -19,6 +19,7 @@ setupDeleteCascade('application', {
 setupDeleteCascade('device', {
 	device_config_variable: 'device',
 	device_environment_variable: 'device',
+	device_profile_override: 'device',
 	device_service_environment_variable: 'device',
 	device_tag: 'device',
 	image_install: 'device',
--- a/src/migrations/00115-add-device-profile-override.sql
+++ b/src/migrations/00115-add-device-profile-override.sql
@@ -1,0 +1,37 @@
+CREATE TABLE IF NOT EXISTS "device profile override" (
+	"created at" TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
+,	"modified at" TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
+,	"device" INTEGER NOT NULL
+,	"overrides-profile name" VARCHAR(255) NOT NULL
+,	"on-application" INTEGER NOT NULL
+,	"id" SERIAL NOT NULL PRIMARY KEY
+,	"is active" BOOLEAN DEFAULT FALSE NOT NULL
+,	FOREIGN KEY ("device") REFERENCES "device" ("id")
+,	FOREIGN KEY ("on-application") REFERENCES "application" ("id")
+,	UNIQUE("device", "overrides-profile name", "on-application")
+,	-- It is necessary that each device profile override has a profile name that has a Length (Type) that is greater than 1 and is less than or equal to 100.
+CONSTRAINT "device profile override$4IQoYl4sSKpj7XIAC5UG/vrecsyzot8Yo8n6W/P" CHECK (1 < LENGTH("overrides-profile name")
+AND LENGTH("overrides-profile name") <= 100
+AND LENGTH("overrides-profile name") IS NOT NULL
+AND "overrides-profile name" IS NOT NULL)
+);
+
+DO
+$$
+BEGIN
+	IF NOT EXISTS(
+		SELECT 1
+		FROM "information_schema"."triggers"
+		WHERE "event_object_table" = 'device profile override'
+		AND "trigger_name" = 'device profile override_trigger_update_modified_at'
+	) THEN
+		CREATE TRIGGER "device profile override_trigger_update_modified_at"
+		BEFORE UPDATE ON "device profile override"
+		FOR EACH ROW
+		EXECUTE PROCEDURE "trigger_update_modified_at"();
+	END IF;
+END;
+$$;
+
+CREATE INDEX IF NOT EXISTS "device_profile_override_on_application_idx"
+ON "device profile override" ("on-application");
--- a/test/27_profiles.ts
+++ b/test/27_profiles.ts
@@ -1,5 +1,6 @@
 import { expect } from 'chai';
 import * as fixtures from './test-lib/fixtures.js';
+import * as fakeDevice from './test-lib/fake-device.js';
 import type { UserObjectParam } from './test-lib/supertest.js';
 import { supertest } from './test-lib/supertest.js';
 import * as versions from './test-lib/versions.js';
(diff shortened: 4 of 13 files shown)


src/balena.sbvr
```diff
@@ -1004,6 +1004,14 @@ Fact type: application1 activates profile name on application2
 	Necessity: each application profile has a profile name that has a Length (Type) that is greater than 1 and is less than or equal to 100.
 	Necessity: each application profile has an application2 that is host and is of a class that is not equal to "block".
 
+Fact type: device overrides profile name on application
+	Term Form: device profile override
+	Database Table Name: device profile override
+	Necessity: each device profile override has a profile name that has a Length (Type) that is greater than 1 and is less than or equal to 100.
+	Necessity: each device profile override has an application that is of a class that is not equal to "block".
```
otaviojacobi: Consistent with the permission we used for application profile

src/balena.sbvr
```diff
@@ -1023,3 +1031,4 @@ Rule: It is necessary that each release that should manage a device that is of a
 Rule: It is necessary that each application that is not host, is updated by no application.
 Rule: It is necessary that each application that is of a class that is not equal to "block", updates no application.
 Rule: It is necessary that each application1 that activates a profile name on an application2, is of a class that is equal to "fleet".
+Rule: It is necessary that each device profile override that has a device and has an application, has a device that belongs to the application or should be operated by a release that belongs to the application or should be managed by a release that belongs to the application.
```
otaviojacobi: We could alternatevely reduce to only is operated by as this is the only case where it can really be used so far but this rule would be the "target state" for the rule. I see no harm in starting with the end goal here but no strong opinions.

src/balena.sbvr
```diff
@@ -1004,6 +1004,14 @@ Fact type: application1 activates profile name on application2
 	Necessity: each application profile has a profile name that has a Length (Type) that is greater than 1 and is less than or equal to 100.
 	Necessity: each application profile has an application2 that is host and is of a class that is not equal to "block".
 
+Fact type: device overrides profile name on application
+	Term Form: device profile override
+	Database Table Name: device profile override
+	Necessity: each device profile override has a profile name that has a Length (Type) that is greater than 1 and is less than or equal to 100.
+	Necessity: each device profile override has an application that is host and is of a class that is not equal to "block".
```
otaviojacobi, line 1011: Same patterns as application profile

src/features/cascade-delete/hooks.ts
```diff
@@ -12,13 +12,15 @@ setupDeleteCascade('application', {
 	application_environment_variable: 'application',
 	application_tag: 'application',
 	application_profile: ['application', 'on__application'],
+	device_profile_override: 'on__application',
```
thgreasi: Is this necessary?
Is it atm just for handling the case of a hostApp being deleted?

otaviojacobi: It is for the case of hostapp deletion, altough we haven't done those on cloud this should not fail because of the profile cascade deletion missing so yes, it is imo necessary

src/migrations/00115-add-device-profile-override.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS "device profile override" (
+	"created at" TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
+,	"modified at" TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
+,	"device" INTEGER NOT NULL
+,	"overrides-profile name" VARCHAR(255) NOT NULL
+,	"on-application" INTEGER NOT NULL
+,	"id" SERIAL NOT NULL PRIMARY KEY
+,	"is active" BOOLEAN DEFAULT FALSE NOT NULL
+,	FOREIGN KEY ("device") REFERENCES "device" ("id")
+,	FOREIGN KEY ("on-application") REFERENCES "application" ("id")
```
thgreasi, line 10: Let's also add an index for `on-application`, since otherwise the `device_profile_override: 'on__application'` cascade delete, will have to do a full table scan on `device profile override` .

otaviojacobi: Sure, altough I think it is premature optimization

src/balena.sbvr
```diff
@@ -1004,6 +1004,14 @@ Fact type: application1 activates profile name on application2
...
+
```
thgreasi: Let's keep the fact types of the same table together :)
```suggestion
```

src/features/cascade-delete/hooks.ts
Page-: This is so that it's possible to delete a host app whilst devices are pointed at it? I'd rather have that fail tbh rather than have a cascade built in since we shouldn't be deleting them (easily) whilst they're in use

src/migrations/00115-add-device-profile-override.sql
Page-: In general we just add foreign key indexes by default because the negative of missing one when you need it massively outweighs the case of having it when you don't need it (it's a very rare case where omitting a foreign key index is of noticeable benefit)

src/balena.sbvr
thgreasi: +1 on checking only `should_be_operated_by__release` since as you said atm it's a bit contradicting w/ the other rule right above.
Happy to keep a `TODO` w/ the target state rule.
Even happier to see a test about it, but I think I know what you would answer :P

otaviojacobi: I am fine with only the `should_be_operated_by__release` for now too, as I said, I have no strong opinions as one could argue this is the "smallest" change to do aswell

src/features/cascade-delete/hooks.ts
thgreasi: +1 on not having it (that;s why I was asking)
We are atm also on purpose missing such cascades for other hostApp deletion related actions, so it's good to be consistent.

otaviojacobi: @Page-  FWIW I don't think what you mention will happen because I think it fails to delete the actual devices (so that the cascade delete as a whole and the whole transaction collapses precisely because it fails to delete the hostapp if devices should_be__operated_by it yet, I just think the overrides should not be the reason for this tx fail and rather the device itself)

Page-: If this only applies to host apps I would still avoid it for safety/to avoid prematurely adding things, if/when it gets expanded to user apps where the cascade makes sense I would be happy for the cascade to be added. That said I wouldn't block on it as I agree with you that I'd expect it just shunts the failure down the chain a bit

otaviojacobi: Sure, I removed the cascade deletion

test/27_profiles.ts
```diff
@@ -372,6 +380,227 @@ export default () => {
...
+						.expect(201);
+
+					expect(res.body)
+						.to.have.nested.property('device.__id')
+						.that.equals(this.hostAppOperatedDevice.id);
+					expect(res.body)
+						.to.have.nested.property('on__application.__id')
+						.that.equals(this.hostApp.id);
+					expect(res.body.overrides__profile_name).to.equal('kernel-modules');
+					expect(res.body.is_active).to.equal(true);
```
thgreasi: not a blocker, but slightly shorter:

```suggestion
					expect(res.body).to.have.nested.property('device.__id', this.hostAppOperatedDevice.id);
					expect(res.body).to.have.nested.property('on__application.__id', this.hostApp.id);
					expect(res.body.overrides__profile_name, 'kernel-modules');
					expect(res.body.is_active).to.equal(true);
```

or something even cooler that I saw recently
```suggestion
					expect(res.body).to.deep.include({
						device: {__id: this.hostAppOperatedDevice.id },
						on__application: {__id: this.hostApp.id },
						overrides__profile_name: 'kernel-modules',
						is_active: true,
					});
```

test/27_profiles.ts
```diff
@@ -372,6 +380,227 @@ export default () => {
...
+				});
+
+				it('should reject a duplicated override for the same device/profile/application', async function () {
+					await supertest(this.user)
+						.post(`/${version}/device_profile_override`)
+						.send({
+							device: this.hostAppOperatedDevice.id,
+							overrides__profile_name: 'kernel-modules',
+							on__application: this.hostApp.id,
+							is_active: true,
```
thgreasi: optional: I guess you can even have the props that's not part of the unique index have a different value, just for the shake of it
```suggestion
							is_active: false,
```

test/27_profiles.ts
```diff
@@ -372,6 +380,227 @@ export default () => {
...
+						})
+						.expect(409);
+				});
+
+				it('should reject an on__application the device is only operating a different release for', async function () {
```
thgreasi: Does that mean v ?
```suggestion
				it('should reject a device_profile_override that is for a different hostApp than the one operating the device', async function () {
```

test/27_profiles.ts
```diff
@@ -372,6 +380,227 @@ export default () => {
...
+						})
+						.expect(201);
+
+					await supertest(this.user)
+						.delete(`/${version}/device(${tempDevice.id})`)
+						.expect(200);
+
+					const res = await supertest(this.user)
+						.get(
+							`/${version}/device_profile_override?$select=id&$filter=device eq ${tempDevice.id}`,
```
thgreasi, line 571: Optional: In other places afair we are using the id returned by the POST, but tbh this test is already more than what I would have asked

test/27_profiles.ts
```diff
@@ -372,6 +380,227 @@ export default () => {
...
+						)
+						.expect(200);
+					expect(res.body).to.have.nested.property('d.length', 0);
+				});
+
```
thgreasi, line 576: An interesting test would be trying to changing the `should_be_operated_by__release` to a different release of the same hostApp, since that's happening after every HUP.

test/27_profiles.ts
```diff
@@ -372,6 +380,227 @@ export default () => {
...
+							is_active: true,
+						})
+						.expect(201);
+
+					await supertest(this.user)
+						.delete(`/${version}/application(${this.hostApp3.id})`)
+						.expect(409);
+				});
+
+				it('should reject an on__application the device only belongs to (but does not operate a release of)', async function () {
```
thgreasi: Having a device being part of a hostApp might confuse us in the future, and it's not something that we do as you said.
I would say let's remove it, since I'm already happy w/ the `should reject a target application that is not a hostApp'` test. Maybe on that one you could add an
```
expect(this.app1Device).to.have.deep.property('belongs_to__application.__id', this.app1.id);
```
to confirm that we are indeed using the fleet that the device belongs to.

thgreasi (approved):

PR #2433, opened 2026-08-25
Do not change supervisor pin when provisioning if pin already set

pipex:
The API sets the `should_be_managed_by__release` property on the device on first report of the `supervisor_version`, however it sets this value even if there is already a pin set. This should not really cause issues for customers but it affects supervisor developers testing with draft releases.

Change-type: patch

### DIFF
--- a/src/features/supervisor-app/hooks/supervisor-app.ts
+++ b/src/features/supervisor-app/hooks/supervisor-app.ts
@@ -36,7 +36,10 @@ hooks.addPureHook('POST', 'resin', 'device', {
 				},
 			);
 
-			if (supervisorRelease == null) {
+			if (
+				supervisorRelease == null &&
+				request.values.should_be_managed_by__release != null
+			) {
 				return;
 			}
 			// We are not using setSupervisorReleaseResource in a POSTRUN, since that only sets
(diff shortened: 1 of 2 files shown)


src/features/supervisor-app/hooks/supervisor-app.ts
```diff
@@ -36,7 +36,10 @@ hooks.addPureHook('POST', 'resin', 'device', {
 				},
 			);
 
-			if (supervisorRelease == null) {
+			if (
+				supervisorRelease == null &&
+				request.values.should_be_managed_by__release != null
```
thgreasi, line 41: This can probably be higher in the top
```
		if (
			typeof request.values.supervisor_version === 'string' &&
			typeof request.values.is_of__device_type === 'number' &&
			request.values.should_be_managed_by__release == null
		) {
```

thgreasi: Hmmmm on second though, I guess we should still overwrite the should_be_managed_by__release in case it's a lower version than the supervisor_release, to avoid a "supervisor preloaded device" downgrading right after it provisions.

pipex:
This PR is pointless, the POST hook is for registration only

PR #2447, opened 2026-09-02
Allow setting PROXY_DEVICE_TYPES_V1_ENDPOINTS for getting device types

otaviojacobi:
Change-type: major


src/features/device-types/device-types-list.ts
```diff
@@ -81,150 +83,181 @@ for (const contractPath of CONTRACT_ALLOWLIST) {
...
+	}
+	return Object.fromEntries(deviceTypes.map((dt) => [dt.slug, dt]));
+};
+
+export const getDeviceTypeJsons = multiCacheMemoizee(
+	async (): Promise<Record<string, DeviceTypeJson>> => {
+		const result =
+			PROXY_DEVICE_TYPES_V1_ENDPOINTS != null
+				? await getProxiedDeviceTypeJsons(PROXY_DEVICE_TYPES_V1_ENDPOINTS)
+				: await getDbDeviceTypeJsons();
```
thgreasi, line 260: Let's just add a prefix to make clear that this should be internal calls not used anywhere else
```suggestion
				? await $getProxiedDeviceTypeJsons(PROXY_DEVICE_TYPES_V1_ENDPOINTS)
				: await $getDbDeviceTypeJsons();
```

src/lib/config.ts
```diff
@@ -203,6 +203,9 @@ export const LOGS_HOST = optionalVar('LOGS_HOST');
 export const MIXPANEL_TOKEN = requiredVar('MIXPANEL_TOKEN');
 export const NODE_ENV = optionalVar('NODE_ENV');
 export const NODE_EXTRA_CA_CERTS = optionalVar('NODE_EXTRA_CA_CERTS');
+export const PROXY_DEVICE_TYPES_V1_ENDPOINTS = optionalVar(
```
thgreasi, line 206: I guess we also need this in env.tmpl ?

src/features/device-types/device-types-list.ts
otaviojacobi: I was not aware of this pattern @thgreasi as far as I can see $ calls are just the ones that have an equivalent name so the names do not clash?

thgreasi: The one that I had in mind where it's used as an internal-use-only indicator was probably this v since afaict we have inlined most other places that I remember (primarily memoizations)
https://github.com/balena-io/open-balena-api/blob/v49.0.26/src/infra/auth/jwt.ts#L79

would welcome it as an extra indicator but wouldn't block on it

PR #2465, opened 2026-09-17
Add application profile catalog modeling

otaviojacobi:
See: https://balena.fibery.io/Work/Project/Enabling-adding-aditional-information-to-a-profile-2841/Model-proposal-for-Application-Profile-Catalog-v1.0-5581
Change-type: minor


src/migrations/00117-backfill-application-profile-catalog.async.ts
```diff
@@ -0,0 +1,45 @@
+import type { Migrator } from '@balena/pinejs';
+
+const migration: Migrator.AsyncMigration = {
+	asyncSql: `\
+		WITH candidate_apps AS (
```
otaviojacobi, line 5: @thgreasi I wonder, what do you think of doing a standard sync .sql migration here? the data is minimal in any environment, and probably non-existing on any open-balena installation

src/lib/auth.ts
```diff
@@ -65,6 +65,8 @@ export const ROLES: {
 		'resin.image_profile.all',
 		'resin.application_profile.all',
 		'resin.device_profile_override.all',
+		'resin.application_profile_catalog.read?application/canAccess()',
+		'resin.application_profile_catalog.update?application/canAccess()',
```
thgreasi: Afair canAccess wasn't working for permissions other than `.read`. Not sure what changed and it now works here.
I'm not sure whether it will under the hood translate to `application.read` or `application.update`, but in this case I don't think it matters and it could even be `resin.application_profile_catalog.all` for oB

src/migrations/00117-backfill-application-profile-catalog.async.ts
thgreasi: Yes, let's make it sync, even for bC, since there are minimal data.
As a result the data population can even be part of `src/migrations/00116-add-application-profile-catalog.sql`

src/migrations/00117-backfill-application-profile-catalog.async.ts
```diff
@@ -0,0 +1,45 @@
...
+			LIMIT %%ASYNC_BATCH_SIZE%%
+		)
+		INSERT INTO "application profile catalog" ("application", "catalogs-profile name")
+		SELECT DISTINCT r."belongs to-application", ip."profile name"
+		FROM candidate_apps ca
+		JOIN "release" r ON r."belongs to-application" = ca.application_id
+		JOIN "image-is part of-release" ri ON ri."is part of-release" = r.id
+		JOIN "image profile" ip ON ip."release image" = ri.id
+		WHERE r.status = 'success'
+		ON CONFLICT ("application", "catalogs-profile name") DO NOTHING;
```
thgreasi: Isn't all this just v ? Which gets the profile names along w/ the app ids and avoids a number of duplicate JOINs later?
```suggestion
		WITH app_profile_catalog AS (
			SELECT DISTINCT r."belongs to-application", ip."profile name"
			FROM "release" r
			JOIN "image-is part of-release" ri ON ri."is part of-release" = r.id
			JOIN "image profile" ip ON ip."release image" = ri.id
			WHERE r.status = 'success'
			AND NOT EXISTS (
				SELECT 1 FROM "application profile catalog" apc
				WHERE apc."application" = r."belongs to-application"
				AND apc."profile name" = ip."profile name"
			)
			LIMIT %%ASYNC_BATCH_SIZE%%
		)
		INSERT INTO "application profile catalog" ("application", "catalogs-profile name")
		SELECT apc."belongs to-application", apc."profile name"
		FROM app_profile_catalog apc;
```

src/migrations/00117-backfill-application-profile-catalog.async.ts
```diff
@@ -0,0 +1,45 @@
...
+	`,
+	syncSql: `\
+		INSERT INTO "application profile catalog" ("application", "catalogs-profile name")
+		SELECT DISTINCT r."belongs to-application", ip."profile name"
+		FROM "release" r
+		JOIN "image-is part of-release" ri ON ri."is part of-release" = r.id
+		JOIN "image profile" ip ON ip."release image" = ri.id
+		WHERE r.status = 'success'
+		ON CONFLICT ("application", "catalogs-profile name") DO NOTHING;
```
thgreasi: Not trying to insert is better than adding the tuple and only finding out that there is a conflict while checking the unique index.
```suggestion
		AND NOT EXISTS (
			SELECT 1 FROM "application profile catalog" apc
			WHERE apc."application" = r."belongs to-application"
			AND apc."profile name" = ip."profile name"
		);
```

src/features/profiles/hooks.ts
```diff
@@ -53,3 +69,162 @@ registerProfileValidation(
...
+			resource: 'image__is_part_of__release',
+			id: releaseImageId,
+			passthrough: { tx: args.tx, req: permissions.rootRead },
+			options: {
+				$select: 'id',
+				$expand: {
+					is_part_of__release: { $select: 'belongs_to__application' },
+				},
+			},
+		});
```
thgreasi: NITs
1. Do we actually need rootRead in any of these GETs?
2. since we don't really need to read anything from the release_image, maybe we can start the query from the release.

```suggestion
		const [release] = await args.api.get({
			resource: 'release',
			options: {
				$select: 'belongs_to__application',
				$filter: {
					release_image: {
						$any: {
							$alias: 'ri',
							$expr: {
								ri: {
									id: releaseImageId,
								},
							},
						},
					},
				}
			},
		});
```

src/features/profiles/hooks.ts
```diff
@@ -53,3 +69,162 @@ registerProfileValidation(
...
+
+		await args.tx.getTxLevelLock(CATALOG_LOCK_NAMESPACE_KEY, applicationId);
+
+		const existing = await sbvrUtils.api.resin.get({
+			resource: 'application_profile_catalog',
+			id: {
+				application: applicationId,
+				catalogs__profile_name: catalogProfileName,
+			},
+			passthrough: { tx: args.tx, req: permissions.rootRead },
```
thgreasi: `passthrough` right after `resource` plz 🙏  (or removed if we can)

src/features/profiles/hooks.ts
```diff
@@ -53,3 +69,162 @@ registerProfileValidation(
...
+		for (const row of rows) {
+			const applicationId =
+				row.release_image[0].is_part_of__release[0].belongs_to__application
+					.__id;
+			candidatesByKey.set(`${applicationId}|${row.profile_name}`, {
+				applicationId,
+				catalogProfileName: row.profile_name,
+			});
+		}
+		args.request.custom.profileCatalogCandidates = [
```
thgreasi: Let's define a 
```ts
interface DeleteRequestCustomObject {
	profileCatalogCandidates?: ...
}
```

and then do in the PRERUN
```ts
(args.request.custom as DeleteRequestCustomObject).profileCatalogCandidates
```
and in the POSTRUN
```ts
const { profileCatalogCandidates } = (args.request.custom as DeleteRequestCustomObject);
```

src/features/profiles/hooks.ts
```diff
@@ -53,3 +69,162 @@ registerProfileValidation(
...
+	},
+	POSTRUN: async (args) => {
+		const candidates = args.request.custom.profileCatalogCandidates as
+			Array<{ applicationId: number; catalogProfileName: string }> | undefined;
+		if (!candidates?.length) {
+			return;
+		}
+
+		await Promise.all(
+			candidates.map(async ({ applicationId, catalogProfileName }) => {
```
thgreasi: We should probably iterate on these instead of using a `Promise.all` and also have the `candidates` sorted by application id, so that we can avoid deadlocks.

src/features/profiles/hooks.ts
```diff
@@ -53,3 +69,162 @@ registerProfileValidation(
...
+				await args.tx.getTxLevelLock(CATALOG_LOCK_NAMESPACE_KEY, applicationId);
+				if (
+					await stillHasImageProfile(args.tx, applicationId, catalogProfileName)
```
thgreasi: This query, and the advisory lock acquisition are going to be called once for each profile name, which when deleting a fleet, might mean a lot of requests.

It would be nice if we could
1. create a `custom.profileNamesByApplicationId: Map<number, string[]> = ...` in the PRERUM
2. rewrite this query, so that it finds all no longer existing profile_names per application (which in the current form, will not be able to use `$top: 1` and thus might return tons of data for still-in-use profile_names).

If we had an image_profile -> applicatio_profile_catalog FK like we were discussing, we could even check all the profile_names in a single query.
Should we add it as we discussed?

src/migrations/00117-backfill-application-profile-catalog.async.ts
otaviojacobi: I think we remove the whole async thing

src/migrations/00117-backfill-application-profile-catalog.async.ts
otaviojacobi: I was mostly wanting to avoid in case it conflicted with a hook but yeah, it could have both, either case it the problem goes away with a sync version

src/features/profiles/hooks.ts
otaviojacobi: Well the answer to rootRead will depend on how permissions look like on balena-api side of things (but theoretically they will be fine). I wanted to use it for consistency with the idea of a "managed hook" which does not rely on your permissions

src/features/profiles/hooks.ts
otaviojacobi: I was planning on adding it in a follow up to be honest, I will try to get us to discuss it again

src/features/profiles/hooks.ts
```diff
@@ -53,3 +69,173 @@ registerProfileValidation(
...
+			return;
+		}
+
+		const applicationIds = [...profileNamesByApplicationId.keys()].sort(
+			(a, b) => a - b,
+		);
+		for (const applicationId of applicationIds) {
+			const candidateProfileNames = [
+				...profileNamesByApplicationId.get(applicationId)!,
+			];
```
thgreasi: We can avoid the `.get()` & the `!`
```suggestion
		const profileNamesByApplicationIdEntries = [...profileNamesByApplicationId.entries()].sort(
			(a, b) => a[0] - b[0],
		);
		for (const [applicationId, candidateProfileNames] of profileNamesByApplicationIdEntries) {
```

src/migrations/00116-add-application-profile-catalog.sql
```diff
@@ -0,0 +1,53 @@
...
+		EXECUTE PROCEDURE "trigger_update_modified_at"();
+	END IF;
+END;
+$$;
+
+INSERT INTO "application profile catalog" ("application", "catalogs-profile name")
+SELECT DISTINCT r."belongs to-application", ip."profile name"
+FROM "release" r
+JOIN "image-is part of-release" ri ON ri."is part of-release" = r.id
+JOIN "image profile" ip ON ip."release image" = ri.id
```
thgreasi: For consistency
```suggestion
JOIN "image-is part of-release" ri ON ri."is part of-release" = r."id"
JOIN "image profile" ip ON ip."release image" = ri."id"
```

src/features/profiles/hooks.ts
thgreasi: I will insist on not using root, unless we later find out that we absolutely must do so.
We are generally following a "user permissions by default" approach, and whenever we use root we should have a comment on why we had to.
The benefit of using user permissions, is that the code is inherently more secure and has less chance of unwanted side-effects in the case of a bug.

Since `track-latest-release.ts` is able to w/o root, I would expect that this one should also be able to do so, (other than the POST & DELETE of the application_profile_catalog of course, which need root b/c users shouldn't have access to manage them directly).

src/lib/auth.ts
```diff
@@ -65,6 +65,7 @@ export const ROLES: {
 		'resin.image_profile.all',
 		'resin.application_profile.all',
 		'resin.device_profile_override.all',
+		'resin.application_profile_catalog.all',
```
thgreasi: This should probably be just `.read`, since we wouldn't want users to be able to manually modify this computed resource, right?
```suggestion
		'resin.application_profile_catalog.read',
```

src/migrations/00116-add-application-profile-catalog.sql
```diff
@@ -0,0 +1,53 @@
...
+WHERE r.status = 'success'
+AND NOT EXISTS (
+	SELECT 1 FROM "application profile catalog" apc
+	WHERE apc."application" = r."belongs to-application"
+	AND apc."profile name" = ip."profile name"
```
thgreasi:
```suggestion
	AND apc."catalogs-profile name" = ip."profile name"
```

src/lib/auth.ts
otaviojacobi: @thgreasi  We need at least `.update` and my initial idea was to use `canAccess` here but you explicitely asked me not to and to use .all as it is now https://github.com/balena-io/open-balena-api/pull/2465#discussion_r4069214584 :sweat_smile:

src/migrations/00116-add-application-profile-catalog.sql
otaviojacobi: Nice finding, I will run it on bok with some data to see in details how it goes, thanks

test/27_profiles.ts
```diff
@@ -43,6 +43,149 @@ export default () => {
 				await fixtures.clean(this.loadedFixtures);
 			});
 
+			describe('application profile catalog', function () {
+				let catalogTestImageProfile1Id: number;
+				let catalogTestImageProfile2Id: number;
+				let catalogTestEntryId: number;
+
+				it('should have created a catalog entry for the pre-existing image_profile fixtures', async function () {
```
thgreasi: I guess this is more like
```suggestion
				it('should be able to read a pre-existing catalog entry created for the pre-existing image_profile fixtures', async function () {
```

test/27_profiles.ts
```diff
@@ -43,6 +43,149 @@ export default () => {
...
+						.expect(201);
+					catalogTestImageProfile2Id = imageProfile.id;
+
+					await expectResourceToMatch(
+						pineUser,
+						'application_profile_catalog',
+						{
+							application: this.app1.id,
+							catalogs__profile_name: 'catalog-test',
+						},
```
thgreasi: I guess in order to be able to test this properly you would need to use $filter, b/c by the fact that you are using an alternate notation id, you already know that pine knows that they are a natural key of the table.

Maybe a more interesting test case would have been to create two image profiles for the same new profile on two different release images concurrently w/ a Promise.all, to confirm that it works and there is a single catalog entry.

src/lib/auth.ts
thgreasi: yea, I'm sorry, it was Friday....
Or maybe I hadn't read enough of the PR to remember that we shouldn't allow users to modify the catalog directly.

thgreasi (approved):
Draftpproved to give you the chance to check my comment on the tests.

